PR #5, opened 2020-02-03
Basic inference-on-jvm scaffold.  

jakemannix:
Doesn't really do anything yet, nor have a working test.  More to come!


jakemannix:
ok @darshshah  - I pushed changes which verify that with TF1.15, we can run inference on the simple models the python subdirectory trains, using @lastmansleeping's TFRecord serving signature.

This code is a total mess right now, but if you run ml4ir.inference.tensorflow.TensorFlowInferenceTest#testSavedModelBundle(), you can step through the construction of the protobuf and feeding it to the TF model via the serving signature.

Oh crap, it looks like I have a hard-coded path to my local filesystem in there.  I need to check in a tiny SavedModelBundle into test-resources.  Then I'll push again.

darshshah:
I have pulled these changes locally and am in the process of reviewing them.

jvm/ml4ir-inference/pom.xml
```diff
@@ -0,0 +1,183 @@
...
+	  </configuration>
+	</plugin>
+      </plugins>
+    </pluginManagement>
+  </build>
+
+  <dependencies>
+    <dependency>
+    	<groupId>org.tensorflow</groupId>
+    	<artifactId>tensorflow</artifactId>
```
darshshah: Can we add an explicit version number here?

jvm/ml4ir-inference/pom.xml
```diff
@@ -0,0 +1,183 @@
...
+	</executions>
+      </plugin>
+      <plugin>
+	<artifactId>maven-compiler-plugin</artifactId>
+	<configuration>
+	  <source>1.7</source>
+	  <target>1.7</target>
+	</configuration>
+      </plugin>
+		<!-- disable surefire - - >
```
darshshah: Will we need this in the future? If not, then remove

jvm/ml4ir-inference/pom.xml
jakemannix: I think it's in the parent pom already (you have to specify)

jvm/ml4ir-inference/src/test/scala/ml4ir/inference/tensorflow/TensorFlowInferenceTest.scala
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  @Test
+  def testTFRecordInference() = {
+    val queryString = "magic"
+    val docsToScore = Array(
+      Map("feat_0" -> 0.04f, "feat_1" -> 0.08f, "feat_2" -> 0.01f),
+      Map("feat_0" -> 0.4f, "feat_1" -> 0.8f, "feat_2" -> 0.1f)
+    )
+    val (query, docs) = (
```
darshshah: This test looks incomplete. What are we trying to validate here? The query/docs val is not used after creation

darshshah: We should call PointwiseDenseWithCharsModelExecutor() with the query and docs

jvm/ml4ir-inference/pom.xml
jakemannix: This is the way to make sure maven can work with scalatest: 
http://www.scalatest.org/user_guide/using_the_scalatest_maven_plugin

jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/PointwiseDenseWithCharsModelExecutor.scala
```diff
@@ -0,0 +1,48 @@
...
+      .fetch(OUTPUT_NODE)
+      .run()
+      .get(0)
+      .copyTo(ranking)
+
+    ranking(0)
+  }
+
+  // TODO: ensure this is in sync with training in python
+  def textFeatureProcessor(str: String): Array[Float] =
```
lastmansleeping: I think raw strings would be better, as I can do some clean up if needed.

jvm/ml4ir-inference/pom.xml
```diff
@@ -0,0 +1,183 @@
...
+      <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-project-info-reports-plugin</artifactId>
+          <version>3.0.0</version>
+      </plugin>
+      <!-- mixed scala/java compile -->
+      <plugin>
+	<groupId>net.alchim31.maven</groupId>
+	<artifactId>scala-maven-plugin</artifactId>
+        <version>3.3.1</version>
```
jerbome: I'd advise to use a more version of the scala plugin. I experienced bugs wrt scala console when using this version

jvm/ml4ir-inference/pom.xml
```diff
@@ -0,0 +1,183 @@
...
+	    <goals>
+	      <goal>compile</goal>
+	    </goals>
+	  </execution>
+	</executions>
+      </plugin>
+      <plugin>
+	<artifactId>maven-compiler-plugin</artifactId>
+	<configuration>
+	  <source>1.7</source>
```
jerbome: this is inconsistent with the java.version property in the parent pom

jvm/ml4ir-inference/pom.xml
```diff
@@ -0,0 +1,183 @@
...
+              <!-- for fatjar -->
+      <plugin>
+	<groupId>org.apache.maven.plugins</groupId>
+	<artifactId>maven-assembly-plugin</artifactId>
+	<version>3.1.0</version>
+	<configuration>
+	  <descriptorRefs>
+	    <descriptorRef>jar-with-dependencies</descriptorRef>
+	  </descriptorRefs>
+	  <finalName>uber-ml4ir-parent-0.0.1-SNAPSHOT</finalName>
```
jerbome: I prefer not to use finalName in plugins that produces others maven artifact: you lose the auto versioning of names when you do that. Furthermore, using classifiers it more the maven way

jvm/ml4ir-inference/pom.xml
```diff
@@ -0,0 +1,183 @@
...
+              <goals>
+                <goal>test</goal>
+              </goals>
+            </execution>
+          </executions>
+        </plugin> -->
+              <!-- for fatjar -->
+      <plugin>
+	<groupId>org.apache.maven.plugins</groupId>
+	<artifactId>maven-assembly-plugin</artifactId>
```
jerbome: Is a uber jar actually needed for this project?

jvm/ml4ir-inference/pom.xml
```diff
@@ -0,0 +1,183 @@
...
+	</executions>
+      </plugin>
+      <plugin>
+	<groupId>org.apache.maven.plugins</groupId>
+	<artifactId>maven-jar-plugin</artifactId>
+	<configuration>
+	  <archive>
+	    <manifest>
+	      <addClasspath>true</addClasspath>
+	      <mainClass>fully.qualified.MainClass</mainClass>
```
jerbome: seems incorrect

jvm/ml4ir-inference/pom.xml
```diff
@@ -0,0 +1,183 @@
...
+	    <artifactId>proto</artifactId>
+    </dependency>
+	<dependency>
+	  <groupId>com.google.protobuf</groupId>
+	  <artifactId>protobuf-java-util</artifactId>
+	  <version>3.11.4</version>
+	</dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
```
jerbome: should be scoped to test

jvm/ml4ir-inference/src/main/java/ml4ir/inference/tensorflow/SequenceExampleJavaBuilder.java
```diff
@@ -0,0 +1,58 @@
+package ml4ir.inference.tensorflow;
```
jerbome: the maven build does not compile this file, it is not present in the jar

jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/PointwiseDenseWithCharsModelExecutor.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import ml4ir.inference.tensorflow.utils.TensorUtils
+import org.tensorflow.Session
+
+class PointwiseDenseWithCharsModelExecutor(session: Session) {
+
+  val INPUT_QUERY_CHAR_NODE = "INPUT_QUERY_CHAR"
+  val INPUT_DOC_CHAR_NODE = "INPUT_DOC_CHAR"
+  val INPUT_DOC_NUMERIC_NODE = "INPUT_DOC_NUMERIC"
+  val OUTPUT_NODE = "OUTPUT"
```
jerbome: minor: shouldn't those 4 constant be in the companion object?

jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/utils/TensorUtils.scala
```diff
@@ -0,0 +1,18 @@
...
+import org.tensorflow.Tensor
+
+/**
+  * Simple helpers for creating TensorFlow {@see Tensor}s of the right shapes.
+  */
+object TensorUtils {
+  def replicate(encoding: Array[Float], length: Int): Array[Array[Float]] =
+    Array.ofDim[Float](length, encoding.length).map(_ => encoding.clone())
+
+  def create1Tensor(encoding: Array[Float]): Tensor[java.lang.Float] =
```
jerbome: the class `Tensors` provide exactly this funcitonality

jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/utils/TensorUtils.scala
```diff
@@ -0,0 +1,18 @@
+package ml4ir.inference.tensorflow.utils
+
+import org.tensorflow.Tensor
+
+/**
+  * Simple helpers for creating TensorFlow {@see Tensor}s of the right shapes.
+  */
+object TensorUtils {
+  def replicate(encoding: Array[Float], length: Int): Array[Array[Float]] =
+    Array.ofDim[Float](length, encoding.length).map(_ => encoding.clone())
```
jerbome: `Array.fill` may be cleaner?

jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/PointwiseDenseWithCharsModelExecutor.scala
jakemannix: so maybe I should just delete all but the new-and-fancy SavedModelBundle versions in here, for code-cleanliness.  We can always resurrect them out of git-history if needed...

jvm/ml4ir-inference/pom.xml
jakemannix: upgrading to 4.3.1!

jvm/ml4ir-inference/pom.xml
jakemannix: changing to be `${java.version}`!

jvm/ml4ir-inference/pom.xml
jakemannix: not really.  I do have another branch with a bunch of scala-spark preprocessing stuff, but this branch?  Doesn't need it.  Removing.

jvm/ml4ir-inference/pom.xml
jakemannix: hah.  Yeah, no mainClass being used - we don't have standalone usage (for now).  Removing!

jvm/ml4ir-inference/src/main/java/ml4ir/inference/tensorflow/SequenceExampleJavaBuilder.java
jakemannix: @darshshah do you have a PR which got this code in the jar?  How did you build locally to fix my mixup here?

jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/PointwiseDenseWithCharsModelExecutor.scala
jakemannix: yep, much better in the companion.  But this class probably needs to go away.

jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/utils/TensorUtils.scala
jakemannix: deleting that class but yes, good point.

PR #6, opened 2020-02-04
Adding ranking validation metrics

lastmansleeping:
This PR contains the following changes

- MRR, ACR validation metrics
- ~Functionality to compute additional group based aggregations on metrics~
- Writing old rank, new rank, new score for each record being ranked
- Unit tests for the above

### DIFF
--- a/python/ml4ir/model/metrics/metric_factory.py
+++ b/python/ml4ir/model/metrics/metric_factory.py
@@ -19,6 +19,10 @@ def get_metric(metric_key):
 
 def get_metric_impl(metric: Union[str, Type[MeanMetricWrapper]], **kwargs) -> Union[str, Metric]:
     if isinstance(metric, str):
-        return metric
+        return [metric]
     else:
-        return metric(**kwargs)
+        # Compute both current and new metric after reranking
+        return [
+            metric(rerank=False, **kwargs),
+            metric(rerank=True, **kwargs),
+        ]
--- a/python/ml4ir/model/metrics/metric_factory.py
+++ b/python/ml4ir/model/metrics/metric_factory.py
@@ -1,16 +1,24 @@
 from ml4ir.config.keys import MetricKey
-
-# import tensorflow as tf
+from ml4ir.model.metrics.metrics_impl import MRR, ACR, MeanMetricWrapper
+from typing import Union, Type
+from tensorflow.keras.metrics import Metric
 
 
 def get_metric(metric_key):
     if metric_key == MetricKey.MRR:
-        raise NotImplementedError
+        return MRR
     elif metric_key == MetricKey.ACR:
-        raise NotImplementedError
+        return ACR
     elif metric_key == MetricKey.NDCG:
         raise NotImplementedError
     elif metric_key == MetricKey.CATEGORICAL_ACCURACY:
         return "categorical_accuracy"
     else:
         raise NotImplementedError
+
+
+def get_metric_impl(metric: Union[str, Type[MeanMetricWrapper]], **kwargs) -> Union[str, Metric]:
+    if isinstance(metric, str):
+        return metric
+    else:
+        return metric(**kwargs)
--- a/python/ml4ir/model/metrics/metric_factory.py
+++ b/python/ml4ir/model/metrics/metric_factory.py
@@ -1,10 +1,10 @@
 from ml4ir.config.keys import MetricKey
-from ml4ir.model.metrics.metrics_impl import MRR, ACR, MeanMetricWrapper
-from typing import Union, Type
+from ml4ir.model.metrics.metrics_impl import MRR, ACR, CategoricalAccuracy
+from typing import Type, List
 from tensorflow.keras.metrics import Metric
 
 
-def get_metric(metric_key):
+def get_metric(metric_key: str) -> Metric:
     if metric_key == MetricKey.MRR:
         return MRR
     elif metric_key == MetricKey.ACR:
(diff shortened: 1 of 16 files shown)


lastmansleeping:
@jakemannix Still fixing/adding unittests. But the two main portions of the PR are done
- adding validation metrics : MRR and ACR
- writing old rank, new rank to file with other features

python/ml4ir/model/metrics/metric_factory.py
```diff
@@ -1,16 +1,28 @@
...
     elif metric_key == MetricKey.NDCG:
         raise NotImplementedError
     elif metric_key == MetricKey.CATEGORICAL_ACCURACY:
         return "categorical_accuracy"
     else:
         raise NotImplementedError
+
+
+def get_metric_impl(metric: Union[str, Type[MeanMetricWrapper]], **kwargs) -> Union[str, Metric]:
+    if isinstance(metric, str):
```
jakemannix: when do we use this if branch?

python/ml4ir/model/ranking_model.py
```diff
@@ -152,11 +160,21 @@ def build_model(
         # Create model with functional Keras API
         model = Model(inputs=inputs, outputs={"ranking_scores": predictions})
 
+        # Get loss and metrics
+        loss_fn = loss.get_loss_fn(mask=metadata_features["mask"])
+        metric_fns: List[kmetrics.Metric] = list()
```
jakemannix: just do this as a for-comprension, no?

`metric_fns: List[kmetrics.Metric] = [metric_factor.get_metric_impl(metric, ...) for metric in metrics]`

python/ml4ir/tests/test_serving.py
```diff
@@ -87,32 +87,40 @@ def get_dataset(parse_tfrecord):
...
-        model_predictions = model.predict(parsed_dataset_batch)
+        model_predictions = model.predict(parsed_dataset_batch)["new_score"].values
         default_signature_predictions = default_signature(**parsed_sequence_examples)[
             "ranking_scores"
         ]
         tfrecord_signature_predictions = tfrecord_signature(
             sequence_example_protos=sequence_example_protos
         )["ranking_scores"]
 
+        @tf.function
```
jakemannix: tf.functions inside of test classes?  Can you write up what these are doing?

lastmansleeping: They don't _have to_ be `tf.function` actually. Can remove if you think it is inappropriate. I am trying to replicate the same flatten and filtering out padded records that I do in the `.predict()` method of `RankingModel` to the signatures as well in order to unit test that the predictions from all serving signatures + the actual keras model is the same.

python/ml4ir/model/ranking_model.py
lastmansleeping: I can. It will be long and dirty. I favored clean code here as opposed to memory(very minimal). Also, I am doing an `extend` and not an `append`. So it will be nested list comprehension. Again, doable - just ugly. Let me know if you think that is better.

python/ml4ir/model/metrics/metric_factory.py
lastmansleeping: For inbuilt tf metrics - https://www.tensorflow.org/api_docs/python/tf/keras/metrics/CategoricalAccuracy

Look at this above
```
elif metric_key == MetricKey.CATEGORICAL_ACCURACY:
        return "categorical_accuracy"	        return "categorical_accuracy"
```

python/ml4ir/model/ranking_model.py
jakemannix: oh get_metric_impl() returns a list... hence you'd really like a flatMap method :\

ok, so this *could* be done with a `reduce()` call.  But no, it wouldn't be any more readable than what you've got.

python/ml4ir/tests/test_serving.py
jakemannix: you don't need to remove it, but it's confusing that there are functions meant to be wired into the graph, and yet we're in the test classes...

You're writing code which replicates what is done with padding in the main model, to test it.  I think it's just hard for me to read this to be sure I understand the logic we "know" is correct.

python/ml4ir/model/metrics/metric_factory.py
jakemannix: isn't there an actual function in the core tf library we can refer to, instead of by name?  Like using a wrapper around `tf.keras.metrics.categorical_accuracy`?

lastmansleeping: Yes, I was hoping you wouldn't catch it. Let me change it.

jakemannix: whenever you see your method signatures have super weird `Union` types (like `[str,Type[MeanMetricWrapper]]`), there's a hint that maybe it could be refactored.

python/ml4ir/model/ranking_model.py
```diff
@@ -73,7 +73,7 @@ def __init__(
             )
 
             # Define metrics
-            metrics: List[str] = [
+            metrics: List[Type[kmetrics.Metric]] = [
```
jakemannix: why does this need too be `List[Type[kmetrics.Metric]]` and not just `List[kmetrics.Metric]`, do you know?

python/ml4ir/model/metrics/metrics_impl.py
```diff
@@ -163,3 +163,15 @@ def __init__(self, name="ACR", rerank=True, **kwargs):
     def _get_matches_hook(self, y_pred_click_ranks):
         """Return click ranks for MRR"""
         return tf.cast(y_pred_click_ranks, tf.float32)
+
+
+class CategoricalAccuracy(metrics.CategoricalAccuracy):
```
jakemannix: question about these "non-public" keras classes.  Can you really not import it anyways?  Why do we need to copy/paste their class?

python/ml4ir/model/ranking_model.py
lastmansleeping: It is `Type[Metric]` because it is a class and not instantiated yet.

python/ml4ir/model/metrics/metrics_impl.py
lastmansleeping: I don't understand. What do you mean by copy/paste their class?

python/ml4ir/model/ranking_model.py
jakemannix: it's a class and not instantiated yet?  I... don't understand that.  It's returning a function which returns a Metric?  Is that what you mean?

PR #8, opened 2020-02-28
Making feature_config more generic + cleanup + fixing FIXMEs

lastmansleeping:


lastmansleeping:
@jakemannix Can finally train some models!!!

P.S. When you spend the whole night developing awesome code and then `mypy` just doesn't want you to commit - 
image

jakemannix:
> @jakemannix Can finally train some models!!!
> 
> P.S. When you spend the whole night developing awesome code and then `mypy` just doesn't want you to commit -
> image

commit early, commit often (to your local branch)!

python/ml4ir/tests/data/tfrecord/feature_config.yaml
```diff
@@ -0,0 +1,89 @@
...
+    dtype: bytes
+    log_at_inference: true
+    feature_layer_info:
+      type: string
+      max_length: 20
+      shape: null
+    serving_info:
+      name: query_str
+      required: false
+    tfrecord_type: sequence
```
jakemannix: tfrecord_type should be context for query_str, right?

lastmansleeping:
@jakemannix Can be looked at again. Moved few things around and refined listwise losses and logging.

python/ml4ir/data/tfrecord_reader.py
```diff
@@ -99,7 +99,9 @@ def _parse_sequence_example_fn(sequence_example_proto):
                     mask = tf.expand_dims(mask, axis=2)
 
                     def crop_fn():
-                        tf.print("\n[WARN] Number of records in query : ", tf.shape(mask))
+                        tf.print(
+                            "\n[WARN] Bad query found. Number of records : ", tf.shape(mask)[1]
+                        )
                         return image.crop_to_bounding_box(
```
jakemannix: we still need to use `image.crop_to_bounding_box(`?  No text-based padding yet?

python/ml4ir/model/architectures/dnn.py
```diff
@@ -7,17 +7,17 @@ def architecture_op(ranking_features):
         first_dense = layers.Dense(num_nodes, activation="relu", name="first_dense")(
             ranking_features
         )
-        # first_dropout = layers.Dropout(rate=0.3, name="first_dropout")(first_dense)
+        first_dropout = layers.Dropout(rate=0.3, name="first_dropout")(first_dense)
```
jakemannix: make configurable?  dropout rate, and if 0, leave off the layer entirely?

python/ml4ir/data/tfrecord_reader.py
lastmansleeping: I don't think there can be - only this method allows me to pad to a **final shape**, the others expect you to specify the **extra shape you want to pad**. 
I've spent a lot of time on this previously and unsuccessfully. Can dig deeper if needed.

python/ml4ir/model/architectures/dnn.py
lastmansleeping: Okay. I was debating if I should and in turn probably make it more generic than it has to be. But good call, I like configs. Will do.

python/ml4ir/config/features.py
```diff
@@ -1,120 +1,237 @@
...
+            raise KeyError("'label' key not found in the feature_config specified")
+
+        try:
+            self.features: List = features_dict.get(FeatureConfigKey.FEATURES)
+            self.all_features.extend(self.features)
+        except KeyError:
+            raise KeyError("'features' key not found in the feature_config specified")
+
+        # Features that can be used for training the model
+        self.ranking_features = list()
```
jakemannix: what type of list is this?  can we have types for these?  `List[str]` or whatnot?

python/ml4ir/config/features.py
```diff
@@ -1,120 +1,237 @@
...
+                "Context Features : \n{}".format("\n".join(self.get_ranking_features("name")))
+            )
+            logger.info(
+                "Sequence Features : \n{}".format("\n".join(self.get_ranking_features("name")))
+            )
+
+        self.mask = self.generate_mask()
+        self.all_features.append(self.get_mask())
+
+    def _process_dict(self, dict_, key: str = None):
```
jakemannix: is "_process_dict" really the best name for this?  What is the type of this method, even?  It's the value of a the dict at a key, unless it's "node_name", in which case grab the "name", but if None is passed for the key, just return the dict?  

I'm confused.

python/ml4ir/config/features.py
lastmansleeping: I initially had it as `List[Dict[str, dict]]`, but then everything blew up and led to the `35` errors I posted above. I'm not sure how I would define the type for this effectively without breaking anything downstream.

The list essentially contains the feature info dictionaries for each feature.

Look at `get_ranking_features()`. It makes it even worse to define a dictionary because that method could essentially return anything(value in the dictionary).

If you want me to add the type for **only** this list, I could may be do that though.

python/ml4ir/config/features.py
lastmansleeping: You either get the whole dictionary, or a particular value if you have specified the key. I'm using this as a wrapper for `ranking_features`, `metadata_features`, `context_features`, `sequence_features`..
This method is needed because I have to define `node_name` with the default value of `name`

lastmansleeping: `_process_dict` is not a good name. I agree. But then again, I'm not sure what is. Do you have any suggestions?

python/ml4ir/config/features.py
jakemannix: oh I see recursive type is pretty nested.  Maybe a struct class to handle this config, instead of basically just JSON in memory?  But not for now I guess.   But something to consider...

python/ml4ir/config/features.py
lastmansleeping: It is a fancy `.get()`

jakemannix: `get_key_or_name` maybe?

python/ml4ir/config/features.py
```diff
@@ -1,120 +1,237 @@
...
+            if feature_info.get("log_at_inference", False):
+                self.features_to_log.append(feature_info)
+
+        if len(self.ranking_features) == 0:
             raise Exception("No trainable features specified in the feature config")
-        if not self.label:
-            raise Exception("No label field defined in the feature config")
 
-    def get_feature_names(self, trainable=None, include_label=False, dtype=None) -> List[str]:
+        if logger:
```
jakemannix: probably separate this whole if block into a logInitialization() method, to avoid clutter in the constructor

python/ml4ir/data/tfrecord_reader.py
```diff
@@ -68,76 +65,117 @@ def _parse_sequence_example_fn(sequence_example_proto):
...
             # If feature is a string, then decode into numbers
-            if feature_config.get_dict()[feat]["type"] == FeatureTypeKey.STRING:
-                t = io.decode_raw(
-                    t,
+            if feature_layer_info["type"] == FeatureTypeKey.STRING:
+                feature_tensor = io.decode_raw(
+                    feature_tensor,
                     out_type=tf.uint8,
-                    fixed_length=feature_config.get_dict()[feat]["max_length"],
+                    fixed_length=feature_layer_info["max_length"],
```
jakemannix: do we enforce anywhere that `feature_layer_info["max_length"]` is set?

python/ml4ir/data/tfrecord_reader.py
```diff
@@ -68,76 +65,117 @@ def _parse_sequence_example_fn(sequence_example_proto):
...
+                #     feature_tensor = tf.expand_dims(t, axis=0)
                 # if len(t.shape) == 2:
-                #     t = tf.pad(t, paddings=[[0, 0], [0, max_num_records]])
-                #     t = tf.squeeze(t)
+                #     feature_tensor = tf.pad(t, paddings=[[0, 0], [0, max_num_records]])
+                #     feature_tensor = tf.squeeze(t)
                 # else:
                 #     raise Exception('Invalid input : {}'.format(feat))
-                raise ValueError("Invalid input : {}".format(feat))
+                raise ValueError("Invalid input : {}".format(feature_name))
```
jakemannix: so we still don't handle dense?  MRUAffinity features aren't dense?  What does this else cover?

python/ml4ir/data/tfrecord_reader.py
```diff
@@ -68,76 +65,117 @@ def _parse_sequence_example_fn(sequence_example_proto):
...
+
+            features_dict[feature_name] = feature_tensor
 
-            features_dict[feat] = t
+        labels = features_dict.pop(feature_config.get_label(key="name"))
+
+        # Check if label is one-hot and correctly masked
+        tf.debugging.assert_equal(tf.cast(tf.reduce_sum(labels), tf.float32), tf.constant(1.0))
```
jakemannix: how expensive is tf.debugging?  When doing training, can you turn it off so your training runs go faster?

python/ml4ir/model/architectures/dnn.py
```diff
@@ -2,15 +2,22 @@
 from tensorflow.keras import layers
 
 
-def get_architecture_op(num_nodes=512):
+def get_architecture_op(num_nodes=256):
```
jakemannix: as discussed elsewhere, this method could take a dict of "num_nodes" -> int, "dropout_1" -> float, "dense_2_dim" -> int, "final_dropout" -> float, which could carry around the defaults from some higher place (eventually externalized in a config).

python/ml4ir/model/losses/listwise_losses.py
```diff
@@ -8,4 +12,40 @@ def _make_loss_fn(self, **kwargs):
...
+
+    def _final_activation_op(self):
+        # Without masking padded records
+        softmax = layers.Softmax(axis=-1, name="ranking_scores")
+
+        # Listwise Top 1 RankNet Loss
+        def masked_softmax(logits, mask):
+            """
+            NOTE:
+            Tried to manually compute softmax with tf operations,
```
jakemannix: why would you manually compute this with tf operations?  shouldn't we be making as much use of simple keras layers as possible?

python/ml4ir/data/tfrecord_reader.py
lastmansleeping: It didn't add any significant lag to the training step. But this is necessary. What I am doing is here is throwing an error if this assert condition fails, and when I parse the `tf.data.Dataset`, I map it through an additional `ignore_errors` stage, which thereby ignores bad records.

python/ml4ir/data/tfrecord_reader.py
lastmansleeping: No. We should. Should probably go in `FeatureConfig`. We should have a validate step. I'm going to add a `validate()` method in the `FeatureConfig` class, so that we fail fast. We will populate it with more validation as we go.

python/ml4ir/data/tfrecord_reader.py
lastmansleeping: It doesn't cover anything for now. Since, the way I define the feature spec for parsing the TFRecord is a bunch of `tf.io.VarLenFeature`, this should not be called at all for sequence features. I had it initially when I was actually getting dense features. But that was long ago. Had let the code be in case we were to get dense features in the future. But I don't think we will. Can remove.

python/ml4ir/model/architectures/dnn.py
lastmansleeping: I want to do a **very** simple config sooner rather than later. Let me take a shot.

python/ml4ir/model/losses/listwise_losses.py
lastmansleeping: How would you compute a masked softmax with "simple keras layers"? The code I have now is one possible way which I eventually landed on after multiple gradient explosions with my other approaches.

python/ml4ir/data/tfrecord_reader.py
lastmansleeping: In order to get dense features here, I need to define each feature(say MRU Affinity) as a `FixedLenFeature`(additionally specifying that fixed length/shape), which they are not - since each query has different number of records. Hence I use `tf.io.ValLenFeature`

lastmansleeping:
@jakemannix Think I've addressed most of the comments here and added a simple hardcoded model_config YAML. It is very minimal. But should suffice for now.

python/ml4ir/config/default_model_config.yaml
```diff
@@ -0,0 +1,22 @@
+architecture_key: dnn
```
jakemannix: this is great.  Exactly what I was looking for in terms of externalizing simple arch config.

python/ml4ir/data/tfrecord_reader.py
jakemannix: Got it, yes VarLenFeature because of variable record counts

python/ml4ir/model/architectures/dnn.py
```diff
@@ -1,20 +1,58 @@
...
+            else:
+                raise KeyError("Dense layer type is not supported : {}".format(layer_type))
+
+            layer_ops.append(layer_op)
+
+        return layer_ops
+
+    def get_architecture_op(self):
+        def _architecture_op(ranking_features):
+            _ = ranking_features
```
jakemannix: what does this line do?  can we just actually name our variables?

lastmansleeping: I can rename `_` to `layer_input`. Does that make sense?

jakemannix: yeah that's better.  Anything is better than "_" ;P

jakemannix: although since you're using it as a temp variable, maybe calling it just "layer" is better.  It doesn't make sense to call it anything "input" related after the beginning.

python/ml4ir/model/losses/listwise_losses.py
```diff
@@ -8,4 +12,39 @@ def _make_loss_fn(self, **kwargs):
...
+
+        def _loss_fn(y_true, y_pred):
+            batch_size = tf.cast(tf.shape(y_true)[0], tf.float32)
+
+            # Mask the padded records
+            y_true = tf.gather_nd(y_true, tf.where(tf.equal(mask, tf.constant(1.0))))
+            y_pred = tf.gather_nd(y_pred, tf.where(tf.equal(mask, tf.constant(1.0))))
+
+            # Reshape the tensors
+            y_true = tf.expand_dims(tf.squeeze(y_true), axis=-1)
```
jakemannix: what does `expand_dims(squeeze(y_true), axis=-1)` do to the shape?

python/ml4ir/model/losses/listwise_losses.py
jakemannix: huh, ok.  It's a shame that the typical way of dealing with variable sequence length tensors (i.e. for anything text related) can't be used. I'm definitely not familiar with how gradient explosions would be showing up... either you correctly compute gradients which are only on real data you have, or you accidentally compute it on "fake" records.  The latter lead to exploding gradients, you say?

python/ml4ir/model/losses/loss_base.py
```diff
@@ -29,6 +29,18 @@ def _loss_fn(y_true, y_pred):
...
+        """
+        Returns the final activation layer
+        """
+        return self._final_activation_op()
+
+    def _final_activation_op(self):
+        """
+        Define the final activation layer
+        """
+        raise NotImplementedError
```
jakemannix: shouldn't we default to identity activation, and always apply it?  Same effect, right?

Why are there two methods here?  I don't understand why there's `get_final_activation_op` and `_final_activation_op`.  What does it get us?

python/ml4ir/model/losses/pointwise_losses.py
```diff
@@ -10,18 +11,24 @@ def _make_loss_fn(self, **kwargs):
...
         return _loss_fn
+
+    def _final_activation_op(self):
+        # Pointwise sigmoid loss
+        sigmoid = layers.Activation("sigmoid", name="ranking_scores")
+
+        def sigmoid_op(logits, mask):
+            return sigmoid(logits)
+
+        return sigmoid_op
```
jakemannix: the mask is ignored here?  why is it even passed in?

python/ml4ir/model/losses/pointwise_losses.py
```diff
@@ -10,18 +11,24 @@ def _make_loss_fn(self, **kwargs):
...
+            return bce(y_true, y_pred)
 
         return _loss_fn
+
+    def _final_activation_op(self):
+        # Pointwise sigmoid loss
+        sigmoid = layers.Activation("sigmoid", name="ranking_scores")
+
+        def sigmoid_op(logits, mask):
+            return sigmoid(logits)
```
jakemannix: could the body of this method just be:

`return lambda logits, mask: layers.Activation("sigmoid", name="ranking_scores")(logits)`

python/ml4ir/model/losses/listwise_losses.py
lastmansleeping: Not really. In fact, after my entire weekend of debugging and fixing every possible gradient explosion ever, I still was seeing my loss hit `nan`. And then I gave up and used this instead.

So, loss becomes null when you have `y_pred` as 0/inf/ `y_true_i * lop(y_pred_i)`.
While computing softmax, I had to handle the bounds of `tf.float32` and negative values.
So essentially lots and lots of careful clipping.

Then I realized, I can just pass in a very small value for padded records into the softmax and they would become `0.0` that way and I also get all the stability of tensorflow's softmax with the loss functions.

python/ml4ir/model/architectures/dnn.py
lastmansleeping: It is an input to the layer, no?

python/ml4ir/model/losses/listwise_losses.py
lastmansleeping: `tf.squeeze()` is probably an overkill here as `y_true` and `y_pred`(if we have written a good model architecture) should already be 1-dimensional tensor.

Otherwise,
`tf.squeeze()` collapses empty dimensions on the tensor, so **[1, x, 1] -> [x]**
`tf.expand_dims(axis=-1)` is similar to `np.new_axis` in numpy, it changes a tensor of dimension **[x] -> [x, 1]**

python/ml4ir/model/losses/loss_base.py
lastmansleeping: I can default it to identity activation, yes.

You're right. I can merge the two methods.

python/ml4ir/model/losses/pointwise_losses.py
lastmansleeping: For consistency across different methods. For example, the softmax activation uses it.

This is how I call this from `ranking_model` -> `return activation_fn(logits, mask)`

python/ml4ir/model/losses/pointwise_losses.py
lastmansleeping: I mean... yes. But this is more readable, right?

python/ml4ir/model/losses/listwise_losses.py
jakemannix: Ok.  I thought that there were layers in keras (for text / sequences) which kinda naturally did this stuff for you...

python/ml4ir/model/ranking_model.py
```diff
@@ -571,7 +567,9 @@ def _build_callback_hooks(
                 callbacks_list.append(cp_callback)
 
             # Early Stopping
-            early_stopping_callback = callbacks.EarlyStopping(monitor="val_new_MRR", mode="max")
+            early_stopping_callback = callbacks.EarlyStopping(
+                monitor="val_new_MRR", mode="max", patience=2, verbose=1, restore_best_weights=True
```
jakemannix: yeah the "patience" param here is key when having `restore_best_weights` - because you let yourself overshoot.

Do you know if "patience" is measured in *epochs*?  If so, it this won't help much if we only do 3 epochs...

python/ml4ir/tests/test_ranking_model.py
```diff
@@ -64,8 +65,8 @@ def test_csv_and_tfrecord(self):
         )
 
         # Check if the loss and accuracy on the test set is the same
-        assert np.isclose(csv_loss, 0.62368, rtol=0.05)
-        assert np.isclose(csv_mrr, 0.564156, rtol=0.05)
+        assert np.isclose(csv_loss, 0.55835, rtol=0.05)
```
jakemannix: these magic numbers... what are we testing here?  Every time you add something new, they'll change.  How will you know if the change is good?

python/ml4ir/io/file_io.py
```diff
@@ -185,6 +186,23 @@ def read_json(infile, log=None) -> dict:
...
+    Args:
+        infile: path to the json file; can be hdfs path
+
+    Returns:
+        python dictionary
+    """
+    if infile.startswith("hdfs"):
+        raise NotImplementedError
+    else:
+        dict_ = yaml.safe_load(open(infile, "r"))
```
jakemannix: why not simply `return yaml.safe_load(open(infile, "r"))` ?

python/ml4ir/tests/test_ranking_model.py
lastmansleeping: You are right. I'm not sure. do you have any ideas for doing an end-to-end test?

python/ml4ir/model/ranking_model.py
lastmansleeping: `patience` is measured in `epochs`, yes. Should we make `patience` a function of `num_epochs`? Say 20%?

lastmansleeping: I don't know if it needs to be an argument in itself.

lastmansleeping: Also, FYI - I'm currently doing 10 epochs since I have all these built in.

python/ml4ir/model/losses/listwise_losses.py
jakemannix: sounds reasonable.

PR #10, opened 2020-03-11
Updating pom file to build java files

darshshah:


jvm/ml4ir-inference/pom.xml
```diff
@@ -15,108 +16,139 @@
...
+        <executions>
+          <execution>
+            <id>add-source</id>
+            <phase>generate-sources</phase>
+            <goals>
+              <goal>add-source</goal>
+            </goals>
+            <configuration>
+              <sources>
+                <source>src/main/java</source>
```
darshshah: This is the main addition

jakemannix: cool will pull it down and build!

jakemannix:
> Thanks for the contribution! It looks like @darshshah is an internal user so signing the CLA is not required. However, [we need to confirm this](https://docs.google.com/forms/d/e/1FAIpQLSfB_vIrwFLweXIGt7ID98JhISCU2zm2ErOugAnGAcCvAzr7NQ/viewform).

@darshshah  did you follow that link and/or resolve the "open source CLA" so github will allow your contribution here?

jakemannix:
looks like it fixes the issue alright:

```ansible|jmannix-ltm1:jvm jmannix$ jar -tf ml4ir-inference/target/ml4ir-inference-0.0.1-SNAPSHOT.jar  | grep ava
ml4ir/inference/tensorflow/SequenceExampleJavaBuilder.class```

PR #12, opened 2020-03-13
Character embeddings + Cleaning up context feature tiling

lastmansleeping:


python/ml4ir/model/ranking_model.py
```diff
@@ -8,10 +8,12 @@
 from tensorflow import TensorSpec, TensorArray
 from tensorflow import data
 from tensorflow.keras import metrics as kmetrics
+from tensorflow.keras import backend as kbackend
```
jakemannix: don't people often abbreviate this as simply `K`, conventionally?

`from tensorflow.keras import backend as K`

Not sure why... but seems standardish

python/ml4ir/model/ranking_model.py
```diff
@@ -416,6 +417,16 @@ def _predict_score(features, label):
                             "{} was not found in input training data".format(feature_name)
                         )
 
+                # Explode context features to each record for logging
```
jakemannix: "for logging" only?

python/ml4ir/model/ranking_model.py
lastmansleeping: Yeah, good call.

python/ml4ir/model/ranking_model.py
lastmansleeping: So these are metadata fields mostly. But even for things like query text, we explode it here so that we can log. The actual LSTM would be trained in the feature layer on a single context feature per query and then tiled for each record in that query.

python/ml4ir/config/features.py
```diff
@@ -259,11 +259,16 @@ def define_inputs(self, max_num_records: int) -> Dict[str, Input]:
             Dictionary of tensorflow graph input nodes
         """
 
-        def get_shape(feature_layer_info: dict):
+        def get_shape(feature_info: dict):
+            feature_layer_info = feature_info["feature_layer_info"]
+            if feature_info["tfrecord_type"] == TFRecordTypeKey.CONTEXT:
+                num_records = 1
```
lastmansleeping: Doing this so that we don't explode all context features by default.

lastmansleeping: ... and define the right input shape.

python/ml4ir/model/optimizer.py
```diff
@@ -18,14 +22,14 @@ def get_optimizer(
     )
 
     if optimizer_key == OptimizerKey.ADAM:
-        return tf_optimizers.Adam(learning_rate=learning_rate_schedule)
+        return tf_optimizers.Adam(learning_rate=learning_rate_schedule, clipvalue=CLIP_VALUE)
```
lastmansleeping: This helps handle gradient explosions. I've set it to a value that I found to be fairly stable from my experiments. But, we will have to fine tune this as a hyperparameter at some point.

python/ml4ir/model/ranking_model.py
```diff
@@ -721,12 +732,47 @@ def _get_dense_feature(inputs, feature, shape=(1,)):
...
+                            layers.LSTM(
+                                int(feature_layer_info["embedding_size"] / 2),
+                                return_sequences=False,
+                            ),
+                            merge_mode="concat",
+                        )(input_feature)
+                        if feature_info["tfrecord_type"] == TFRecordTypeKey.CONTEXT:
+                            # If feature is a context feature then tile it for all records
+                            dense_feature = tf.expand_dims(dense_feature, axis=1)
+                            dense_feature = kbackend.repeat_elements(
```
lastmansleeping: Just an FYI. Doing first the BiLSTM and then the tile

python/ml4ir/tests/data/csv/feature_config.yaml
```diff
@@ -76,13 +76,15 @@ features:
     tfrecord_type: sequence
   - name: query_str
     node_name: query_str
-    trainable: false
+    trainable: true
```
lastmansleeping: And voila!

python/ml4ir/model/ranking_model.py
lastmansleeping: You are looking at the inference code here. Where I need to convert each TFRecord back to a log-able CSV(or a pandas dataframe) with one record per row.

python/ml4ir/model/optimizer.py
jakemannix: but should be passed in, right?  Not a global constant?

python/ml4ir/model/ranking_model.py
```diff
@@ -732,24 +732,47 @@ def _get_dense_feature(inputs, feature, shape=(1,)):
                     inputs, feature_node_name, shape=(self.max_num_records, 1)
                 )
                 if feature_info["trainable"]:
-                    ranking_features.append(tf.cast(dense_feature, tf.float32))
+                    dense_feature = tf.expand_dims(tf.cast(dense_feature, tf.float32), axis=-1)
```
jakemannix: this `_get_dense_feature` function is pretty... convoluted to unit test, yeah?

python/ml4ir/model/ranking_model.py
jakemannix: yep.  but probably extract this out to another method.

lastmansleeping: Yes, I am planning on doing that soon. Letting some models train for now.

python/ml4ir/model/optimizer.py
lastmansleeping: Yes, for sure. Will do that.

python/ml4ir/data/tfrecord_reader.py
```diff
@@ -49,6 +49,22 @@ def make_parse_fn(feature_config: FeatureConfig, max_num_records: int = 25) -> t
 
     punctuation_regex = "|".join([re.escape(c) for c in list(string.punctuation)])
 
+    @tf.function
```
jakemannix: you should annotate what the expected input and output shapes and types are.  And we should have a unit test verifying it. ;)

python/ml4ir/data/tfrecord_reader.py
```diff
@@ -49,6 +49,22 @@ def make_parse_fn(feature_config: FeatureConfig, max_num_records: int = 25) -> t
...
+    def _preprocess_text(feature_tensor, preprocessing_info):
+        # TODO: Move this to a separate function
```
jakemannix: and yes, this should be a top-level function somewhere :)

lastmansleeping:
@jakemannix Yes, I learned about this coding pattern when mypy sent me insane recently. This is a good place to start using it. Good idea. But as I said, I'm making small non-breaking changes at the moment, because I am training models in the background.

lastmansleeping:
> can't you just do
> 
> ```
> dtype = feature_info["dtype"]
> if dtype == tf.float32:
>   default_value = 0.0
> ...
> ```

Now that our dtypes are actual tf types, I can, yes!

jakemannix:
>  But as I said, I'm making small non-breaking changes at the moment, because I am training models in the background.

yep - consider suggestions for once you have models and are changing code again. :)

lastmansleeping:
> this is exactly the kind of check we ned lots of.
> 
> Incidentally, do we think that "get_sequence_embedding" should really be a "private" function, given this? Remember, we're making a library here.

I don't think it should. Writing a unit test made me think about this as well. Will change.

lastmansleeping:
> seems like there should be a specific check that after processing, this string becomes "abcabc123", right?

I'm kinda doing it in the final `assert`. But yeah, I can make it explicit.

python/ml4ir/tests/test_features.py
```diff
@@ -0,0 +1,74 @@
...
+    def test_sequence_embedding(self):
+        """
+        Unit test sequence embedding
+
+        """
+        batch_size = 50
+        max_length = 20
+        embedding_size = 128
+        feature_info = {
+            "feature_layer_info": {"embedding_size": embedding_size, "embedding_type": "bilstm"},
```
jakemannix: so I'm missing a config piece here, and terminology: the "embedding" size should be the dimension of the character / token Embedding layer right after one-hot encoding (unless that is not used, but still: first step of the chain).  For us, that should be 256 (since we're not hashing, are we?).  Then there's the lstm output dimension.  The output of the bilstm will be 2 * this value.  

This value is not the "embedding" dimension, it's independent, and should be called the "text_encoding" dimension or something like that.

lastmansleeping: Updated the implementation.

lastmansleeping:
@jakemannix Can I merge this PR?

python/ml4ir/model/optimizer.py
```diff
@@ -3,6 +3,10 @@
 from tensorflow.keras.optimizers.schedules import ExponentialDecay
 
 
+# Constants
+CLIP_VALUE = 5.0
+
+
 def get_optimizer(
```
jakemannix: pass in `clipvalue=CLIP_VALUE` here, and use it for all inside?

python/ml4ir/config/parse_args.py
```diff
@@ -203,10 +203,17 @@ def define_args() -> ArgumentParser:
...
         type=int,
-        default=50,
+        default=25,
         help="Minimum number of queries per group to be used to computed groupwise metrics.",
     )
 
+    parser.add_argument(
+        "--gradient_clip_value",
+        type=float,
+        default=5.0,
```
jakemannix: this 5.0 should just reference a constant in the optimizers module, perhaps?

python/ml4ir/model/optimizer.py
lastmansleeping: Wait. I think you are looking at outdated code. I've fixed this with reading clipvalue as an arg.

python/ml4ir/config/parse_args.py
lastmansleeping: Hmm. That way we have the optimizer module kinda self-contained and easily readable?

python/ml4ir/model/optimizer.py
jakemannix: ok I'm walking through the PR in full right now!

lastmansleeping: Cool. I will stop replying for a while.

python/ml4ir/features/feature_layer.py
```diff
@@ -0,0 +1,107 @@
...
+    NOTE: Can remove this in the future and
+          pass inputs[feature] directly
+    """
+    feature_col = feature_column.numeric_column(feature, shape=shape)
+    dense_feature = layers.DenseFeatures(feature_col)(inputs)
+
+    return dense_feature
+
+
+def get_sequence_embedding(input, feature_info, max_num_records):
```
jakemannix: `get_sequence_encoding`?  since not always an embedding

python/ml4ir/model/optimizer.py
```diff
@@ -8,6 +8,7 @@ def get_optimizer(
     learning_rate: float,
     learning_rate_decay: float,
     learning_rate_decay_steps: int,
+    gradient_clip_value: float,
```
jakemannix: got this!

python/ml4ir/tests/data/tfrecord/feature_config.yaml
```diff
@@ -76,21 +76,25 @@ features:
...
+    trainable: true
+    dtype: string
     log_at_inference: true
     feature_layer_info:
       type: string
-      max_length: 20
       shape: null
+      encoding_type: bilstm
+      encoding_size: 128
+      embedding_size: 128
```
jakemannix: default embedding should probably be way lower, right?  64 or even 32?

jakemannix: for a unit test, even tinier ;)

jakemannix: but really it doesn't matter, for a unit test. nevermind!

python/ml4ir/tests/test_metrics.py
```diff
@@ -1,17 +1,20 @@
...
 import numpy as np
 
 
 # Constants
 GOLD_METRICS = {
-    "old_MRR": 0.7806604,
-    "new_MRR": 0.54550856,
     "old_ACR": 1.6669034,
-    "new_ACR": 2.549716,
+    "new_ACR": 2.3998579,
```
jakemannix: aiyiyi. what value is verifying that we get -22% "improvement" in MRR?

python/ml4ir/tests/test_ranking_model.py
```diff
@@ -85,9 +94,9 @@ def test_csv_and_tfrecord(self):
         )
 
         # Check if the loss and accuracy on the test set is the same
-        assert np.isclose(tfrecord_loss, 0.55004, rtol=0.05)
-        assert np.isclose(tfrecord_mrr, 0.70067, rtol=0.05)
+        assert np.isclose(tfrecord_loss, 0.59263, rtol=0.01)
```
jakemannix: oh look. it changed.  How am I supposed to review that this new value is "the right" one?  If someone else issues a PR, and this test fails because now the loss is 0.573, do they just update this test to say we now expect 0.573?  What are we even checking here?

lastmansleeping: Haha. I don't know how else to check if the end to end pipeline works though. We will remove this once we have more unit tests added. Deal!

python/ml4ir/tests/test_metrics.py
lastmansleeping: It is less about the improvement and more about being able to compute the metrics. This is not training a model at all. It just does a forward pass and computes metrics. This in all its weirdness, could probably qualify as an unit test if we bend the rules a bit.

python/ml4ir/features/feature_layer.py
lastmansleeping: Will change.

python/ml4ir/tests/test_ranking_model.py
jakemannix: I'm just not even sure these asserts are asserting anything of value.  If these three lines were replaced with `assert np.greaterthan(tfrecord_loss, 0.0)`, would you get the same information (i.e that the code made it to this point)?

lastmansleeping:
> so another way to avoid big if/switch statements is to build a `Dict[str, Callable]`:
> 
> ```
> feature_fns = {
>  tf.string: _bytes_feature,
>  tf.float32: _float_feature,
>  tf.int64: _int64_feature
> }
> 
> def _get_feature_fn(dtype):
>   try:
>     return feature_fns[dtype]
>   except KeyError:
>     raise Exception("Feature dtype {} not supported".format(dtype))
> ```
> 
> not sure if that's better in this case, but in general giant long if/else blocks... smell yucky. :)

Also, I forgot to tell you. I tried doing this. But we end up with a weird scenario where when `tf.int64` is used as a dict key, then it only matches with `int64` and not with `int`. Same with `float`. :/

jakemannix:
> > so another way to avoid big if/switch statements is to build a `Dict[str, Callable]`:
> > ```
> > feature_fns = {
> >  tf.string: _bytes_feature,
> >  tf.float32: _float_feature,
> >  tf.int64: _int64_feature
> > }
> > 
> > def _get_feature_fn(dtype):
> >   try:
> >     return feature_fns[dtype]
> >   except KeyError:
> >     raise Exception("Feature dtype {} not supported".format(dtype))
> > ```
> > 
> > 
> > not sure if that's better in this case, but in general giant long if/else blocks... smell yucky. :)
> 
> Also, I forgot to tell you. I tried doing this. But we end up with a weird scenario where when `tf.int64` is used as a dict key, then it only matches with `int64` and not with `int`. Same with `float`. :/

right, but I think we *want* to be explicit: in jvm languages you're going to confuse people if you say "int" means a 64bit int.  That's a "long".  Ideally we literally always be explicit, if it's possible.

Unless we'll run against the weirdness that TF itself isn't consistent between java and python :\

lastmansleeping:
> > > so another way to avoid big if/switch statements is to build a `Dict[str, Callable]`:
> > > ```
> > > feature_fns = {
> > >  tf.string: _bytes_feature,
> > >  tf.float32: _float_feature,
> > >  tf.int64: _int64_feature
> > > }
> > > 
> > > def _get_feature_fn(dtype):
> > >   try:
> > >     return feature_fns[dtype]
> > >   except KeyError:
> > >     raise Exception("Feature dtype {} not supported".format(dtype))
> > > ```
> > > 
> > > 
> > > not sure if that's better in this case, but in general giant long if/else blocks... smell yucky. :)
> > 
> > 
> > Also, I forgot to tell you. I tried doing this. But we end up with a weird scenario where when `tf.int64` is used as a dict key, then it only matches with `int64` and not with `int`. Same with `float`. :/
> 
> right, but I think we _want_ to be explicit: in jvm languages you're going to confuse people if you say "int" means a 64bit int. That's a "long". Ideally we literally always be explicit, if it's possible.
> 
> Unless we'll run against the weirdness that TF itself isn't consistent between java and python :\

Firstly, this still will not work because JVM expects `float` but `tf.float32` dict key matches *only* with `"float32"`. 

Secondly, I think the python training module should not be this limited(expecting it to be explicit) by the JVM, especially when we are using tf compatible types already. Would you not want to train a model if you specified `float32` instead of `float`? Because in tensorflow python, the type is called `float32` and not `float`. It would just be confusing for potential users of the package to have to use `float`(this is a tf explicit type that is not compatible with JVM) for one and `int64`(this is a tf explicit type compatible with JVM) for the other, if we choose to build it so closely coupled with JVM code.
Ref: https://www.tensorflow.org/api_docs/python/tf/dtypes/DType

python/ml4ir/tests/test_ranking_model.py
lastmansleeping: What I was hoping to check was... If someone set up `ml4ir` on their local box, this is a quick way to validate that the results are not too far off from what we have observed. I have fixed the data, the seeds, etc.

I can do a `np.greaterthan..` as well. But that would be a softer check.

jakemannix:
> > > > so another way to avoid big if/switch statements is to build a `Dict[str, Callable]`:
> > > > ```
> > > > feature_fns = {
> > > >  tf.string: _bytes_feature,
> > > >  tf.float32: _float_feature,
> > > >  tf.int64: _int64_feature
> > > > }
> > > > 
> > > > def _get_feature_fn(dtype):
> > > >   try:
> > > >     return feature_fns[dtype]
> > > >   except KeyError:
> > > >     raise Exception("Feature dtype {} not supported".format(dtype))
> > > > ```
> > > > 
> > > > 
> > > > not sure if that's better in this case, but in general giant long if/else blocks... smell yucky. :)
> > > 
> > > 
> > > Also, I forgot to tell you. I tried doing this. But we end up with a weird scenario where when `tf.int64` is used as a dict key, then it only matches with `int64` and not with `int`. Same with `float`. :/
> > 
> > 
> > right, but I think we _want_ to be explicit: in jvm languages you're going to confuse people if you say "int" means a 64bit int. That's a "long". Ideally we literally always be explicit, if it's possible.
> > Unless we'll run against the weirdness that TF itself isn't consistent between java and python :\
> 
> Firstly, this still will not work because JVM expects `float` but `tf.float32` dict key matches _only_ with `"float32"`.
> 
> Secondly, I think the python training module should not be this limited(expecting it to be explicit) by the JVM, especially when we are using tf compatible types already. Would you not want to train a model if you specified `float32` instead of `float`? Because in tensorflow python, the type is called `float32` and not `float`. It would just be confusing for potential users of the package to have to use `float`(this is a tf explicit type that is not compatible with JVM) for one and `int64`(this is a tf explicit type compatible with JVM) for the other, if we choose to build it so closely coupled with JVM code.
> Ref: https://www.tensorflow.org/api_docs/python/tf/dtypes/DType

ok yes, *because* TF is not internally consistent between languages, you've got it set up correctly.  My suggestion simply can't work here.  I'm still worried that java people will think "int" means 32bit, but we are fine.

python/ml4ir/tests/test_ranking_model.py
jakemannix: > If someone set up ml4ir on their local box, this is a quick way to validate that the results are not too far off from what we have observed. I have fixed the data, the seeds, etc.

Right, so how could this ever fail?  This "hard" check has to pass, right?  I guess it's a "point in time" test: any refactoring you do which leads to this test passing is basically guaranteed to have the same underlying ML under the hood, I guess?

But any change in which you *expect* the ML to change, even just slightly, will lead to this test failing, and the only fix would be to update the test with the new observed value, and "declare" that "this is the newly fixed Gold Standard", right? :\

lastmansleeping: My assumption was that it would fail if the right dependencies(or the right versions) weren't set up. But you are right about the second part - any small change to the actual ML side of the code will result in this value changing. :/

PR #16, opened 2020-03-25
Defining a serving signature with inference specific parsing_fn

lastmansleeping:
Things covered in the PR:
- create dummy values for fields that are not required
- not pad records for inference
- regenerate the models with updated serving signature(without retraining)

### DIFF
--- a/python/ml4ir/config/parse_args.py
+++ b/python/ml4ir/config/parse_args.py
@@ -223,6 +223,13 @@ def define_args() -> ArgumentParser:
         "If that is not the case, then you can still use a SavedModel from a model_file for inference/evaluation only",
     )
 
+    parser.add_argument(
+        "--pad_records_at_inference",
+        type=bool,
+        default=False,
+        help="Whether to pad records at inference time. Used to define the TFRecord serving signature in the SavedModel",
+    )
+
     return parser
 
 
--- a/python/ml4ir/config/parse_args.py
+++ b/python/ml4ir/config/parse_args.py
@@ -214,6 +214,13 @@ def define_args() -> ArgumentParser:
         help="Gradient clipping value/threshold for the optimizer.",
     )
 
+    parser.add_argument(
+        "--compile_keras_model",
+        type=bool,
+        default=False,
+        help="Whether to compile a loaded SavedModel into a Keras model. NOTE: This requires ",
+    )
+
     return parser
 
 
--- a/python/ml4ir/config/parse_args.py
+++ b/python/ml4ir/config/parse_args.py
@@ -218,7 +218,9 @@ def define_args() -> ArgumentParser:
         "--compile_keras_model",
         type=bool,
         default=False,
-        help="Whether to compile a loaded SavedModel into a Keras model. NOTE: This requires ",
+        help="Whether to compile a loaded SavedModel into a Keras model. "
+        "NOTE: This requires that the SavedModel's architecture, loss, metrics, etc are the same as the RankingModel"
+        "If that is not the case, then you can still use a SavedModel from a model_file for inference/evaluation only",
     )
 
     return parser
--- a/python/ml4ir/data/tfrecord_helper.py
+++ b/python/ml4ir/data/tfrecord_helper.py
@@ -1,0 +1,64 @@
+from tensorflow import train
+import tensorflow as tf
+from ml4ir.config.keys import TFRecordTypeKey
+
+
+def _bytes_feature(values):
+    """Returns a bytes_list from a string / byte."""
+    values = [value.encode("utf-8") for value in values]
+    return train.Feature(bytes_list=train.BytesList(value=values))
+
+
+def _float_feature(values):
+    """Returns a float_list from a float / double."""
+    return train.Feature(float_list=train.FloatList(value=values))
+
+
+def _int64_feature(values):
+    """Returns an int64_list from a bool / enum / int / uint."""
+    return train.Feature(int64_list=train.Int64List(value=values))
+
+
+def _get_feature_fn(dtype):
+    """Returns appropriate feature function based on datatype"""
+    if dtype == tf.string:
+        return _bytes_feature
+    elif dtype == tf.float32:
+        return _float_feature
+    elif dtype == tf.int64:
+        return _int64_feature
+    else:
+        raise Exception("Feature dtype {} not supported".format(dtype))
+
+
+def get_sequence_example_proto(group, context_features, sequence_features):
+    """
+    Get a sequence example protobuf from a dataframe group
+
+    Args:
+        - group: pandas dataframe group
+        - context_features: feature configuration for context
+        - sequence_features: feature configuration for sequence
+    """
+    sequence_features_dict = dict()
+    context_features_dict = dict()
+
+    for feature_info in context_features:
+        feature_name = feature_info["name"]
+        feature_fn = _get_feature_fn(feature_info["dtype"])
+        context_features_dict[feature_name] = feature_fn([group[feature_name].tolist()[0]])
+
+    for feature_info in sequence_features:
+        feature_name = feature_info["name"]
+        feature_fn = _get_feature_fn(feature_info["dtype"])
+        if feature_info["tfrecord_type"] == TFRecordTypeKey.SEQUENCE:
+            sequence_features_dict[feature_name] = train.FeatureList(
+                feature=[feature_fn(group[feature_name].tolist())]
+            )
+
+    sequence_example_proto = train.SequenceExample(
+        context=train.Features(feature=context_features_dict),
+        feature_lists=train.FeatureLists(feature_list=sequence_features_dict),
+    )
+
+    return sequence_example_proto
--- a/python/ml4ir/data/tfrecord_reader.py
+++ b/python/ml4ir/data/tfrecord_reader.py
@@ -114,45 +114,6 @@ def make_parse_fn(
             feature_tensor = sequence_features.get(feature_node_name, default_tensor)
 
             if isinstance(feature_tensor, sparse.SparseTensor):
-                if feature_node_name == feature_config.get_rank(key="node_name"):
-                    # Add mask for identifying padded records
-                    mask = tf.ones_like(sparse.to_dense(sparse.reset_shape(feature_tensor)))
-                    mask = tf.expand_dims(mask, axis=2)
-
-                    def crop_fn():
-                        tf.print(
-                            "\n[WARN] Bad query found. Number of records : ", tf.shape(mask)[1]
-                        )
-                        return image.crop_to_bounding_box(
-                            mask,
-                            offset_height=0,
-                            offset_width=0,
-                            target_height=1,
-                            target_width=max_num_records,
-                        )
-
-                    mask = tf.cond(
-                        tf.shape(mask)[1] < max_num_records,
-                        # Pad if there are missing records
-                        lambda: image.pad_to_bounding_box(
-                            mask,
-                            offset_height=0,
-                            offset_width=0,
-                            target_height=1,
-                            target_width=max_num_records,
-                        ),
-                        # Crop if there are extra records
-                        crop_fn,
-                    )
-                    mask = tf.squeeze(mask)
-
-                    # Check validity of mask
-                    tf.debugging.assert_greater(
-                        tf.cast(tf.reduce_sum(mask), tf.float32), tf.constant(0.0)
-                    )
-
-                    features_dict["mask"] = mask
-
                 feature_tensor = sparse.reset_shape(feature_tensor, new_shape=[1, max_num_records])
                 feature_tensor = sparse.to_dense(feature_tensor)
                 feature_tensor = tf.squeeze(feature_tensor)
@@ -165,24 +126,67 @@ def make_parse_fn(
 
             features_dict[feature_node_name] = feature_tensor
 
-        """
-        Define dummy mask if the rank field is not a required field for serving
-
-        NOTE:
-        This masks all max_num_records as 1 as there is no real way to know
-        the number of records in the query. There is no predefined required field,
-        and hence we would need to do a full pass of all features to find the record shape.
-        This approach might be unstable if different features have different shapes.
-
-        Hence we just mask all records
-
-        TODO: Might be beneficial to define one mandatory required sequence feature,
-        such as record_key and use that to define the mask. Alternatively, we could
-        make rank a mandatory required feature, but a dummy value would have to be
-        passed to rank at inference time.
-        """
+        # Define mask to identify padded records
         if required_only and not feature_config.get_rank("serving_info")["required"]:
+            """
+            Define dummy mask if the rank field is not a required field for serving
+
+            NOTE:
+            This masks all max_num_records as 1 as there is no real way to know
+            the number of records in the query. There is no predefined required field,
+            and hence we would need to do a full pass of all features to find the record shape.
+            This approach might be unstable if different features have different shapes.
+
+            Hence we just mask all records
+
+            TODO: Might be beneficial to define a mandatory required sequence feature,
+            such as record_key and use that to define the mask. Alternatively, we could
+            make rank a mandatory required feature, but a dummy value would have to be
+            passed to rank at inference time.
+            """
             features_dict["mask"] = tf.constant(value=1, shape=[max_num_records], dtype=tf.int64)
+            features_dict["num_records"] = tf.constant(max_num_records, dtype=tf.int64)
+        else:
+            # Typically used at training time, to pad/clip to a fixed number of records per query
+
+            # Use rank as a reference tensor to infer shape/num_records in query
+            reference_tensor = sequence_features.get(feature_config.get_rank(key="node_name"))
+
+            # Add mask for identifying padded records
+            mask = tf.ones_like(sparse.to_dense(sparse.reset_shape(reference_tensor)))
+            mask = tf.expand_dims(mask, axis=2)
+
+            def crop_fn():
+                tf.print("\n[WARN] Bad query found. Number of records : ", tf.shape(mask)[1])
+                return image.crop_to_bounding_box(
+                    mask,
+                    offset_height=0,
+                    offset_width=0,
+                    target_height=1,
+                    target_width=max_num_records,
+                )
+
+            mask = tf.cond(
+                tf.shape(mask)[1] < max_num_records,
+                # Pad if there are missing records
+                lambda: image.pad_to_bounding_box(
+                    mask,
+                    offset_height=0,
+                    offset_width=0,
+                    target_height=1,
+                    target_width=max_num_records,
+                ),
+                # Crop if there are extra records
+                crop_fn,
+            )
+            mask = tf.squeeze(mask)
+            num_records = tf.cast(tf.reduce_sum(mask), tf.int64)
+
+            # Check validity of mask
+            tf.debugging.assert_greater(num_records, tf.constant(0, dtype=tf.int64))
+
+            features_dict["mask"] = mask
+            features_dict["num_records"] = tf.reduce_sum(mask)
 
         labels = features_dict.pop(feature_config.get_label(key="name"))
 
--- a/python/ml4ir/data/tfrecord_reader.py
+++ b/python/ml4ir/data/tfrecord_reader.py
@@ -31,7 +31,7 @@ def make_parse_fn(
     feature_config: FeatureConfig,
     max_num_records: int = 25,
     required_only: bool = False,
-    pad_records: bool = False,
+    pad_records: bool = True,
 ) -> tf.function:
     """
     Create a parse function using the context and sequence features spec
--- a/python/ml4ir/data/tfrecord_reader.py
+++ b/python/ml4ir/data/tfrecord_reader.py
@@ -16,17 +16,6 @@ data in the train.SequenceExample protobuf format
 """
 
 
-def _get_default_value(dtype, serving_info):
-    if dtype == tf.float32:
-        return serving_info.get("default_value", 0.0)
-    elif dtype == tf.int64:
-        return serving_info.get("default_value", 0)
-    elif dtype == tf.string:
-        return serving_info.get("default_value", "")
-    else:
-        raise Exception("Unknown dtype {}".format(dtype))
-
-
 def make_parse_fn(
     feature_config: FeatureConfig,
     max_num_records: int = 25,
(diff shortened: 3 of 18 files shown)


lastmansleeping:
I've been trying different techniques for a while and I was able filter out fields that are not required at serving time. This was weirdly way more complicated than I thought.

But what is even more complicated is the ability to run inference on a dynamic number of records , a.k.a, skipping padding. This is my conclusion after trying multiple different approaches and storming my brain:

@jakemannix I need at least one mandatory sequence feature as a contract between inference and training to infer the shape of the query. There is no way to know the shape of a query(the number of records) otherwise. Currently, I use `rank` field to generate the mask and thereby pad accordingly. But since `rank` could and most probably will be a `required=False` field at inference, it makes things difficult. 

I have two options:
1. Pass dummy int64 values to the `rank` field at inference time.
2. Define a "mandatory `required=False` at all times" sequence field, say "record_key" or "record_id", and we make sure to pass in some value at all times.

Given this, I still have to figure out how to make my model layers behave nicely with non hard coded `max_num_records`, which has also been some pain.

lastmansleeping:
There is also Option 3: We infer num_records from a random(but fixed at training time) input trainable feature(which will be `required=True` by default). This might be a tiny bit unstable because input features can be of any shapes and size by themselves as opposed to something predefined like `record_id` or `rank`. But should be okay for now and is also an option if we want to not have any dummy fields being passed at inference.

python/ml4ir/config/parse_args.py
```diff
@@ -214,6 +214,13 @@ def define_args() -> ArgumentParser:
         help="Gradient clipping value/threshold for the optimizer.",
     )
 
+    parser.add_argument(
+        "--compile_keras_model",
+        type=bool,
+        default=False,
+        help="Whether to compile a loaded SavedModel into a Keras model. NOTE: This requires ",
```
jakemannix: this requires... what? :)

python/ml4ir/data/tfrecord_reader.py
```diff
@@ -9,39 +9,56 @@
...
 This module contains helper methods for reading and writing
 data in the train.SequenceExample protobuf format
 """
 
 
-def make_parse_fn(feature_config: FeatureConfig, max_num_records: int = 25) -> tf.function:
-    """Create a parse function using the context and sequence features spec"""
+def _get_default_value(dtype, serving_info):
+    if dtype == tf.float32:
+        return serving_info.get("default_value", 0.0)
```
jakemannix: we should make a note/TODO to make sure we track whether it matters to fill in with values which are valid inputs (instead of known sentinel values which will hopefully never be input).  It may affect per-feature normalization, too?

jakemannix:
> There is also Option 3: We infer num_records from a random(but fixed at training time) input trainable feature(which will be `required=True` by default). This might be a tiny bit unstable because input features can be of any shapes and size by themselves as opposed to something predefined like `record_id` or `rank`. But should be okay for now and is also an option if we want to not have any dummy fields being passed at inference.

let's discuss.  It should be fine to fix some field like "initial_rank" (meaning "the rank as provided by an underlying, lower-level, ranking system [L1 or L2]") which will always exist, in every dataset.

python/ml4ir/config/parse_args.py
lastmansleeping: I think it probably got deleted at some point. But great catch!

python/ml4ir/data/tfrecord_reader.py
lastmansleeping: Right now you can define defaults in the feature_config YAML, as part of the serving_info. If that is absent, I use preset default based on the dtype. 
We do not have any normalization yet other than batch norm, but we might in the future, yes.

lastmansleeping: Is there something else you were referring to with the TODO?

lastmansleeping:
@calliebradley Tagging you on the PR so that you can shadow Jake. Let me know if you have questions. I should do a deep dive/ code walkthrough for the team in the coming weeks.

python/ml4ir/data/tfrecord_reader.py
```diff
@@ -9,39 +9,45 @@
...
 data in the train.SequenceExample protobuf format
 """
 
 
-def make_parse_fn(feature_config: FeatureConfig, max_num_records: int = 25) -> tf.function:
-    """Create a parse function using the context and sequence features spec"""
+def make_parse_fn(
+    feature_config: FeatureConfig,
+    max_num_records: int = 25,
+    required_only: bool = False,
```
lastmansleeping, line 22: @jakemannix You can now essentially pass `required_only`(parse only required features) and `pad_records`(pad records for training) as arguments to the TFRecord parse_fn.

python/ml4ir/data/tfrecord_reader.py
```diff
@@ -79,72 +88,102 @@ def _parse_sequence_example_fn(sequence_example_proto):
...
+            Hence we just mask all records
+            """
+            features_dict["mask"] = tf.constant(
+                value=1, shape=[max_num_records], dtype=feature_config.get_rank("dtype")
+            )
+            num_records = tf.constant(max_num_records, dtype=tf.int64)
+        else:
+            # Typically used at training time, to pad/clip to a fixed number of records per query
+
+            # Use rank as a reference tensor to infer shape/num_records in query
```
lastmansleeping, line 111: @jakemannix I use `rank` as a reference tensor to infer the number of records in the query and mask it.

python/ml4ir/data/tfrecord_reader.py
```diff
@@ -79,72 +88,102 @@ def _parse_sequence_example_fn(sequence_example_proto):
...
+            num_records = max_num_records if pad_records else num_records
+
         # Pad sequence features to max_num_records
         for feature_info in feature_config.get_sequence_features():
             feature_node_name = feature_info.get("node_name", feature_info["name"])
             feature_layer_info = feature_info["feature_layer_info"]
             preprocessing_info = feature_info.get("preprocessing_info", {})
 
-            feature_tensor = sequence_features.get(feature_node_name)
+            default_tensor = tf.fill(
```
lastmansleeping, line 160: @jakemannix For fields that were not passed in the proto, we fill it with their default value. This would be fields that not `required`.

python/ml4ir/data/tfrecord_reader.py
```diff
@@ -79,72 +88,102 @@ def _parse_sequence_example_fn(sequence_example_proto):
...
 
             features_dict[feature_node_name] = feature_tensor
 
         labels = features_dict.pop(feature_config.get_label(key="name"))
 
-        # Check if label is one-hot and correctly masked
-        tf.debugging.assert_equal(tf.cast(tf.reduce_sum(labels), tf.float32), tf.constant(1.0))
+        if not required_only:
+            # Check if label is one-hot and correctly masked
+            tf.debugging.assert_equal(tf.cast(tf.reduce_sum(labels), tf.float32), tf.constant(1.0))
```
lastmansleeping, line 186: @jakemannix At training time, I've still retained the assert to check if at least one record was clicked.

python/ml4ir/features/feature_config.py
```diff
@@ -252,24 +256,39 @@ def get_group_metrics_keys(self, key: str = None):
         """
         return self._get_list_of_keys_or_dicts(self.group_metrics_keys, key=key)
 
-    def define_inputs(self, max_num_records: int) -> Dict[str, Input]:
+    def get_dtype(self, feature_info: dict):
```
lastmansleeping, line 259: @jakemannix Defining `get_dtype` and `get_default_value` under FeatureConfig. I've explicitly assigned dtypes to each tensor from this PR. This is necessary since I have to do some dynamic slicing on tensors later on. We should have always been doing this anyways.

python/ml4ir/features/feature_config.py
```diff
@@ -294,12 +317,41 @@ def generate_mask(self):
...
             "name": "mask",
             "trainable": False,
-            "dtype": "float",
+            "dtype": self.get_rank("dtype"),
             "feature_layer_info": {"type": FeatureTypeKey.NUMERIC, "shape": None},
             "serving_info": {"name": "mask", "required": False},
             "tfrecord_type": TFRecordTypeKey.SEQUENCE,
         }
 
+    def create_dummy_sequence_example(self, num_records=1, required_only=False):
```
lastmansleeping, line 326: @jakemannix Added functionality to create a dummy sequence example with given number of records.

python/ml4ir/features/feature_layer.py
```diff
@@ -62,15 +48,19 @@ def feature_layer(inputs):
         ranking_features = list()
         metadata_features = dict()
 
+        numeric_tile_shape = tf.shape(tf.expand_dims(tf.gather(inputs["mask"], indices=0), axis=0))
```
lastmansleeping, line 51: @jakemannix This is the trickiest part - figuring out the dynamic tile shape.

python/ml4ir/data/tfrecord_reader.py
jakemannix: Were we wanting to do this? I’ve been interpreting  ‘required’ as meaning the SequenceExample has to have the feature.  You should possibly be doing an assert() that the field is already filled in, not fill it in yourself, right?

python/ml4ir/features/feature_config.py
jakemannix: Would be nice to allow passing in of a dict of feature values to put in the SequenceExample :)

python/ml4ir/data/tfrecord_reader.py
lastmansleeping: This is for fields that the model needs, say something like `organization_id`, but which is not a required field and will NOT be in the input proto at inference.

I should do an assert though. That might be a good way to check we are not silently failing.

python/ml4ir/config/parse_args.py
```diff
@@ -214,6 +214,22 @@ def define_args() -> ArgumentParser:
         help="Gradient clipping value/threshold for the optimizer.",
     )
 
+    parser.add_argument(
+        "--compile_keras_model",
+        type=bool,
+        default=False,
+        help="Whether to compile a loaded SavedModel into a Keras model. "
```
jakemannix, line 221: What does this mean, exactly?

jakemannix: and what is the purpose?

python/ml4ir/data/tfrecord_helper.py
```diff
@@ -0,0 +1,64 @@
+from tensorflow import train
+import tensorflow as tf
+from ml4ir.config.keys import TFRecordTypeKey
+
+
+def _bytes_feature(values):
```
jakemannix, line 6: are these moved from elsewhere?

python/ml4ir/features/feature_config.py
lastmansleeping: This story has already been so far away from 1SP, I don't want to scope creep it. I will add a new work item to do this if that's fine with you.

python/ml4ir/config/parse_args.py
lastmansleeping: This is the key really. "Compile" means to be able to load a SavedModel file and then assign those weights to the RankingModel.model, which is a keras model. If you only want to run inference, you do not need to compile it. This was the "trick" to resave models. As a by-product it also allows us to retrain models.

python/ml4ir/data/tfrecord_helper.py
lastmansleeping: Yep. I had to move it to a new module to avoid cyclic imports.

lastmansleeping: Not new code.

python/ml4ir/features/feature_config.py
lastmansleeping: Created new work item -> https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000007QMb2IAG/view

It's a simple change. But we might want to use this ability to add more unit tests and update the model training tests to check for weights and loss convergence.

python/ml4ir/data/tfrecord_reader.py
```diff
@@ -79,72 +88,102 @@ def _parse_sequence_example_fn(sequence_example_proto):
...
+        # Define mask to identify padded records
+        if required_only and not feature_config.get_rank("serving_info")["required"]:
+            """
+            Define dummy mask if the rank field is not a required field for serving
+
+            NOTE:
+            This masks all max_num_records as 1 as there is no real way to know
+            the number of records in the query. There is no predefined required field,
+            and hence we would need to do a full pass of all features to find the record shape.
+            This approach might be unstable if different features have different shapes.
```
jakemannix, line 100: some features *do* have different shapes, right?  Text sequence features, for example.  Have we unit tested if it breaks with that?

python/ml4ir/data/tfrecord_reader.py
jakemannix: Yeah, that's fine given how we train now.  In theory, providing a list of *all negative examples* for a query is a valid thing to do, but we'd have to adapt our loss functions to support that.

(In particular, if we did strictly pointwise stuff, with no lists as we have now, there would be no requiremnet that every query had a positive example)

python/ml4ir/features/feature_config.py
```diff
@@ -252,24 +256,39 @@ def get_group_metrics_keys(self, key: str = None):
...
         Returns:
             Dictionary of tensorflow graph input nodes
         """
 
         def get_shape(feature_info: dict):
             feature_layer_info = feature_info["feature_layer_info"]
             preprocessing_info = feature_info.get("preprocessing_info", {})
+
+            # Setting size to None for sequence features as the num_records is variable
+            num_records = None
```
jakemannix, line 288: make symmetrical: 

```
if feature_info["tfrecord_type"] == TFRecordTypeKey.CONTEXT:
  num_records = None
else
  num_records = 1
```

python/ml4ir/features/feature_layer.py
```diff
@@ -43,11 +28,12 @@ def get_sequence_encoding(input, feature_info, max_num_records):
     if feature_info["tfrecord_type"] == TFRecordTypeKey.CONTEXT:
         # If feature is a context feature then tile it for all records
         encoding = tf.expand_dims(encoding, axis=1)
-        encoding = K.repeat_elements(encoding, rep=max_num_records, axis=1)
-
+        # encoding = tf.tile(encoding,
```
jakemannix, line 31: commented out?

python/ml4ir/data/tfrecord_reader.py
lastmansleeping: This is just the mask though. It has a defined shape. `[batch_size, num_records]`. In this particular case, which we will not hit anymore(since we have `rank` defined as a mandatory field at inference time), we just create a mask of 25 1s, because there is no way to know how many records are there.

python/ml4ir/features/feature_layer.py
```diff
@@ -100,7 +102,7 @@ def feature_layer(inputs):
         Reshape ranking features to create features of shape
         [batch, max_num_records, num_features]
         """
-        ranking_features = tf.concat(ranking_features, axis=-1)
+        ranking_features = tf.concat(ranking_features, axis=-1, name="ranking_features")
```
jakemannix, line 105: what is meant by "ranking features" here?  What other features are there?  Also, are we changing types again?  `ranking_features` starts as a list, now it's a tensor?  I'm surprised mypy lets you do that!

python/ml4ir/features/feature_layer.py
lastmansleeping: Yeah, I moved it out of this method. Look below.

```
tile_dims = tf.shape(
                            tf.expand_dims(
                                tf.expand_dims(tf.gather(inputs["mask"], indices=0), axis=0),
                                axis=-1,
                            )
                        )
                        encoding = tf.tile(encoding, tile_dims)
```

I will remove this commented out code.

python/ml4ir/model/pipeline.py
```diff
@@ -244,15 +244,34 @@ def run(self):
...
                     test_dataset=ranking_dataset.test,
                     inference_signature=self.args.inference_signature,
                     logs_dir=self.logs_dir,
                     logging_frequency=self.args.logging_frequency,
                 )
 
+            # Save model
+            # NOTE: Model will be saved with the latest serving signatures
+            if self.args.execution_mode in {
+                ExecutionModeKey.TRAIN_INFERENCE_EVALUATE,
```
jakemannix, line 261: it seems like these should be broken down (in a separate story), into whether you want to save the model, train a model, etc.  Instead of one enum with a combinatorial explosion of different values.

python/ml4ir/model/pipeline.py
```diff
@@ -244,15 +244,34 @@ def run(self):
...
+                ExecutionModeKey.TRAIN_INFERENCE,
+                ExecutionModeKey.TRAIN_ONLY,
+                ExecutionModeKey.INFERENCE_EVALUATE_RESAVE,
+                ExecutionModeKey.EVALUATE_RESAVE,
+                ExecutionModeKey.INFERENCE_RESAVE,
+                ExecutionModeKey.RESAVE_ONLY,
+            }:
+                # Save model
+                ranking_model.save(
+                    models_dir=self.models_dir, pad_records=self.args.pad_records_at_inference
```
jakemannix, line 272: so why do we have the `pad_records_at_inference` flag?  When would we want to change that flag's value?

python/ml4ir/model/ranking_model.py
```diff
@@ -531,96 +550,17 @@ def load_model(self, model_file: str) -> Model:
         model = tf.keras.models.load_model(model_file, compile=False)
 
         self.logger.info("Successfully loaded SavedModel from {}".format(model_file))
-        self.logger.warning("Retraining is not supported. Model is loaded with compile=False")
+        self.logger.warning("Retraining is not yet supported. Model is loaded with compile=False")
```
jakemannix, line 553: :)

python/ml4ir/features/feature_layer.py
lastmansleeping: There are three two types of features - `ranking_features` and `metadata_features`. The former is set of features that we learn weights on and are required at inference time to get the ranking score. `metadata_features` are needed at training time like `rank`, `organization_id`, `mask`, etc that are not linked to a weight but help us compute custom ranking losses and metrics.

python/ml4ir/model/pipeline.py
lastmansleeping: I agree.

python/ml4ir/model/serving.py
```diff
@@ -0,0 +1,119 @@
...
+    Limitation: This limits the serving signature to only run inference on a single query
+    at a time given the current implementation. This is a tricky issue to fix because
+    there is no real way to generate a dense tensor of ranking scores from different queries,
+    as they might have varying number of records in each of them.
+
+    Workaround: To infer on multiple queries, run predict() on each of the queries separately.
+    """
+    tfrecord_parse_fn = make_parse_fn(
+        feature_config=feature_config,
+        max_num_records=25,
```
jakemannix, line 55: we still have this "max" of 25?

python/ml4ir/model/pipeline.py
lastmansleeping: I'm just leaving an option open. Not padding limits us to only predict scores on one query at a time. If for some use case, we care more about batch inference and less about padded records, then we can use this.

`pad_records_at_inference` is set to False by default. So should be fine.

python/ml4ir/model/serving.py
lastmansleeping: Yep. I was hoping you wouldn't catch this.

So, this path would only be invoked if you don't pass in a rank field in the data - then there is no way to guess the num_records in the query and it reverts back to this set `max_num_records`. Don't worry, it will not be invoked in our case since we have set `rank` as `required=True`. It will break earlier if you do not pass `rank` in the proto.

python/ml4ir/data/tfrecord_reader.py
lastmansleeping: Agreed.

python/ml4ir/tests/test_serving.py
```diff
@@ -130,3 +135,93 @@ def _filter_records(x, mask):
...
+        tfrecord_model = kmodels.load_model(
+            os.path.join(self.output_dir, "final", "tfrecord"), compile=False
+        )
+        assert ServingSignatureKey.TFRECORD in tfrecord_model.signatures
+        tfrecord_signature = tfrecord_model.signatures[ServingSignatureKey.TFRECORD]
+
+        for num_records in range(1, 250):
+            proto = tf.constant(
+                [
+                    feature_config.create_dummy_sequence_example(
```
jakemannix, line 176: we definitely need to be able to start testing with sequence examples with a variety of values in them (instead of all zeroes).  Not for this PR, but soon.  This is barely testing the model.

python/ml4ir/tests/test_serving.py
```diff
@@ -130,3 +135,93 @@ def _filter_records(x, mask):
...
+                        num_records=num_records
+                    ).SerializeToString()
+                ]
+            )
+            try:
+                tfrecord_signature(sequence_example_protos=proto)
+            except Exception:
+                assert False
+
+    def test_serving_required_fields_only(self):
```
jakemannix, line 186: these two tests share almost all of their logic.  Maybe just have that logic live in one helper method, which you call with first `required_only` being True, then False?

python/ml4ir/data/tfrecord_reader.py
lastmansleeping: Going back to this
- There is no need for an explicit assert. If the required feature is not present, then `tf.io.parse_single_sequence_example()` will blow up.
- Might be difficult/impossible to do an assert to check presence of a dictionary key using `tf.debugging`

python/ml4ir/tests/test_serving.py
lastmansleeping: This test is not designed to test the model. This is only testing if the serving signature works for different number of records in a query from 1 to 250 records. The model training tests are under `test_ranking_model`. But I concede that those tests are not great.

python/ml4ir/tests/test_serving.py
lastmansleeping: Yep. I got lazy at the end. Will update.

python/ml4ir/features/feature_config.py
jakemannix: for future reference: links to internal GUS work items won't be helpful in the open source world ;P

PR #17, opened 2020-04-03
Scala yaml parsing

jakemannix:
scala makes simple data structs very concise, and it plays nicely with jackson, including the nice:

```
@JsonIgnoreProperties(ignoreUnknown = true)
```
annotation!


jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/SavedModelBundleExecutor.scala
```diff
@@ -19,23 +19,22 @@ import org.tensorflow.example.SequenceExample
...
   */
-class SavedModelBundleExecutor(dirPath: String, config: ModelExecutorConfig)
-    extends (SequenceExample => Array[Float]) {
+class SavedModelBundleExecutor(dirPath: String, config: ModelExecutorConfig) extends (SequenceExample => Array[Float]) {
   val savedModelBundle = SavedModelBundle.load(dirPath, "serve")
   val session = savedModelBundle.session()
 
   override def apply(proto: SequenceExample): Array[Float] = {
-    val ModelExecutorConfig(inputNode, outputNode, padTo, _) = config
+    val ModelExecutorConfig(inputNode, outputNode, _, _) = config
```
darshshah, line 27: Let's remove the 3rd and 4th args from ModelExecutorConfig? Currently, I manually set it to 50 for num docs per query and 20 as max query len. But since we are not using this anymore, we should just remove to simplify the client

jakemannix: Good point! Will do that.

jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/data/FeaturePreprocessor.scala
```diff
@@ -65,3 +65,32 @@ case class PrimitiveProcessor() {
...
 }
+
+object PrimitiveProcessors {
+
+  def fromFunctionMaps(featuresConfig: FeaturesConfig,
+                       tfRecordType: String,
+                       fFuncs: Map[String, Float => Float],
+                       lFuncs: Map[String, Long => Long],
+                       sFuncs: Map[String, String => String]): Map[DataType, Map[String, PrimitiveProcessor]] = {
+    featuresConfig
```
darshshah, line 76: Do we want to support boolean values here for extensibility? Now that we are making this generic, we should support boolean features (they can still be passed as floats 1/0). Maybe in v2.

darshshah: DataType.BOOL

jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/data/FeatureProcessors.scala
```diff
@@ -16,42 +17,42 @@ object FeatureProcessors {
...
-              (servingName, new PrimitiveProcessor() {
-                override def processString(s: String): String = strfns(servingName)(s)
-              })
-          }
-      }
-    StringMapFeatureProcessor(featuresConfig, perFieldPrimitiveProcessors)
+                    floatFns: JMap[String, JFunction[JFloat, JFloat]],
+                    longFns: JMap[String, JFunction[JLong, JLong]],
+                    strFns: JMap[String, JFunction[JString, JString]]): StringMapFeatureProcessor = {
+    val (ffs, lfs, sfs) = toScalaFns(floatFns, longFns, strFns)
```
darshshah, line 38: minor: ffs/lfs/sfs can be more descriptive.

jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/data/FeaturePreprocessor.scala
jakemannix: SequenceExample in TF protobuf-land does not support those - only 32bit float, String == utf8-bytes, and 64bit long ints

jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/data/FeaturesConfigHelper.scala
```diff
@@ -1,23 +1,20 @@
...
-import scala.collection.JavaConverters._
-import org.tensorflow.DataType
-
 case class NodeWithDefault(nodeName: String, defaultValue: String)
 
 object FeaturesConfigHelper {
-  implicit class ModelFeaturesConverter(modelFeatures: ModelFeatures) {
+  implicit class MFConverter(mf: ModelFeaturesConfig) {
     def toFeaturesConfig(tfRecordType: String): FeaturesConfig = {
+      val features = mf.features ++ (if (tfRecordType.equals("sequence")) List(mf.initialRank) else List.empty)
```
darshshah, line 8: Special handling for rank :/

jakemannix: yes, it needs documentation explaining why

jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/data/ModelFeaturesConfig.scala
```diff
@@ -0,0 +1,38 @@
...
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+case class ModelFeaturesConfig(@JsonProperty("rank") initialRank: FeatureConfig,
+                               @JsonProperty("features") features: List[FeatureConfig])
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+case class FeatureConfig(@JsonProperty("node_name") nodeName: String,
+                         @JsonProperty("dtype") dTypeString: String,
+                         @JsonProperty("serving_info") servingConfig: ServingConfig,
+                         @JsonProperty("default_value") defaultValue: String,
```
darshshah, line 32: With this, we are NOT moving default_value under serving_info. Is this something we stick with for now? We can update it in v2 but this will have a cascading effect later on (i.e. changes to yaml and backward compatibility).

jakemannix: ah, we can do either way - what did we say we wanted to do?  I can change it to whatever way we want the config to be.  I didn't want to break any current configs.

We don't need to worry about backwards compatibility until this ships, and even then, until a major version of ml4ir ships, we can change, as long as we know models go out with new jars as needed.

jvm/ml4ir-inference/pom.xml
```diff
@@ -10,6 +10,10 @@
   <artifactId>ml4ir-inference</artifactId>
   <packaging>jar</packaging>
 
+  <properties>
+    <jackson.version>2.7.5</jackson.version>
```
darshshah, line 14: Update this to 2.10.0

darshshah: We can re-use the jackson version from parent's pom instead of mentioning it here again. But we have to use 2.10.0 (as this version is not compatible with spring)

jvm/pom.xml
```diff
@@ -13,6 +13,7 @@
     <scala.version>2.11.8</scala.version>
     <scala.artifact.suffix>2.11</scala.artifact.suffix>
     <tensorflow.version>1.15.0</tensorflow.version>
+    <jackson.version>2.7.5</jackson.version>
```
darshshah, line 16: 2.10.0 here as well

jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/data/ModelFeaturesConfig.scala
darshshah: Also update the model_features.yaml to move the default_value under serving_info.

jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/data/SequenceExampleBuilder.scala
```diff
@@ -0,0 +1,227 @@
...
+
+  /**
+    * Effectively transforms an array of maps of features into a map of arrays of features: the "transpose" operation
+    * @param docFeatures to have their features extracted out into one dense array per feature
+    * @return map of feature-name -> padded dense vector of numeric features
+    */
+  private def transpose[T: ClassTag](
+      docFeatures: Array[Map[String, T]]
+  ): Map[String, Array[T]] = {
+    // TODO: check that we don't need to pad anymore
```
darshshah, line 82: Old comment? I don't think we need this anymore.

jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/data/TFRecordIO.scala
```diff
@@ -0,0 +1,70 @@
+package ml4ir.inference.tensorflow.data
+import java.io.{ByteArrayOutputStream, File, FileOutputStream}
+
+import com.google.common.hash.Hashing
+import com.google.common.io.LittleEndianDataOutputStream
+import org.tensorflow.example.SequenceExample
+
+object TFRecordIO {
+  val kMaskDelta = 0xa282ead8
```
darshshah, line 9: A comment on what/why this number is used will be helpful.

jakemannix: magic!

jakemannix: TBH, I'll just delete this.  There is a much better way of TFRecord IO in scala.  It's called "spark" ;P

darshshah:
Some of the code in GenericSequenceExampleBuildingTest.scala is not used in any tests. E.g.  serializeTestData/sampleSequenceExamples, etc. Ideally, we should remove such code but I think we are planning to create a test case using this in the next iteration. So we can keep it for now.

jakemannix:
> Some of the code in GenericSequenceExampleBuildingTest.scala is not used in any tests. E.g. serializeTestData/sampleSequenceExamples, etc. Ideally, we should remove such code but I think we are planning to create a test case using this in the next iteration. So we can keep it for now.

I can remove it and put it on another branch.  Cleaner that way

darshshah:
> > Some of the code in GenericSequenceExampleBuildingTest.scala is not used in any tests. E.g. serializeTestData/sampleSequenceExamples, etc. Ideally, we should remove such code but I think we are planning to create a test case using this in the next iteration. So we can keep it for now.
> 
> I can remove it and put it on another branch. Cleaner that way

Sounds good. Let's do that so that we have a clean 0.0.2

PR #24, opened 2020-05-07
Refactor to onboard new ML applications

lastmansleeping:
Still work in progress. Check the **[jupyter notebook](https://github.com/salesforce/ml4ir/blob/ashish/generic_library/python/notebooks/PointwiseRankingDemo.ipynb)** attached to see progress and level of abstraction.


python/ml4ir/base/model/relevance_model.py
```diff
@@ -51,9 +48,9 @@ def __init__(
         self.scorer = scorer
         self.tfrecord_type = tfrecord_type
 
-        try:
+        if scorer:
```
jakemannix, line 51: :)

python/ml4ir/applications/ranking/tests/test_base.py
```diff
@@ -43,7 +55,7 @@ def setUp(
         random.seed(123)
```
jakemannix, line 52: seed should probably be configurable via args, right?

python/ml4ir/applications/ranking/tests/test_base.py
```diff
@@ -7,18 +7,30 @@
...
+from typing import Union, List, Type
 
 warnings.filterwarnings("ignore")
 
 
-OUTPUT_DIR = "ml4ir/tests/test_output"
-ROOT_DATA_DIR = "ml4ir/tests/data"
+OUTPUT_DIR = "ml4ir/applications/ranking/tests/test_output"
+ROOT_DATA_DIR = "ml4ir/applications/ranking/tests/data"
 FEATURE_CONFIG_FNAME = "feature_config.yaml"
```
jakemannix, line 31: should this be a static filename, or configurable as well?

python/ml4ir/applications/ranking/tests/test_features.py
```diff
@@ -55,12 +47,16 @@ def test_sequence_encoding(self):
         encoding_size = 512
         feature_info = {
             "feature_layer_info": {
-                "embedding_size": embedding_size,
-                "encoding_type": "bilstm",
-                "encoding_size": encoding_size,
+                "type": "numeric",
+                "fn": "get_sequence_encoding",
+                "args": {
+                    "embedding_size": embedding_size,
```
jakemannix, line 50: can the test try and verify that embedding_size != encoding_size is ok?

jakemannix: oh wait, it already is checking this, right?

python/ml4ir/applications/ranking/tests/test_metrics.py
```diff
@@ -4,64 +4,59 @@
...
 
 # Constants
 GOLD_METRICS = {
-    "old_ACR": 1.6669034,
-    "new_ACR": 2.3998579,
-    "old_MRR": 0.7806604,
-    "new_MRR": 0.6045914,
-    "perc_improv_ACR": -43.9710,
-    "perc_improv_MRR": -22.5538,
+    "old_ACR": 1.6669,
```
jakemannix, line 12: as discusssed: we still need a better way to do this...  :\

python/ml4ir/applications/ranking/tests/test_base.py
lastmansleeping: This is in only for the tests. It is configurable through command line arguments when you are running the actual pipeline itself.

python/ml4ir/applications/ranking/tests/test_base.py
lastmansleeping: I don't understand why you would want to make the test data configurable? What are we testing in that case?

python/ml4ir/applications/ranking/tests/test_features.py
lastmansleeping: I think `flake8` has auto-formatted the code in some "weird" way. But yeah, in the test itself the encoding_size != embedding_size

python/ml4ir/applications/ranking/tests/test_metrics.py
lastmansleeping: Let me file a github issue to track this.

jakemannix: I would reach out to Mario, actually, on what his recommendations are for unit tests training ML models.  I tend to go for sanity checks: take a "toy" training dataset where clicks show up relatively randomly (with a simple positional bias) yet where where the "popularity" feature is clearly correlated with clicks.  Compute by hand what the MRR of the positional bias random click would be, and then assert that training for 5 epochs using the popularity feature (and the others) gets strictly better MRR, and that in fact 10 epochs is even better than 5 epochs.

We don't need this for _this_ PR, but we really need more like that kind of testing.

python/ml4ir/applications/ranking/tests/test_base.py
jakemannix: nevermind, it's test, got it.

PR #27, opened 2020-05-21
Create structure for synthetic data (including CF)

calliebradley:


python/applications/ranking/scripts/run_create_dataset.py
```diff
@@ -0,0 +1,19 @@
+# Run from ml4ir/python directory as: python3 applications/ranking/scripts/run_create_dataset.py
+
+from applications.ranking.data.scripts.create_dataset import create_dataset
+
+
+def main():
```
lastmansleeping, line 6: @calliebradley Why not add the `main()` to `create_dataset.py` itself?

lastmansleeping: Oh. This is under `ranking/scripts/..`. I actually think it makes sense to be one single self contained python module under `ranking/data/scripts/create_dataset.py`.

python/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # Build data
+        example_data = load_example_data(data_dir, logger)
+
+        df_synthetic = fill_data(logger, example_data, max_num_records, feature_config, num_samples)
+        df_synthetic.to_csv(out_file, index=False)
+        logger.info('Synthetic data created! Location: {}'.format(out_file))
+        return df_synthetic
+
+    except Exception as e:
```
lastmansleeping, line 58: @calliebradley What error are you trying to catch here? Can you add the specific type?

python/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,148 @@
...
+        set_seeds(random_state)
+        logger.info('Set seeds with initial random state {}'.format(random_state))
+
+        # Load and parse feature config #TODO check on tfrecord_type
+        feature_config: FeatureConfig = parse_config(tfrecord_type='', feature_config=feature_config_src, logger=logger)
+        logger.info("Feature config parsed and loaded")
+
+        # Create output location
+        if not os.path.exists(data_out_dir):
+            os.makedirs(data_out_dir)
```
lastmansleeping, line 46: @calliebradley You should be using `ml4ir.io.file_io` for all this. If there's no method already to create, I would encourage you to add it. I probably should do a pass through to look at where I've used python methods directly.

python/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,148 @@
...
+        out_file = os.path.join(data_out_dir,
+                                'synthetic_data_{}.csv'.format(datetime.datetime.now().strftime('%Y%m%d-%H%M%S')))
+
+        # Build data
+        example_data = load_example_data(data_dir, logger)
+
+        df_synthetic = fill_data(logger, example_data, max_num_records, feature_config, num_samples)
+        df_synthetic.to_csv(out_file, index=False)
```
lastmansleeping, line 54: @calliebradley Again, should be using `ml4ir.io.file_io` here as well.

python/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,148 @@
...
+    1. Generate a query key
+    2. Random sample number of results (with replacement)
+    3. Random sample to fill each result (using FeatureConfig)
+    4. TODO Check that no record ID is used twice
+    """
+
+    # Set up new DF and example data distributions for sampling options
+    rows_df = []
+    seed_dict = {}  # the keys are source data columns, the values are lists of source data column values
+    seed_dict_labeled_pos = {}  # option to get characteristics of labeled results separately
```
lastmansleeping, line 97: @calliebradley We had discussed to make this k-class compatible, right?

python/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,148 @@
...
+    seed_dict_labeled_neg = {}
+
+    name_num_sequences = [x['serving_info']['name'] for x in feature_config.all_features if x['name']=='num_sequences'][0]
```
lastmansleeping, line 100: @calliebradley So three things.
1. try to use `feature_config.get_all_features()`
2. why are you using `serving_info`. That's more for serving side information for the inference JVM code. You should avoid dependency on it.
3. I think the best way to achieve what you are trying to do is to accept the name of the field from the user itself(as an argument) instead of hard coding it.

python/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,148 @@
...
+    name_label_rank = [x['serving_info']['name'] for x in feature_config.all_features if x['name']=='label_rank'][0]
```
lastmansleeping, line 101: @calliebradley I don't understand what is the goal of this statement.

python/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,148 @@
...
+            for mycol in list(example_data.columns):
+                # Could also let user specify distribution to sample from for each feature
+                if mycol not in row_dict.keys() and is_labeled_pos:
+                    # row_dict[mycol] = random.sample(seed_dict_labeled_pos[mycol])
+                    row_dict[mycol] = None  # Default value, check type # Could also be functions we pass in
+                elif mycol not in row_dict.keys() and not is_labeled_pos:
+                    # row_dict[mycol] = random.sample(seed_dict_labeled_neg[mycol])
+                    row_dict[mycol] = None
+
+            # Create catastrophic failure # Rename, possible to extenuate this for each high value feature
```
lastmansleeping, line 138: We should probably come up with a more descriptive name than "catastrophic failure" mostly because it might not be catastrophic for everyone and also it contains no information about what the failure actually is.

python/applications/ranking/data/scripts/create_dataset.py
lastmansleeping: Why can't you just use `rank`?

python/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,148 @@
...
+    seed_dict_labeled_neg = {}
+
+    name_num_sequences = [x['serving_info']['name'] for x in feature_config.all_features if x['name']=='num_sequences'][0]
+    name_label_rank = [x['serving_info']['name'] for x in feature_config.all_features if x['name']=='label_rank'][0]
+    name_highval_feature = [x['serving_info']['name'] for x in feature_config.all_features if x['name']=='high_val_feature_1'][0]
+
+    for mycol in list(example_data.columns):
+        # Generate data distributions from sample data
+        seed_dict[mycol] = list(example_data[mycol].values)
+        seed_dict_labeled_pos[mycol] = list(example_data[example_data[feature_config.get_label('name')] == 1].values)
```
lastmansleeping, line 107: @calliebradley We shouldn't write ourselves into the 0/1 corner IMO. Should be easy to make this k-class.

python/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,148 @@
...
+        seed_dict_labeled_neg[mycol] = list(example_data[example_data[feature_config.get_label('name')] == 0].values)
+
+    seed_dict[name_num_sequences] = list(filter(lambda x: x <= max_num_records, seed_dict[name_num_sequences]))
+
+    for _ in range(num_samples):
+        # Generate a synthetic query ID
+        seed_id = str(random.sample(seed_dict[feature_config.get_query_key('name')], 1)[0])
```
lastmansleeping, line 114: @calliebradley While I think this is okay for now, I think it might be more robust if you can just generate a random string or number based on the `dtype` of `query_key`.

python/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,148 @@
...
+        q_id = ''.join(random.sample(seed_id, len(seed_id)))
+
+        # Sample the number of results and label rank from example data
+        q_nseq = random.sample(seed_dict[name_num_sequences], 1)[0]
```
lastmansleeping, line 118: @calliebradley It's okay to be descriptive in terms of variable names to enhance readability. Especially since this is public.

python/applications/ranking/data/scripts/create_dataset.py
lastmansleeping: On second thoughts, let's leave it at 0/1 click/no-click for now. We can change later if needed.

python/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,148 @@
...
+                    # row_dict[mycol] = random.sample(seed_dict_labeled_pos[mycol])
+                    row_dict[mycol] = None  # Default value, check type # Could also be functions we pass in
+                elif mycol not in row_dict.keys() and not is_labeled_pos:
+                    # row_dict[mycol] = random.sample(seed_dict_labeled_neg[mycol])
+                    row_dict[mycol] = None
+
+            # Create catastrophic failure # Rename, possible to extenuate this for each high value feature
+            # {high_value_feature_1:[label1_val, label2_val], high_value_feature_2: [unclicked_val, clicked_val]}
+            # Not creating examples of non-catastrophic failure, ie where the clicked record is not a name match
+            # Configurable feature to pass in here! Pass in: feature name
```
lastmansleeping, line 141: @calliebradley What is `highval_feature`? I think we need a better name.

python/applications/ranking/data/scripts/create_dataset.py
lastmansleeping: Also, might be better to move this to a separate method to keep it cleaner for the future.

python/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,148 @@
...
+                feature_config.get_query_key('name'): q_id,
+                name_num_sequences: q_nseq,
+                name_label_rank: q_labelrank,
+                feature_config.get_rank('name'): i + 1
+            }
+            for mycol in list(example_data.columns):
+                # Could also let user specify distribution to sample from for each feature
+                if mycol not in row_dict.keys() and is_labeled_pos:
+                    # row_dict[mycol] = random.sample(seed_dict_labeled_pos[mycol])
+                    row_dict[mycol] = None  # Default value, check type # Could also be functions we pass in
```
lastmansleeping, line 133: @calliebradley I think you should use the default value here as opposed to None

python/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,148 @@
...
+        # Generate a synthetic query ID
+        seed_id = str(random.sample(seed_dict[feature_config.get_query_key('name')], 1)[0])
+        q_id = ''.join(random.sample(seed_id, len(seed_id)))
+
+        # Sample the number of results and label rank from example data
+        q_nseq = random.sample(seed_dict[name_num_sequences], 1)[0]
+        q_labelrank = random.sample(seed_dict[name_label_rank], 1)[0]
+
+        for i in range(int(q_nseq)):
+            is_labeled_pos = (i + 1 == q_labelrank)
```
lastmansleeping, line 122: @calliebradley Should be okay to say "click"/"no-click" since this is under `ranking/` anyway.

python/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,148 @@
...
+            row_dict = {
+                feature_config.get_query_key('name'): q_id,
+                name_num_sequences: q_nseq,
+                name_label_rank: q_labelrank,
```
lastmansleeping, line 126: @calliebradley I don't understand what `name_label_rank` is.

lastmansleeping:
@calliebradley Did a first pass through and added review comments. 

Two things that I noticed not in the above comments:
- We should really be defining the synthetic data info in the feature config yaml itself as `synthetic_data_info` or something.
- We should also enable the method to write TFRecords in addition to CSVs.

python/applications/ranking/scripts/run_create_dataset.py
calliebradley: @lastmansleeping Yeah, I don't mind moving it. I thought we talked about this in our first syncup, that I built the run example separately when I built the create_dataset module as a class callable from pipeline.py (similar to NER's search assistant modeling process). I'll move it for now and we can always change it in the future.

python/applications/ranking/data/scripts/create_dataset.py
calliebradley: @lastmansleeping As we discussed, no specific error, just want a graceful exit with a good callback if anything goes wrong.

python/applications/ranking/data/scripts/create_dataset.py
calliebradley: Thanks @lastmansleeping, good point and changed to use `file_io.make_directory`. Question about that method though - why does it take `dir_path` as an argument and also return `dir_path` ? Seems like the return object is unnecessary?

python/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,148 @@
...
+        logger.info("Feature config parsed and loaded")
+
+        # Create output location
+        if not os.path.exists(data_out_dir):
+            os.makedirs(data_out_dir)
+        out_file = os.path.join(data_out_dir,
+                                'synthetic_data_{}.csv'.format(datetime.datetime.now().strftime('%Y%m%d-%H%M%S')))
+
+        # Build data
+        example_data = load_example_data(data_dir, logger)
```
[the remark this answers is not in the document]
calliebradley: Yeah, not sure yet what will make the most sense but changed to `seed_data` for now.

python/applications/ranking/data/scripts/create_dataset.py
calliebradley: Done

python/applications/ranking/data/scripts/create_dataset.py
calliebradley: Yes, I don't want to build that yet but in the future, yes.

python/applications/ranking/data/scripts/create_dataset.py
calliebradley: Thanks @lastmansleeping , I think I've made all the changes we discussed now.

python/applications/ranking/data/scripts/create_dataset.py
calliebradley: updated

python/applications/ranking/data/scripts/create_dataset.py
calliebradley: moved

python/applications/ranking/data/scripts/create_dataset.py
calliebradley: @lastmansleeping This was Jake's suggestion for cat failure. If you don't like cat failure or highval feature, feel free to suggest something else ;) Otherwise I think this is enough for now. The 'name_match' makes it more explicit.

python/applications/ranking/data/scripts/create_dataset.py
calliebradley: @lastmansleeping I want to leave this for the L3 synthetic data story because I think it can easily scope creep. It wasn't meant to be part of this story.

python/applications/ranking/data/scripts/create_dataset.py
calliebradley: I've commented/described things more explicitly now. We can continue to work with these names in the future.

calliebradley:
Thanks @lastmansleeping , I think I've made all the updates/changes we discussed now. The following are great points for us to keep in mind for the L3 synth data story (but not including here because I see potential scope creep in them): 
- We should really be defining the synthetic data info in the feature config yaml itself as synthetic_data_info or something.
- We should also enable the method to write TFRecords in addition to CSVs.

calliebradley:
@lastmansleeping Woohoo, finished resolving all conflicts from the master changes. Review away :)

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,227 @@
...
+                    # row_dict[mycol] = random.sample(seed_dict_labeled_neg[mycol])
+                    row_dict[mycol] = None
+
+            row_dict = add_feature_highval(row_dict, feature_highval, is_labeled_pos, q_labelrank)
+
+            rows_df.append(row_dict)
+    return pd.DataFrame(rows_df)
+
+
+def main():
```
lastmansleeping, line 204: @calliebradley Can you add a comment note here on how to run this? And maybe an example as well?

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,227 @@
...
+
+import os
+import random
+import traceback
+import datetime as dt
+import pandas as pd
+import numpy as np
+from logging import Logger
+
+from ml4ir.io import file_io
```
lastmansleeping, line 11: @calliebradley You need to update this to point to `ml4ir.base.io` and similarly for others as well.

calliebradley:
@lastmansleeping Thanks, updated to new directory structure! Should run for you now.

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
calliebradley: @lastmansleeping Thanks, there's a comment at the top actually, noting how to run, both here and in test_create_dataset.py. For running from here with example data, you'll need to download the example data and feature_config.yaml that I noted in the Gus story, they are posted here: /projects/search_relevance/ml4ir_support_files/synth_data

You don't need to download anything extra to run the test_create_dataset.py though :)

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,227 @@
...
+            is_labeled_pos = (i + 1 == q_labelrank)
+            row_dict = {
+                feature_query_key: q_id,
+                nres: q_nseq,
+                'label_rank': q_labelrank,
+                feature_rank: i + 1
+            }
+            for mycol in list(seed_data.columns):
+                # Could also let user specify distribution to sample from for each feature
+                if mycol not in row_dict.keys() and is_labeled_pos:
```
lastmansleeping, line 191: @calliebradley What happens to features that were in the config and not in the example data? I don't think we should fill all columns in the example data, rather we should fill all columns in the feature config.features. Not sure if you are already doing that.

python/ml4ir/applications/ranking/tests/data/csv/synthetic/feature_config.yaml
```diff
@@ -0,0 +1,114 @@
+query_key:
```
lastmansleeping, line 1: @calliebradley Is this the same as the feature_config under `applications/ranking/tests/data/tfrecord/feature_config.yaml`?

calliebradley: @lastmansleeping Almost - I need/want to use query_str instead of query_key in order to test the building of synthetic query IDs. The random sampling would be a no-op using query_key since it's only one char long.

calliebradley: If you're ok changing tfrecord/feature_config.yaml to use query_str, we could consolidate.

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
calliebradley: @lastmansleeping Interesting, right I am replicating the features that are present in the seed data. Would you be ok making this call as a part of the L3 synthetic data story? In my mind it is peripheral to the creation of the cat failure feature. I can see that you will probably disagree with me though.

python/ml4ir/applications/ranking/tests/data/csv/synthetic/feature_config.yaml
lastmansleeping: It does have `query_str`, right?

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,227 @@
...
+
+            rows_df.append(row_dict)
+    return pd.DataFrame(rows_df)
+
+
+def main():
+    data_dir = 'ml4ir/applications/ranking/data/example/'
+    out_dir = 'ml4ir/applications/ranking/data/synthetic/'
+    feature_config = 'ml4ir/applications/ranking/data/example/feature_config.yaml'
+    feature_highval = {'name_match':[0,1]}
```
lastmansleeping, line 208: @calliebradley [Minor] You should be able to say "text_match_bool" I think since it is a generic ranking term.

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,227 @@
...
+                    row_dict[mycol] = None
+
+            row_dict = add_feature_highval(row_dict, feature_highval, is_labeled_pos, q_labelrank)
+
+            rows_df.append(row_dict)
+    return pd.DataFrame(rows_df)
+
+
+def main():
+    data_dir = 'ml4ir/applications/ranking/data/example/'
```
lastmansleeping, line 205: @calliebradley We should be reading these are command line args, right? Why are we hardcoding these here?

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,227 @@
...
+    q_id = ''.join(random.sample(seed_id, len(seed_id)))
+    if q_id in set(seed_dict.keys()):
+        if log:
+            log.info('Generated query key is duplicate, regenerating. If happens repeatedly, query key may be too short')
+        return generate_key(seed_dict, query_key)
+    return q_id
+
+
+def filter_nres(seed_data, feature_query_key, feature_rank, feature_nres, max_num_records, logger):
+    nres = 'num_results_calc'
```
lastmansleeping, line 102: @calliebradley [Minor] Probably better to move all hardcoded string constants to a separate enum/class at the top. Something like here -> https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/base/model/relevance_model.py#L24

python/ml4ir/applications/ranking/tests/test_create_dataset.py
```diff
@@ -0,0 +1,74 @@
...
+        self.feature_config = feature_config
+        self.output_dir = output_dir
+        self.log_dir = log_dir
+
+        # Set up logging
+        file_io.make_directory(self.log_dir, clear_dir=True)
+        outfile: str = os.path.join(self.log_dir, "output_log.csv")
+        self.logger = setup_logging(reset=True, file_name=outfile, log_to_file=True)
+
+    def test_synthetic_data(self):
```
lastmansleeping, line 39: @calliebradley I like the test. Concise and puts us in a good place to build on top of.

python/ml4ir/applications/ranking/tests/data/csv/synthetic/feature_config.yaml
```diff
@@ -0,0 +1,114 @@
...
+    log_at_inference: false
+    feature_layer_info:
+      type: numeric
+      shape: null
+    serving_info:
+      name: feat_2
+      required: true
+      default_value: 0.0
+    tfrecord_type: sequence
+  - name: query_str
```
lastmansleeping, line 83: @calliebradley [Minor] I would just remove `query_str` from the features section of the YAML just so we are compatible with ml4ir. `query_key` is automatically added to `all_features` and other dictionaries in the feature_config.

lastmansleeping:
@calliebradley Ran the script with the test data and I get this output
```
"query_str","num_results_calc","label_rank","pos","query_key","feat_0","feat_1","feat_2","label","group","name_match"
"4K0OQXXD3",6,1,1,"","","","","","",0
"4K0OQXXD3",6,1,2,"","","","","","",0
"4K0OQXXD3",6,1,3,"","","","","","",0
"4K0OQXXD3",6,1,4,"","","","","","",0
"4K0OQXXD3",6,1,5,"","","","","","",0
"4K0OQXXD3",6,1,6,"","","","","","",0
"1CXEUJZKKI7GE",6,1,1,"","","","","","",0
"1CXEUJZKKI7GE",6,1,2,"","","","","","",0
"1CXEUJZKKI7GE",6,1,3,"","","","","","",0
"1CXEUJZKKI7GE",6,1,4,"","","","","","",0
"1CXEUJZKKI7GE",6,1,5,"","","","","","",0
"1CXEUJZKKI7GE",6,1,6,"","","","","","",0
"UIO6CVJXKBWB11",5,1,1,"","","","","","",0
"UIO6CVJXKBWB11",5,1,2,"","","","","","",0
"UIO6CVJXKBWB11",5,1,3,"","","","","","",0
"UIO6CVJXKBWB11",5,1,4,"","","","","","",0
"UIO6CVJXKBWB11",5,1,5,"","","","","","",0
```

What I see is that the label field is missing and also most of these examples don't have any name matches at all. Is that desired behavior?

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
lastmansleeping: Also, the defaults for `data_dir` and `logs_dir` should be the same as `ml4ir.base` as they are under `.gitignore`. We might have a lot of data files checked into github by mistake otherwise. Alternatively you could add `...../data/synthetic/` to the `.gitignore`. But looks like `logs_dir` is not working. My logs were written to current directory.

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -0,0 +1,227 @@
...
+        logger.error('All seed data has been filtered out!')
+    return seed_data, nres
+
+
+def add_feature_highval(row_dict, feature_highval, is_labeled_pos, q_labelrank):
+    # Create ranking-specific high value scenario
+    # of the format {high_value_feature_1:[label1_val, label2_val]}
+    # Not creating examples of non-catastrophic failure, ie where the clicked record is not a name match
+    for hvk in feature_highval.keys():
+        if is_labeled_pos and q_labelrank > 1:
```
lastmansleeping, line 124: @calliebradley [Minor] Can we better name `is_labeled_pos`. I don't follow what this is. You can use ranking specific terms like `click_rank` if that's what you are trying to do. Don't necessarily have to use `pos`. That was more of an abstraction for other tasks as well.

python/ml4ir/applications/ranking/tests/test_create_dataset.py
```diff
@@ -0,0 +1,74 @@
...
+                                  num_samples,
+                                  random_state = 123)
+        assert len(df) == 37
+        assert df.query_str.nunique() == num_samples
+        assert df.num_results_calc.max() <= max_num_records
+        assert 'name_match' in list(df.columns)
+        assert list(df.name_match.unique()) == [0,1]
+
+
+        df_2 = run_dataset_creation(self.root_data_dir,
```
lastmansleeping, line 59: @calliebradley [Minor] Looks like you need to do some PEP-8 formatting. ;-)

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
calliebradley: Marking this as resolved as we agreed this will be part of the L3 synthetic data creation story. Thanks :)

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
calliebradley: @lastmansleeping Hmm, I'm not sure about this. We specifically want a text match on the **name field**, not a text match on other textual fields. Especially if it's assumed to be a generic term, I could see that generating confusion. Thoughts?

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
calliebradley: The containment of the script to the one location here was @jakemannix's suggestion as far as I remember? I previously had a separate 'driver' script and I thought it was he who thought that was kind of structural overkill at this point. No way do I want to type them into the command line every time. I can go back to a driver script if we want that instead...

calliebradley: @lastmansleeping Also to clarify for the `data_dir`, do you want the default for **both** the seed data and the output synthetic data to be in `python/ml4ir/applications/ranking/data` (that seems dangerous)? Or which one do you want moved? I guess I'd prefer to add these directories to `.gitignore`

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
calliebradley: Ok, no problem. It used to be `click_rank` but I had requests to make it more generalized. Changing it back to `click_rank`.

python/ml4ir/applications/ranking/tests/test_create_dataset.py
calliebradley: Thanks, fixed :)

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
calliebradley: Jake requested it not be a class calling it overly structured ;) I can move these to the top / all-caps.

python/ml4ir/applications/ranking/tests/data/csv/synthetic/feature_config.yaml
calliebradley: Removed second `query_str` feature but also happy to change this to `query_key` if we make a multi-digit `query_key` so that it can test the resampling :)

calliebradley: Oops meant to tag you @lastmansleeping

lastmansleeping: No. This should be good for now. `query_str` should be unique as well, to an extent.

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
lastmansleeping: I get that. Not suggesting that we make the synthetic data creation a class. Just an enum for the constants. There is no enum data structure in python. So you can use a class. Or you can even use https://docs.python.org/3/library/enum.html

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
lastmansleeping: I remember even I suggested consolidation into a single script. But why does one script mean we can't define arguments? For now, even doing `sys.argv[i]` is fine really. But I'm not a fan of having to change code to run the script. I had to do this for running it on test data.

CC: @jakemannix

lastmansleeping: I want **input** `data_dir` to point to `ml4ir/python/data/` and `logs_dir` to point to `ml4ir/python/logs` to maintain consistency with rest of ml4ir. Of course to use this script in any real way you want to override `data_dir`. 

You can add `ml4ir/python/ml4ir/applications/ranking/data/synthetic` to `.gitignore` and write output data to that directory. That should be fine. Alternatively, you can also create a separate directory for input data(I don't think we should). But make sure you add all data and logs directory defaults to `.gitignore`.

lastmansleeping: "No way do I want to type them into the command line every time" 
I really do not get this. We can define defaults for each arg, right? You would only be specifying the `data_dir` and `feature_config` on each run and those should be mandatory args otherwise the script is not usable.

calliebradley:
Thanks @lastmansleeping for all your patience and feedback! :) Made these last few adjustments, including:
- Incorporated argparse (defaults are at top of file in all caps, I don't mind if you decide to move them in future)
- Fixed logs_dir and data_dir to be aligned with defaults. Added `/data/synthetic` to `.gitignore`
- Added missing `label` field
- Generalized `text_match_bool` arg
- General small stuff - pep-8 formatting, etc

So I think the last 'big' thing is prescribing features and default values from `feature_config` rather than from the seed data. Thanks for letting this be part of the future L3 synth data creation story :)

PR #32, opened 2020-06-04
Added cicleci initial config

darshshah:
Initial setup for circleci - the config is copied from CircleCI's getting started doc. Once this starts building, I'll update the workflow to add scala/python tests


jakemannix:
Lgtm! Let’s try it out!

PR #34, opened 2020-06-11
[WIP] Circleci project setup

darshshah:
- Update the circleci config file to add python tests.
- Added pytest-html to save the test reports from ci


darshshah:
Python tests are timing out after 10mins. I'll add a minimum threshold so that we don't hit this again.

jakemannix:
> Python tests are timing out after 10mins. I'll add a minimum threshold so that we don't hit this again.

hey @lastmansleeping, how long should the python tests take?

lastmansleeping:
> > Python tests are timing out after 10mins. I'll add a minimum threshold so that we don't hit this again.
> 
> hey @lastmansleeping, how long should the python tests take?

On my laptop `python3 -m pytest` takes about 3 minutes to complete.

.circleci/config.yml
```diff
@@ -0,0 +1,68 @@
+# Use the latest 2.1 version of CircleCI pipeline process engine. See: https://circleci.com/docs/2.0/configuration-reference
+# Reference doc: https://circleci.com/docs/2.0/project-walkthrough/
```
balikasg, line 2: Thanks for adding this. If you thought that nobody would read it, guess what: I am reading it.

.circleci/config.yml
```diff
@@ -0,0 +1,68 @@
...
+          key: python-deps-{{ checksum "python/requirements.txt" }}
+          paths:
+            - "venv"
+      - run:
+          name: Run Python tests
+          command: |
+            . venv/bin/activate
+            cd python
+            python3 -m pytest --html=pytest_report.html
+          no_output_timeout: 30m  # Time for running the tests before we timeout
```
balikasg, line 56: Are we on a paid version or on the free one? I am clicking on the "$" sign on the left navigation bar but "Something unexpected happens"

.circleci/config.yml
```diff
@@ -0,0 +1,68 @@
...
+      - store_artifacts:
+          path: python/pytest_report.html
+      - store_test_results:
+          path: python/pytest_report.html
+
+workflows:
```
balikasg, line 62: Nice! 
I did not know `circleci` but it is a nice and clean interface.

.circleci/config.yml
balikasg: Well actually by following this I was able to understand everything that is happening in this config.

darshshah: Glad it helped.

.circleci/config.yml
darshshah: We are a part of Salesforce's open-source organization. So it takes care of the credits/billing.

PR #38, opened 2020-06-15
Updating the setup file for packaging

darshshah:
- Added LICENSE.txt from parent dir
- Created MANIFEST.in to include license and readme
- Updated setup.py to add additional information

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -11,10 +11,10 @@
 ml4ir can be installed as a pip package by using the following command
 
 ```
-pip install  'git+https://git@github.com/salesforce/ml4ir#egg=ml4ir&subdirectory=python'
+pip install ml4ir
 ```
 
-This will install ml4ir-0.0.1 (the current version). In future, when this package is available on PyPI, it will be as simple as pip install ml4ir
+This will install ml4ir-0.0.1 (the current version) from PyPI.
 
 
 #### Docker (Recommended)
--- a/python/setup.py
+++ b/python/setup.py
@@ -5,14 +5,29 @@ def load_dependencies():
         required = f.read().splitlines()
     return required
 
+
+def getReadMe():
+    with open("README.md", "r") as f:
+        long_description = f.read()
+    return long_description
+
 setup(
     name="ml4ir",
     packages=find_namespace_packages(include=["ml4ir.*"]),
     version="0.0.1",
     description="Machine Learning libraries for Information Retrieval",
+    long_description=getReadMe(),
+    long_description_content_type="text/markdown",
     author="Search Relevance, Salesforce",
-    author_email="searchrelevancyscrumteam@salesforce.com ",
-    data_files=[("ml4ir/build", ["ml4ir/build/Dockerfile"])],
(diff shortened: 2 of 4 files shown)


darshshah:
Created a whl distribution that can be uploaded to pypi:
$ python3 setup.py sdist bdist_wheel
....
This creates dist:
dist/
- ml4ir-0.0.1-py3-none-any.whl
- ml4ir-0.0.1.tar.gz



Checked the distribution with Twine:

$ twine check dist/*
Checking dist/ml4ir-0.0.1-py3-none-any.whl: PASSED
Checking dist/ml4ir-0.0.1.tar.gz: PASSED

python/setup.py
```diff
@@ -5,14 +5,29 @@ def load_dependencies():
...
     version="0.0.1",
     description="Machine Learning libraries for Information Retrieval",
+    long_description=getReadMe(),
+    long_description_content_type="text/markdown",
     author="Search Relevance, Salesforce",
-    author_email="searchrelevancyscrumteam@salesforce.com ",
-    data_files=[("ml4ir/build", ["ml4ir/build/Dockerfile"])],
+    author_email="searchrelevancyscrumteam@salesforce.com",
+    url="https://www.salesforce.com/",
+    classifiers=[
```
balikasg, line 24: are you following a tutorial? Have you done this in the past?
Looks good to me, but I have not release at pip

darshshah: Yes, I am following https://packaging.python.org/tutorials/packaging-projects/

python/setup.py
```diff
@@ -5,14 +5,29 @@ def load_dependencies():
...
+        "License :: OSI Approved :: Apache Software License",
+        "Programming Language :: Python :: 3 :: Only"
+    ],
+    include_package_data=True,
+    data_files=[("ml4ir/build", ["ml4ir/build/Dockerfile", "ml4ir/build/run_driver.sh"])],
```
balikasg, line 29: should we add the `docker-compose.yml` here? Or doing `pip install` is mutually exclusive with `docher-compose up`. In general, after doing `pip install` I am not able to run the scripts shown in the remaining of the README. Which makes sense, because the paths indicated there are obsolete. We need to fix this I guess. Disclaimer: I am a docker beginner, I need to learn more about it.

darshshah: From what I understand, to use docker we need to git clone the repo and not pip install. @lastmansleeping might have more insights into this

python/README.md
```diff
@@ -11,10 +11,10 @@
 ml4ir can be installed as a pip package by using the following command
```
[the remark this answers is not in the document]
darshshah: Yeah, we can add this when we do 0.0.2 release.

python/setup.py
lastmansleeping: I don't think there is a need to add `docker-compose.yml` to setup. We can also pull the `Dockerfile` now from the virtualenv/pythonpath because we have made sure to add it to the pip installable ml4ir package. So it should be good.

balikasg: ok, thanks!

PR #43, opened 2020-06-30
updates of tf-writer

balikasg:
This PR updates the code needed for writing TF records from csv to files. 
I started diving to the library trying to write tfrecords for the iris dataset (see notebook also). In this current version, the code has a few bugs that I try to fix here. Same for the notebook example PointwiseRankingDemo.ipynb that does not run anymore.

My idea, if these are merged, is to have few such notebook with public data that document the process. Also, my first PR to `ml4ir`!!

### DIFF
--- a/python/ml4ir/base/data/tfrecord_helper.py
+++ b/python/ml4ir/base/data/tfrecord_helper.py
@@ -42,12 +42,11 @@ def get_example_proto(row, features):
     """
 
     features_dict = dict()
-
     for feature_info in features:
         feature_name = feature_info["name"]
         feature_fn = _get_feature_fn(feature_info["dtype"])
-        features_dict[feature_name] = feature_fn([row[feature_name]])
-
+        value_upcasted = row[feature_name].astype(feature_info["dtype"])
+        features_dict[feature_name] = feature_fn([value_upcasted])
     return train.Example(features=train.Features(feature=features_dict))
 
 
--- a/python/ml4ir/base/data/tfrecord_helper.py
+++ b/python/ml4ir/base/data/tfrecord_helper.py
@@ -45,6 +45,11 @@ def get_example_proto(row, features):
     for feature_info in features:
         feature_name = feature_info["name"]
         feature_fn = _get_feature_fn(feature_info["dtype"])
+        # When applying functions with axis=1, pandas performs upcasting,
+        # so if we have a mix of floats/ints, converts everything to float
+        # that breaks this part of the code. Example:
+        # https://stackoverflow.com/questions/47143631/
+        # how-do-i-preserve-datatype-when-using-apply-row-wise-in-pandas-dataframe
         value_upcasted = row[feature_name].astype(feature_info["dtype"])
         features_dict[feature_name] = feature_fn([value_upcasted])
     return train.Example(features=train.Features(feature=features_dict))
--- a/python/ml4ir/base/data/tfrecord_writer.py
+++ b/python/ml4ir/base/data/tfrecord_writer.py
@@ -1,8 +1,43 @@
+"""
+Writes data in Example or SequenceExample protobuf (tfrecords) format.
+
+To use it as a standalone script, refer to the argument spec
+at the bottom
+
+Syntax to convert a single or several CSVs:
+python ml4ir/base/data/tfrecord_writer.py \
+sequence|example \
+--csv-files <SPACE_SEPARATED_PATHS_TO_CSV_FILES> \
+--out-dir <PATH_TO_OUTPUT_DIR> \
+--feature_config <PATH_TO_YAML_FEATURE_CONFIG> \
+--keep-single-files
+
+or to convert all CSV files in a dir
+python ml4ir/base/data/tfrecord_writer.py \
+sequence|example \
+--csv-dir <DIR_WITH_CSVs> \
+--out-dir <PATH_TO_OUTPUT_DIR> \
+--feature_config <PATH_TO_YAML_FEATURE_CONFIG> \
+--keep-single-files
+
+Setting --keep-single-files writes one tfrecord file
+for each CSV file (better performance). If not set,
+joins everything to a single tfrecord file.
+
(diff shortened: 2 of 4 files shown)


python/ml4ir/base/data/tfrecord_helper.py
```diff
@@ -42,12 +42,16 @@ def get_example_proto(row, features):
...
         feature_name = feature_info["name"]
         feature_fn = _get_feature_fn(feature_info["dtype"])
-        features_dict[feature_name] = feature_fn([row[feature_name]])
-
+        # When applying functions with axis=1, pandas performs upcasting,
+        # so if we have a mix of floats/ints, converts everything to float
+        # that breaks this part of the code. Example:
+        # https://stackoverflow.com/questions/47143631/
+        # how-do-i-preserve-datatype-when-using-apply-row-wise-in-pandas-dataframe
+        value_upcasted = row[feature_name].astype(feature_info["dtype"])
```
balikasg, line 53: I documented inline the change.. Essentially, if you have `float` and `int` in a row, the row is converted to floats. This results in failure here, as the protos use the type in the config.

python/ml4ir/base/data/tfrecord_writer.py
```diff
@@ -1,8 +1,43 @@
+"""
```
balikasg, line 1: This is an initiative to update the file. It uses a bit my taste, but I made sure to not break the `write_from_files` and `write_from_file` function APIs. But I changed the API of the command line tool, adding support for both tf.example and tf.sequence etc.

python/ml4ir/base/data/tfrecord_writer.py
```diff
@@ -104,100 +118,94 @@ def write_from_df(
...
 
+def define_arguments():
+    first_doc_line = __doc__.strip().split("\n")[0]
+    parser = ArgumentParser(description=first_doc_line)
+    parser.add_argument(
+        'tfmode',
+        choices=MODES,  # Choices with a dict, shows the keys
+        help="select between `sequence` and `example` to write tf.Example or tf.SequenceExample")
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument(
```
balikasg, line 178: This is an important change. I am adding support for converting several files at once that are explicitly listed or all csv files of a dir. But either the one, or the other, both does not work and the code returns an error..

python/ml4ir/base/data/tfrecord_writer.py
```diff
@@ -104,100 +118,94 @@ def write_from_df(
...
+        "--csv-dir", type=str,  help="Directory with CSV files; every .csv file will be converted."
+    )
+    group.add_argument(
+        "--csv-files", type=str, nargs="+", help="A single or more (space separated) CSV files to be converted."
+    )
+    parser.add_argument(
```
balikasg, line 184: Here, this is another difference: to maintain the API somewhat simple, for writing the files the user can not choose a name. If several files, we will write in the out-dir changing the filenames extension from csv to tfrecords. I believe this is better from the previous solution because you did not know the association from fnames to numbered outputs (out1.tfrecords, out2.tfrecord,..). When though, a single combined tfrecords from multiple files is to be written, it is called `combined.tfrecord`.

python/ml4ir/base/features/feature_config.py
```diff
@@ -97,19 +97,19 @@ def extract_features(self, features_dict, logger: Optional[Logger] = None):
         except KeyError:
             self.query_key = None
             if logger:
-                logger.warning("'query_key' key not found in the feature_config specified")
+                logger.warning("'%s' key not found in the feature_config specified" % FeatureConfigKey.QUERY_KEY)
```
balikasg, line 100: Here we have a class a few lines above that specifies these names, so better to not hardcode them.

python/ml4ir/base/data/tfrecord_helper.py
lastmansleeping: Can you add a `FIXME` tag perhaps? But yeah, looks like there might not be a fix?

lastmansleeping: I've not had any issue with int and floats though.

python/ml4ir/base/data/tfrecord_writer.py
lastmansleeping: Wonderful. I missed updating this as part of my huge refactor work.

python/ml4ir/base/data/tfrecord_writer.py
lastmansleeping: The support was already there, right?

lastmansleeping: Okay. I see what you are doing. Did not know you could define mutually exclusive arguments. Nice! I like it.

python/ml4ir/base/features/feature_config.py
lastmansleeping: Good catch.

python/ml4ir/base/data/tfrecord_helper.py
balikasg: There is nothing to fix here. Line 53 is the fix. I had the issue and spent quite some time to debug. Perhaps different Pandas versions..

python/ml4ir/base/data/tfrecord_writer.py
balikasg: Yes the support was there. The change just make it explicit in the arguments (like you can have either one or the other). If you try to pass both, it fails giving you a clear failure message. 

The other change is in `--csv-files` where you can pass several files, space separated. This simplifies the code in line 140.

balikasg:
@lastmansleeping should I proceed, "Squash and merge" and become a proud `ml4ir` contributor?

lastmansleeping:
> @lastmansleeping should I proceed, "Squash and merge" and become a proud `ml4ir` contributor?

Do it!

PR #44, opened 2020-07-13
Enabling Spark based HDFS I/O for running ml4ir @rev balikasg@

lastmansleeping:
The PR includes changes required to read the following:
- data (any directory -> copied over to local FS)
- feature config + model config (any YAML/JSON file)
- dataframe(eg: vocabulary files)

and writing the following back to HDFS:
- models
- logs, metrics, etc

### DIFF
--- a/python/ml4ir/applications/ranking/data/scripts/create_dataset.py
+++ b/python/ml4ir/applications/ranking/data/scripts/create_dataset.py
@@ -11,29 +11,32 @@ import pandas as pd
 import numpy as np
 from logging import Logger
 
-from ml4ir.base.io import file_io
+from ml4ir.base.io.local_io import LocalIO
 from ml4ir.base.io import logging_utils
-from ml4ir.base.features.feature_config import parse_config, FeatureConfig
+from ml4ir.base.features.feature_config import FeatureConfig
+from ml4ir.base.config.keys import TFRecordTypeKey
 
 # Defaults
-DATA_DIR = 'ml4ir/applications/ranking/data/'
-OUT_DIR = 'ml4ir/applications/ranking/data/synthetic/'
-FEATURE_CONFIG = 'ml4ir/applications/ranking/data/feature_config.yaml'
-FEATURE_HIGHVAL = {'text_match_bool': [0, 1]}
-FEATURE_NUM_RESULTS = 'num_results'
+DATA_DIR = "ml4ir/applications/ranking/data/"
+OUT_DIR = "ml4ir/applications/ranking/data/synthetic/"
+FEATURE_CONFIG = "ml4ir/applications/ranking/data/feature_config.yaml"
+FEATURE_HIGHVAL = {"text_match_bool": [0, 1]}
+FEATURE_NUM_RESULTS = "num_results"
 MAX_NUM_RECORDS = 50
 NUM_SAMPLES = 10
 RANDOM_STATE = 123
 
 
-def run_dataset_creation(data_dir: str = DATA_DIR,
-                         out_dir: str = OUT_DIR,
-                         feature_config: str = FEATURE_CONFIG,
-                         feature_highval: dict = FEATURE_HIGHVAL,
-                         feature_num_results: str = FEATURE_NUM_RESULTS,
-                         max_num_records: int = MAX_NUM_RECORDS,
-                         num_samples: int = NUM_SAMPLES,
-                         random_state: int = RANDOM_STATE):
+def run_dataset_creation(
+    data_dir: str = DATA_DIR,
+    out_dir: str = OUT_DIR,
+    feature_config_path: str = FEATURE_CONFIG,
+    feature_highval: dict = FEATURE_HIGHVAL,
+    feature_num_results: str = FEATURE_NUM_RESULTS,
+    max_num_records: int = MAX_NUM_RECORDS,
+    num_samples: int = NUM_SAMPLES,
+    random_state: int = RANDOM_STATE,
+):
     """
     1. Loads example data
     2. Builds specified synthetic data size by sampling from example data
@@ -41,33 +44,43 @@ def run_dataset_creation(data_dir: str = DATA_DIR,
     4. For now, write out to CSV. In future could return df directly
     """
     # Setup logging
-    logger: Logger = setup_logging()
+    file_io = LocalIO()
+    logger: Logger = setup_logging(file_io)
+    file_io.set_logger(logger)
 
     try:
         # Set seeds
         set_seeds(random_state)
-        logger.info('Set seeds with initial random state {}'.format(random_state))
+        logger.info("Set seeds with initial random state {}".format(random_state))
 
         # Load and parse feature config
-        feature_config: FeatureConfig = parse_config(tfrecord_type='', feature_config=feature_config, logger=logger)
+        feature_config: FeatureConfig = FeatureConfig.get_instance(
+            tfrecord_type=TFRecordTypeKey.SEQUENCE_EXAMPLE,
+            feature_config_dict=file_io.read_yaml(feature_config_path),
+            logger=logger,
+        )
         logger.info("Feature config parsed and loaded")
 
         # Create output location
-        file_io.make_directory(out_dir, log=logger)
-        out_file = os.path.join(out_dir, 'synthetic_data_{}.csv'.format(dt.datetime.now().strftime('%Y%m%d-%H%M%S')))
+        file_io.make_directory(out_dir)
+        out_file = os.path.join(
+            out_dir, "synthetic_data_{}.csv".format(dt.datetime.now().strftime("%Y%m%d-%H%M%S"))
+        )
 
         # Build data
-        seed_data = load_seed_data(data_dir, logger)
+        seed_data = load_seed_data(data_dir, logger, file_io)
 
-        df_synthetic = fill_data(seed_data,
-                                 max_num_records,
-                                 feature_config,
-                                 feature_highval,
-                                 feature_num_results,
-                                 num_samples,
-                                 logger)
+        df_synthetic = fill_data(
+            seed_data,
+            max_num_records,
+            feature_config,
+            feature_highval,
+            feature_num_results,
+            num_samples,
+            logger,
+        )
         file_io.write_df(df_synthetic, outfile=out_file, index=False)
-        logger.info('Synthetic data created! Location: {}'.format(out_file))
+        logger.info("Synthetic data created! Location: {}".format(out_file))
         return df_synthetic
 
     except Exception as e:
@@ -76,16 +89,16 @@ def run_dataset_creation(data_dir: str = DATA_DIR,
         return
 
 
-def setup_logging():
+def setup_logging(file_io: LocalIO):
     run_id = "-".join([socket.gethostname(), time.strftime("%Y%m%d-%H%M%S")])
-    logs_dir: str = os.path.join('logs', run_id)
-    file_io.make_directory(logs_dir, clear_dir=True, log=None)
+    logs_dir: str = os.path.join("logs", run_id)
+    file_io.make_directory(logs_dir, clear_dir=True)
 
     outfile: str = os.path.join(logs_dir, "output_log.csv")
     logger = logging_utils.setup_logging(reset=True, file_name=outfile, log_to_file=True)
 
-    logger.info('Logging initialized. Saving logs to : {}'.format(logs_dir))
-    logger.info('Run ID: {}'.format(run_id))
+    logger.info("Logging initialized. Saving logs to : {}".format(logs_dir))
+    logger.info("Run ID: {}".format(run_id))
     return logger
 
 
@@ -95,7 +108,7 @@ def set_seeds(random_state):
     random.seed(random_state)
 
 
-def load_seed_data(data_dir, logger):
+def load_seed_data(data_dir, logger, file_io: LocalIO):
     # Start with loading in CSV format only
     if file_io.path_exists(data_dir):
         dfs = file_io.get_files_in_directory(data_dir)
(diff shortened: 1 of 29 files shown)


lastmansleeping:
@balikasg SIP MLS YAML manifest to test:
```
jobName: sample-ml4ir-pyspark-app
jobType: spark
base:
  resources:
    driver:
      cores: 2
      memory(inGB): 4
  parameters:
    image: "ops0-artifactrepo1-0-prd.data.sfdc.net/docker-sam/ashish.srinivasa/ml4ir-pyspark:0.1.6"
    runtimeArgs:
      - --data_dir hdfs://prod-EMPTY-hdaas/projects/search_relevance/ashish.srinivasa/data/processed_data/228/test/tfrecord
      - --feature_config hdfs://prod-EMPTY-hdaas/projects/search_relevance/ashish.srinivasa/data/processed_data/228/test/config/feature_config.yaml
      - --run_id test
      - --data_format tfrecord
      - --execution_mode train_inference_evaluate
      - --models_dir hdfs://prod-EMPTY-hdaas/projects/search_relevance/ashish.srinivasa/models/228
      - --logs_dir hdfs://prod-EMPTY-hdaas/projects/search_relevance/ashish.srinivasa/logs/228
    deps:
      pyFiles:
        - "local:///opt/spark/work-dir/*"
    sparkConf:
      spark.driver.extraJavaOptions: "-Djava.security.krb5.conf=/etc/krb5.conf"
      spark.kerberos.keytab: "/opt/spark/work-dir/ml4ir/python/tmp.keytab"
    type: "Python"
    file: "local:///opt/spark/work-dir/ml4ir/python/ml4ir/applications/ranking/pipeline.py"
```

We are testing all of the I/O mentioned in the original description here, namely, data, configs, vocabulary files, logs, metrics.

python/ml4ir/base/io/spark_io.py
```diff
@@ -0,0 +1,119 @@
...
+
+def get_hadoop_config():
+    """Get hadoop configuration object"""
+    if not SparkConfigHolder.HADOOP_CONFIG:
+        SparkConfigHolder.HADOOP_CONFIG = get_spark_context()._jsc.hadoopConfiguration()
+
+    return SparkConfigHolder.HADOOP_CONFIG
+
+
+def get_hdfs():
```
lastmansleeping, line 40: @balikasg This is the main module to review. I had to jump some hoops to get access to the hadoop File system since it is only available in spark's java based APIs.

python/ml4ir/base/data/relevance_dataset.py
```diff
@@ -25,16 +27,26 @@ def __init__(
...
         parse_tfrecord: bool = True,
-        logger: Logger = None,
+        logger: Optional[Logger] = None,
     ):
         self.feature_config = feature_config
         self.max_sequence_size = max_sequence_size
-        self.data_dir: str = data_dir
+        self.logger = logger
+
+        # If data directory is a HDFS path, first copy to local file system
```
jakemannix, line 36: I think maybe we can reorganize this a bit.  This has references to sparkIO directly in this class, which we don't want.  We should encapsulate the IO into a class (ahh! Jake is suggesting a class?!?) or a function which is passed in.

We can chat about that later on if it's not clear.

lastmansleeping: (Haha. I went overboard with NOT making spark_io a class this time.)
I'm not sure how we can eliminate a reference to spark_io here though. I can wrap it under `file_io.copy_dir(src, dest)` and then internally call `spark_io` like I do for the other I/O codepaths, if you like.

Does that work?

jakemannix: what I mean is you could pass into RelevanceDataset's constructor, an object or function which provides IO.  If it's a object which you pass in, the (abstract) class of the object would be DataIO (or something like that), and some particular implementations would be LocalIO and SparkIO.  Then RelevanceDataset has no references to spark, but whoever instantiates a RelevanceDataset may choose to create a spark-based DataIO and pass it into the constructor.

lastmansleeping: I understand what you are suggesting. But I don't get how you would do this specific task - "copying files from HDFS to Local filesystem", which is specific for the spark_io case.

lastmansleeping: Keep in mind that for tfrecord, we don't explicitly(using file_io or spark_io) do any read at all. It's handled by tensorflow.

lastmansleeping: Alternatively, I can do the copying in the main pipeline and force RelevanceDataset to work with only file_io(or local file I/O). This might be cleaner. My reasoning for adding the copying to RelevanceDataset was that if a user wanted to create a RelevanceDataset using data stored on HDFS, then they already can do it. They don't have to write their own version of spark_IO. May be passing in the file handler object into RelevanceDataset is a good middle ground. Not entirely convinced.

lastmansleeping:
@balikasg Synced with @jakemannix offline, and he requested a few structural changes to the code. Will update you when done, but not changing any functionality. So you can take a look at it and try training models.

python/ml4ir/base/config/keys.py
```diff
@@ -98,3 +98,12 @@ class EncodingTypeKey(Key):
...
     CNN = "cnn"
+
+
+class DefaultDirectoryKey(Key):
+    """Default directory paths"""
+
+    MODELS = "models"
+    LOGS = "logs"
+    DATA = "data"
+    TEMP_DATA = "data/temp"
```
balikasg, line 109: since they are temp, should we move this to `/tmp/data`? do we have cleaning in place?

balikasg: although I don't know if flowsnake has /tmp

python/ml4ir/base/features/feature_config.py
```diff
@@ -465,13 +473,11 @@ def parse_config(
     tfrecord_type: str, feature_config, logger: Optional[Logger] = None
 ) -> FeatureConfig:
     if feature_config.endswith(".yaml"):
-        feature_config = file_io.read_yaml(feature_config)
-        if logger:
-            logger.info("Reading feature config from YAML file : {}".format(feature_config))
```
balikasg, line 470: I am just wondering here. Not for this PR. 
But `if logger` repeats a lot in ml4ir. Why not agree we always need a logger, and instantiate one form the beginning? Or decorate key functions/classes to avoid such checks.

python/ml4ir/base/features/feature_config.py
```diff
@@ -465,13 +473,11 @@ def parse_config(
...
+        feature_config = file_io.read_yaml(feature_config, log=logger)
     else:
         feature_config = yaml.safe_load(feature_config)
-        if logger:
-            logger.info("Reading feature config from YAML string")
+    if logger:
+        logger.info("Feature Config \n{}".format(json.dumps(feature_config, indent=4)))
```
balikasg, line 480: btw these are very nice, but not related to the functionality this PR adds!
thanks for the effort though! We could also consider this as logger debug, as it is often long

python/ml4ir/base/io/file_io.py
```diff
@@ -61,26 +64,17 @@ def read_df(
...
-    if infile.startswith("hdfs"):
-        # replace pydoop
-        # fp = hdfs.open(infile)
-        path_type = "HDFS"
+    if log_path:
+        log.info("Loading dataframe from path : {}".format(infile))
+
+    if infile.startswith(HDFS_PREFIX):
+        # NOTE: Move to fully spark dataframe based operations in the future
+        return spark_io.read_df(infile).toPandas()
```
balikasg, line 72: is this lazy? Does it scale or OOM?

python/ml4ir/base/io/file_io.py
```diff
@@ -61,26 +64,17 @@ def read_df(
...
     elif infile.endswith(".gz"):
         fp = gzip.open(os.path.expanduser(infile), "rb")
```
balikasg, line 74: I have the impression that pandas handles `gz` files out of the box.

python/ml4ir/base/io/file_io.py
```diff
@@ -152,17 +146,22 @@ def write_df(df, outfile: str = None, sep: str = ",", index: bool = True, log=No
...
+        )
+        output = output.replace("\\", "\\\\")
 
-    if outfile:
-        fp = open(outfile, "w")
-        fp.write(output)
-        fp.close()
-    return output
+        if outfile:
+            fp = open(outfile, "w")
```
balikasg, line 161: Here using `with open()...` is a more pythonic way. Why not directly write with pd.to_csv? to replace?

python/ml4ir/base/io/spark_io.py
balikasg: thanks for the pointer!
From what I understand it is not a Flowsnake v2 thing, we are expecting HDFS paths and we are handling things as this.

python/ml4ir/base/pipeline.py
```diff
@@ -133,9 +145,20 @@ def validate_args(self):
         return self
 
     def finish(self):
-        # Delete temp directories
+        # Delete temp data directories
         if self.data_format == DataFormatKey.CSV:
             file_io.rm_dir(os.path.join(self.data_dir, "tfrecord"))
+        file_io.rm_dir(DefaultDirectoryKey.TEMP_DATA)
```
balikasg, line 151: Just saw this, there is cleaning!

python/ml4ir/base/config/keys.py
lastmansleeping: I explicitly delete the temp directory once done.

python/ml4ir/base/features/feature_config.py
lastmansleeping: Yeah. Let me file an issue for this. It looks to me like tech debt building.

python/ml4ir/base/features/feature_config.py
lastmansleeping: True. Can change.

python/ml4ir/base/io/file_io.py
lastmansleeping: This is not lazy. The goal is to only use this for small files like vocabulary, etc which you really need in memory. For training data, we should use the lazy loading provided by TFRecordDataset.

python/ml4ir/base/io/file_io.py
lastmansleeping: I used this method from our other projects. So my memory might not be completely correct - I don't think pandas can read `gz` files with the C engine. Can check though.

python/ml4ir/base/io/file_io.py
lastmansleeping: After spending a lot of time with our ranking data, this combination of `read_df` and `write_df` were the only ones that were compatible with each other and the upstream spark jobs. If you notice, I need to replace `\\` with `\\\\` on the next line for compatibility with re-reading. I can add a `FIXME` here for the future though.

lastmansleeping: (Again, this method was written a year ago for some of our other projects. Might not be necessary to do all this hacky read write for ml4ir)

python/ml4ir/base/io/file_io.py
```diff
@@ -1,298 +1,181 @@
...
-        log: logging object
-
-    Returns:
-        dataframe in csv form if outfile is None
-
-    NOTE: Does not support spark write
-    """
-    if outfile and outfile.startswith(HDFS_PREFIX):
+
+class FileIO(object):
```
balikasg, line 7: I like this! Defining FileIO and then having the LocalIO and SparkIO, is clean, slick etc. nice!

python/ml4ir/base/io/local_io.py
```diff
@@ -0,0 +1,251 @@
...
+
+from typing import Optional
+
+from io import StringIO  # type: ignore
+
+
+class LocalIO(FileIO):
+    """Class defining the file I/O handler methods for the local file system"""
+
+    def make_directory(self, dir_path: str, clear_dir: bool = False) -> str:
```
balikasg, line 22: `clear_dir` is an overloaded name in this implementation. There are both functions and variables defined with the same name. Can we consider changing one of the two? For example, this bool arg can be renamed to `remove_dir_content` or something else equally descriptive. `clear_dir_content` would also work I guess.

python/ml4ir/base/io/spark_io.py
```diff
@@ -1,119 +1,227 @@
...
+        self.logger = logger
+        self.spark_session = SparkSession.builder.appName("ml4ir").getOrCreate()
+        self.spark_context = self.spark_session.sparkContext
+        self.hadoop_config = self.spark_context._jsc.hadoopConfiguration()
+        self.hdfs = self.spark_context._gateway.jvm.org.apache.hadoop.fs.FileSystem.get(
+            self.hadoop_config
+        )
+        self.local_fs = self.hdfs.getLocal(self.hadoop_config)
+
+    def make_directory(self, dir_path: str, clear_dir: bool = False) -> str:
```
balikasg: same here with clear_dir..

python/ml4ir/base/io/spark_io.py
```diff
@@ -1,119 +1,227 @@
...
+        """
+        Read JSON file and return a python dictionary
+
+        Args:
+            infile: path to the json file; can be hdfs path
+
+        Returns:
+            python dictionary
+        """
+        self.log("Reading JSON file : {}".format(infile))
```
balikasg, line 74: nice usage of `self.log` here! +1

python/ml4ir/base/io/spark_io.py
```diff
@@ -1,119 +1,227 @@
...
+
+        Args:
+            infiles: paths to the csv input files; can be hdfs paths
+            sep: separator to use for loading file
+            index_col: column to be used as index
+
+        Returns:
+            pandas dataframe
+        """
+        raise NotImplementedError
```
balikasg: SparkIO inherits from FileIO, isn't it redundant to overwrite notImplemented methods that were defined there? Any particular reason for it? it occurs a few times in SparkIO.

balikasg: Another question (if you know): just passing the list of files, wouldn't do the trick? Like: 
```
self.spark_session.read.csv
            .option("header", "true")
            .option("inferschema", "true")
            .option("mode", "DROPMALFORMED")
            .option("mergeSchema", "true")
            .load(infiles)
            .toPandas()
```
I am looking [at this question](https://stackoverflow.com/questions/37639956/how-to-import-multiple-csv-files-in-a-single-load).

lastmansleeping:
> This PR looks good, I like the new design!
> I have left a few non-blocking comments, so I am approving

Thanks @balikasg . Will address the comments and fix the test failures and then merge the PR.

python/ml4ir/base/io/local_io.py
lastmansleeping: Fixed.

python/ml4ir/base/io/spark_io.py
lastmansleeping: Yeah. Added.

python/ml4ir/base/io/spark_io.py
lastmansleeping: Added.

lastmansleeping:
@jakemannix I have made the requested changes. Had to change a lot of files though.

python/ml4ir/base/features/feature_config.py
jakemannix: I agree - we should just always have a Logger - if none is configured, use a dummy / noop logger or something.

python/ml4ir/base/pipeline.py
```diff
@@ -36,19 +39,44 @@ def __init__(self, args: Namespace):
...
 
-        # Setup directories
-        self.models_dir: str = os.path.join(self.args.models_dir, self.run_id)
-        self.data_dir: str = self.args.data_dir
-        file_io.make_directory(self.models_dir, clear_dir=False, log=self.logger)
+        # Set the file handlers and respective setup
+        if self.args.file_handler == FileHandlerKey.LOCAL:
+            self.file_io = self.local_io
+        elif self.args.file_handler == FileHandlerKey.SPARK:
+            self.file_io = SparkIO(self.logger)
```
jakemannix, line 75: Ok so this is basically what I was looking for, yes.  Technically, we need to make one last step of separation: pull spark_io.py out of ml4ir/base and move it into a new python package (ml4ir-spark), and perhaps put the pipeline in its own package as well (ml4ir-apps), and show how you could allow people to run pipelines without having to `pip install pyspark` at all, if not needed. 

But folding that into the next ticket should be fine.

python/ml4ir/applications/ranking/data/scripts/create_dataset.py
```diff
@@ -217,26 +249,28 @@ def fill_data(seed_data, max_num_records, feature_config, feature_highval, featu
...
     parser = argparse.ArgumentParser()
-    parser.add_argument('-data_dir', default=DATA_DIR)
-    parser.add_argument('-out_dir', default=OUT_DIR)
-    parser.add_argument('-feature_config', default=FEATURE_CONFIG)
-    parser.add_argument('-feature_highval', default=FEATURE_HIGHVAL)
-    parser.add_argument('-feature_num_results', default=FEATURE_NUM_RESULTS)
-    parser.add_argument('-max_num_records', default=MAX_NUM_RECORDS)
-    parser.add_argument('-num_samples', default=NUM_SAMPLES)
-    parser.add_argument('-random_state', default=RANDOM_STATE)
+    parser.add_argument("-data_dir", default=DATA_DIR)
```
balikasg, line 267: help messages?

PR #45, opened 2020-07-21
Updates to PointwiseRankingDemo notebook

lastmansleeping:
We need to merge these changes to train other models on ml4ir. 

Once you add test data for entity prediction, 
**I will add the notebook I demoed today.** @Ullimague 

P.S. : I've also added images used in the notebook to the repository and deleted the HTML version of the notebook.


python/ml4ir/base/data/tfrecord_helper.py
```diff
@@ -45,13 +46,17 @@ def get_example_proto(row, features):
     for feature_info in features:
         feature_name = feature_info["name"]
         feature_fn = _get_feature_fn(feature_info["dtype"])
+        # FIXME
         # When applying functions with axis=1, pandas performs upcasting,
         # so if we have a mix of floats/ints, converts everything to float
         # that breaks this part of the code. Example:
         # https://stackoverflow.com/questions/47143631/
         # how-do-i-preserve-datatype-when-using-apply-row-wise-in-pandas-dataframe
-        value_upcasted = row[feature_name].astype(feature_info["dtype"])
```
lastmansleeping, line 53: This fix doesn't really work because `row[feature_name]` is a scalar and not a pandas dataframe or series. It is not an issue for now. But I've added a `FIXME`

python/ml4ir/base/features/feature_config.py
```diff
@@ -259,8 +260,7 @@ def define_inputs(self) -> Dict[str, Input]:
         """
 
         def get_shape(feature_info: dict):
-            # TODO: Expand to other shapes
-            return (1,)
+            return feature_info.get("shape", (1,))
```
lastmansleeping, line 263: @Ullimague Added support for specifying feature shape in the feature config. This is useful for list of entities, etc.

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -271,3 +271,101 @@ def categorical_embedding_with_vocabulary_file(feature_tensor, feature_info, fil
     )
 
     return embedding
+
+
+def categorical_indicator_with_vocabulary_file(feature_tensor, feature_info, file_io: FileIO):
```
lastmansleeping, line 276: @Ullimague Added another feature function which converts a string categorical variable into one-hot representation. Used this to convert entity labels into their one-hot label representation.

python/ml4ir/base/features/feature_fns/sequence.py
```diff
@@ -38,19 +38,21 @@ def bytes_sequence_to_encoding_bilstm(feature_tensor, feature_info, file_io: Fil
         fixed_length=feature_layer_info["args"].get("max_length", None),
     )
 
-    feature_tensor = tf.reshape(feature_tensor, [-1, feature_layer_info["args"]["max_length"]])
+    feature_tensor = tf.squeeze(feature_tensor, axis=1)
```
lastmansleeping, line 41: Minor change to make it more robust.

lastmansleeping:
@Ullimague I've also added images used in the notebook to the repository and deleted the HTML version of the notebook.

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -271,3 +271,101 @@ def categorical_embedding_with_vocabulary_file(feature_tensor, feature_info, fil
...
+    Returns:
+        Categorical one-hot representation of input feature_tensor
+
+    Args under feature_layer_info:
+        vocabulary_file: str; path to vocabulary CSV file for the input tensor
+        num_oov_buckets: int; number of out of vocabulary buckets/slots to be used to
+                         encode strings into categorical indices
+
+    NOTE:
+    The vocabulary CSV file must contain two columns - key, id,
```
Ullimague, line 295: Could you give an example? Why do we need an id? Couldn't the vocabulary file just be:
key, othercolumns_we_ignore <== header
"AAA", ...
"BBB", ...
"CCC", ...
"BBB", ...
"EEE", ...
...

?

python/ml4ir/base/features/feature_fns/categorical.py
Ullimague: Could we add a test for it?

python/ml4ir/base/features/feature_fns/categorical.py
lastmansleeping: You don't. If there is no id, then it autogenerates. Check the next line in the comment.

We can use Id to map multiple categories into the same ID, say from a preprocessing job you've found that these categories are similar, etc.

PR #46, opened 2020-07-21
Classification model

Ullimague:
Adding test data for classification


python/ml4ir/applications/classification/tests/data/csv/validation/file_0.csv
```diff
@@ -0,0 +1,1001 @@
+"query_key","query_feature_0","group_feature_1","group_sequence_feature_2","entity"
```
lastmansleeping, line 1: @Ullimague It will be useful to have more public facing names for these. My suggestion
```
query_key
query_text
domain_id
entity_id
user_context
```

python/ml4ir/applications/classification/tests/generate_test_data.py
```diff
@@ -0,0 +1,73 @@
...
+            writer = csv.writer(file, quoting=csv.QUOTE_NONNUMERIC)
+            writer.writerows(rows)
+
+
+def main():
+    generate_csv_test_data()
+    return
+
+
+if __name__ == "__main__":
```
lastmansleeping, line 72: @Ullimague Can you add a top level comment on what is the goal of this script?

python/ml4ir/applications/classification/tests/generate_test_data.py
```diff
@@ -0,0 +1,73 @@
...
+    "the", "tragedy", "of", "hamlet", "prince", "denmark", "shakespeare", "homepage", "entire", "play", "act", "i",
+    "scene", "elsinore", "a", "platform", "before", "castle", "francisco", "at", "his", "post", "enter", "to", "him",
+    "bernardo", "whos", "there", "nay", "answer", "me", "stand", "and", "unfold", "yourself", "long", "live", "king",
+    "he", "you", "come", "most", "carefully", "upon", "your", "hour", "tis", "now", "struck", "twelve", "get", "thee",
+    "bed", "for", "this", "relief", "much", "thanks", "bitter", "cold", "am", "sick", "heart", "have", "had", "quiet",
+    "guard", "not", "mouse", "stirring", "well", "good", "night", "if", "do", "meet", "horatio", "marcellus", "rivals",
+    "my", "watch", "bid", "them", "make", "haste", "think", "hear", "ho", "friends", "ground", "liegemen", "dane"
+]
+
+FEATURE_1_VOCABULARY = [str(i) for i in range(0, 20)]
```
lastmansleeping, line 23: @Ullimague Can you rename accordingly here?

Also, might be useful to use the same vocab for both?

python/ml4ir/applications/classification/tests/generate_test_data.py
```diff
@@ -0,0 +1,73 @@
...
+    group_sequence_feature_2_generator = FeatureGenerator(20, FEATURE_2_VOCABULARY, sequence_joiner=",")
+    label_generator = FeatureGenerator(1, LABEL_VOCABULARY)
+    generators = [
+        query_feature_0_generator,
+        group_feature_1_generator,
+        group_sequence_feature_2_generator,
+        label_generator
+    ]
+
+    for (path, number_rows) in [(CSV_TRAIN_FILE_PATH, 7000), (CSV_TEST_FILE_PATH, 2000),
```
lastmansleeping, line 59: @Ullimague You might want to make these numbers(shapes) as top level constants just in case you want to convert them into command line args in the future.

balikasg:
@Ullimague @lastmansleeping I appreciate the work here! Happy to see us moving forwards. 

Perhaps you synced offline and agreed on a course, so my question may sound naive.
Have we considered using standard classification datasets provided by packages we already depend on ? sklearn provides iris/20newsgroups etc. We can have them with a single click (they are also hosted). Same for imdb sentiment classification. This would remove a layer of complexity for the community.

lastmansleeping:
@Ullimague I would suggest keeping the test data small. Is there any reason you changed from 1K to 10K rows?

lastmansleeping:
> @Ullimague @lastmansleeping I appreciate the work here! Happy to see us moving forwards.
> 
> Perhaps you synced offline and agreed on a course, so my question may sound naive.
> Have we considered using standard classification datasets provided by packages we already depend on ? sklearn provides iris/20newsgroups etc. We can have them with a single click (they are also hosted). Same for imdb sentiment classification. This would remove a layer of complexity for the community.

@balikasg That would be perfect. The reason we added this was to enable prototyping with test data. Also, I can check in my EP notebook which can be used as a walkthrough of sorts to design the classification pipeline. This data will also help us write unit tests for each of the feature function components - say something like learning embeddings for user's context(list of entities previously seen), etc.

With that said, I think the classification pipeline you design should work with any public classification dataset.

balikasg:
I understand, thanks for the reply @lastmansleeping . 

I just assume that for an open project, we should provide benchmark/code for things people use a lot instead of toy data. 
It also serves for testing in the sense that if we get 60% of eval metric A and everybody else 80% we can use the information. Unfortunately, such work tends to be pushed and never prioritized (due to our voluminous free time!!) so we could decide to start with it. 

Anyway! For EP we have baselines already so it will not be a blocker.

lastmansleeping:
> I understand, thanks for the reply @lastmansleeping .
> 
> I just assume that for an open project, we should provide benchmark/code for things people use a lot instead of toy data.
> It also serves for testing in the sense that if we get 60% of eval metric A and everybody else 80% we can use the information. Unfortunately, such work tends to be pushed and never prioritized (due to our voluminous free time!!) so we could decide to start with it.
> 
> Anyway! For EP we have baselines already so it will not be a blocker.

Currently, Mohammed is working on adding a public benchmark dataset for ranking. You should do something similar for classification as well. Any of the sklearn or tfdataset ones should be good too. 

The reason for this particular data is mostly for unit testing and not end to end testing for benchmarking.

Ullimague:
> @Ullimague I would suggest keeping the test data small. Is there any reason you changed from 1K to 10K rows?
> 
> Everything else looks good! :)

It was 10k at the beginning and I made it 1k by mistake - which I reverted. I can put it to 1k if that's enough.

lastmansleeping:
> > @Ullimague I would suggest keeping the test data small. Is there any reason you changed from 1K to 10K rows?
> > Everything else looks good! :)
> 
> It was 10k at the beginning and I made it 1k by mistake - which I reverted. I can put it to 1k if that's enough.

Yeah, like @balikasg mentioned, let's just use this data for testing. For actual training benchmarks we will use a public dataset. So I think <=1K should be good.

Ullimague:
> @Ullimague @lastmansleeping I appreciate the work here! Happy to see us moving forwards.
> 
> Perhaps you synced offline and agreed on a course, so my question may sound naive.
> Have we considered using standard classification datasets provided by packages we already depend on ? sklearn provides iris/20newsgroups etc. We can have them with a single click (they are also hosted). Same for imdb sentiment classification. This would remove a layer of complexity for the community.

That's a good callout, I feel for the tests, it's important to have a variety of features such as the "user_context" column which isn't easily found in classic public datasets. We could add something like this later if we want to explicitly test the convergence of a given classification model.

I'm wondering if for the community it wouldn't be better to have dedicated examples/notebooks, instead of tests to share how to use ml4ir as a library on public example datasets.

PR #47, opened 2020-07-22
Adding the Entity Prediction notebook and configs along with TFRecord data

lastmansleeping:


python/ml4ir/applications/classification/tests/data/configs/entity_id_vocab.csv
```diff
@@ -0,0 +1,10 @@
+key
+""
```
Ullimague, line 2: Why do we need this value?

lastmansleeping: Up to you really. This allows you to differentiate between an empty entity_id in the user context versus an out of vocab entity_id.

Ullimague:
Could we add the steps in the read me to use the notebooks? @lastmansleeping

PR #50, opened 2020-07-23
Classification model - Adding model pipeline

Ullimague:


python/ml4ir/applications/classification/tests/generate_test_data.py
```diff
@@ -60,17 +73,27 @@ def generate_csv_test_data():
     ]
 
     for (path, number_rows) in [(CSV_TRAIN_FILE_PATH, TOTAL_DATA_SIZE * PARTITION_TRAIN),
-                                (CSV_TEST_FILE_PATH, TOTAL_DATA_SIZE * PARTITION_VALIDATION),
-                                (CSV_VALIDATION_FILE_PATH, TOTAL_DATA_SIZE * PARTITION_TEST)]:
-        rows = [COLUMNS_HEADER] + [['query_id_' + str(idx)] + [g.generate_feature() for g in generators]
+                                (CSV_TEST_FILE_PATH, TOTAL_DATA_SIZE * PARTITION_TEST),
```
Ullimague, line 76: Fixed an inconsistency here.

python/ml4ir/applications/classification/tests/generate_test_data.py
```diff
@@ -60,17 +73,27 @@ def generate_csv_test_data():
...
+    for (path, header, data) in VOCABULARY_PATH_HEADER_DATA:
+        rows = [header] + [[word] for word in data]
+        with open(path, 'w', newline='') as file:
+            writer = csv.writer(file, quoting=csv.QUOTE_NONNUMERIC)
+            writer.writerows(rows)
+
+
 def main():
     generate_csv_test_data()
+    generate_csv_vocabulary_data()
```
Ullimague, line 96: Adding vocabulary generation.

python/ml4ir/applications/classification/config/parse_args.py
```diff
@@ -0,0 +1,31 @@
...
+        super().set_default_args()
+        self.set_defaults(
+            tfrecord_type=TFRecordTypeKey.EXAMPLE,
+            loss_key=LossKey.CATEGORICAL_CROSS_ENTROPY,
+            metrics_keys="['" + MetricKey.CATEGORICAL_ACCURACY + "'"
+                         + ", '" + MetricKey.PRECISION + "'"
+                         + "]",
+            monitor_metric=MetricKey.CATEGORICAL_ACCURACY,
+            monitor_mode="max",
+            max_sequence_size=25,
```
Ullimague, line 25: I don't know why we need max_sequence_size here. It seems to be required?

Ullimague:
Adding possibility to:
- Automatically generate the vocabulary for the tests
- Pipeline for classification model
- Test for classification model pipeline
- Reading vocabulary files (+ not only on "key" column) (+ can specify a max number of row to consider) (still need to add tests for this).

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -119,6 +121,75 @@ def categorical_embedding_with_indices(feature_tensor, feature_info, file_io: Fi
...
+        vocabulary_keys = vocabulary_df.iloc[:,0]
+    if "max_length" in feature_info:
+        vocabulary_keys = vocabulary_keys[:feature_info["max_length"]]
+    vocabulary_keys = vocabulary_keys.fillna(feature_info["default_value"]).values
+
+    vocabulary_ids = (
+        vocabulary_df["id"].values if "id" in vocabulary_df else list(range(len(vocabulary_keys)))
+    )
+
+    num_oov_buckets = args.get("num_oov_buckets", 1)
```
Ullimague, line 167: num_oov_buckets is required later in the stage.
Does it make sens to have num_oov=1 for a label? I think we should make it optional (especially for the label in classification).

Ullimague:
This is for early feedbacks, I still need to work on the tests.

python/ml4ir/applications/classification/config/parse_args.py
lastmansleeping: You don't. You can set it to None or 0 I think. This is only used for SequenceExample.

python/ml4ir/applications/classification/tests/data/configs/entity_id_vocab.csv
```diff
@@ -1,10 +0,0 @@
-key
```
lastmansleeping, line 1: @Ullimague [Minor] Why not use `git mv`. It will keep a cleaner PR.

python/ml4ir/base/features/feature_fns/categorical.py
lastmansleeping: You can set it to 0 in the feature config, no? But to keep it from breaking, would be useful to set to 1 nonetheless. Up to you.

jakemannixOfek: labels don't have an embedding layer, do they?  No need for oov?

python/ml4ir/applications/classification/config/parse_args.py
jakemannix: do we not default to None already, @lastmansleeping ?  We shouldn't require this arg at all for Example-based setups.

lastmansleeping: It is defaulted to 0. I don't understand the error that @Ullimague is facing.

python/ml4ir/base/features/feature_fns/categorical.py
lastmansleeping: Yep. Why are you passing label through an embedding layer? @Ullimague

python/ml4ir/applications/classification/config/parse_args.py
Ullimague: My bad, it was the group_metrics_min_queries line 26 that is required.

python/ml4ir/base/features/feature_fns/categorical.py
Ullimague: I can put default 0 here, but not from the config, because from the config we end-up having:
```
    if num_oov_buckets <= 0:
      raise ValueError("oov_buckets must be > 0.")
```
from tensorflow StaticVocabularyTable
called [here](https://github.com/salesforce/ml4ir/blob/37540c26c01585264f125f9eaeb4db06c44d7486/python/ml4ir/base/features/feature_fns/categorical.py#L156)

Ullimague: This comment is for line 438. I mixed the 2 functions because it didn't appeared as modified in the review..

Ullimague:
Added tests for the global methods added:
- ml4ir.applications.ranking.tests.test_feature_layer.RankingModelTest#test_categorical_embedding_to_encoding_bilstm
- ml4ir.applications.ranking.tests.test_feature_layer.RankingModelTest#test_categorical_embedding_to_encoding_bilstm_file_truncation
- ml4ir.applications.ranking.tests.test_feature_processing.RankingModelTest#test_get_one_hot_vectorizer
- ml4ir.applications.ranking.tests.test_feature_processing.RankingModelTest#test_split_and_pad_string

Ullimague:
For later, I feel ideally:
- a bunch of test could be moved to base/test instead of ranking/test
- have proper folders for tests (eg. one folder for base/test/feature_processing and then have 1 file per function tested), because currently we test too many things per test_ function.

Ullimague:
End-to-end from model training is working.

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -119,6 +121,77 @@ def categorical_embedding_with_indices(feature_tensor, feature_info, file_io: Fi
     return embedding
 
 
+def categorical_embedding_to_encoding_bilstm(feature_tensor, feature_info, file_io: FileIO):
```
lastmansleeping, line 124: @Ullimague Can we refactor this to do only categorical embedding on indices and then have another function to convert categories to indices from vocabulary file? This would allow us to reuse the same function for the char embedding in `sequence.py` as well.

Ullimague: Refactored: extracted vocabulary read and bilstm part @lastmansleeping

python/ml4ir/applications/classification/config/parse_args.py
```diff
@@ -0,0 +1,30 @@
...
+class ClassificationArgParser(RelevanceArgParser):
+    """
+    States default arguments for classification model.
+    """
+    def set_default_args(self):
+        super().set_default_args()
+        self.set_defaults(
+            tfrecord_type=TFRecordTypeKey.EXAMPLE,
+            loss_key=LossKey.CATEGORICAL_CROSS_ENTROPY,
+            metrics_keys="['" + MetricKey.CATEGORICAL_ACCURACY + "'"
```
lastmansleeping, line 20: @Ullimague @balikasg Why are you not using MRR and ACR for EP?

python/ml4ir/applications/classification/model/losses/categorical_cross_entropy.py
```diff
@@ -0,0 +1,35 @@
...
+        raise NotImplementedError
+
+
+class CategoricalCrossEntropy(RelevanceLossBase):
+
+    def get_loss_fn(self, **kwargs):
+        """
+        Define a softmax cross entropy loss
+        """
+        cce = losses.CategoricalCrossentropy(reduction=losses.Reduction.SUM_OVER_BATCH_SIZE)
```
lastmansleeping, line 27: @Ullimague Would be good to check if you want to set this particular reduction. Usually, if your loss value is very high, you can change it to mean.

python/ml4ir/applications/classification/pipeline.py
```diff
@@ -0,0 +1,131 @@
...
+        Creates RelevanceModel suited for classification use-case.
+
+        NOTE: Override this method to create custom loss, scorer, model objects.
+        """
+
+        # Define interaction model
+        interaction_model: InteractionModel = UnivariateInteractionModel(
+            feature_config=self.feature_config,
+            feature_layer_keys_to_fns=feature_layer_keys_to_fns,
+            tfrecord_type=TFRecordTypeKey.EXAMPLE,
```
lastmansleeping, line 42: @Ullimague Might be good to not hard code tfrecord_type, no?

lastmansleeping: You can set it as a default argument.

python/ml4ir/applications/classification/pipeline.py
```diff
@@ -0,0 +1,131 @@
...
+            file_io=self.file_io,
+        )
+
+        # Define metrics objects from metrics keys
+        metrics: List[Union[Type[Metric], str]] = [
+            metric_factory.get_metric(metric_key=metric_key) for metric_key in self.metrics_keys
+        ]
+
+        # Define optimizer
+        optimizer: Optimizer = get_optimizer(
```
lastmansleeping, line 66: @Ullimague @balikasg Adding cyclic learning rate to ml4ir would be awesome(and reusable).
CC: @jakemannix

python/ml4ir/applications/classification/pipeline.py
```diff
@@ -0,0 +1,131 @@
...
+            optimizer_key=self.optimizer_key,
+            learning_rate=self.args.learning_rate,
+            learning_rate_decay=self.args.learning_rate_decay,
+            learning_rate_decay_steps=self.args.learning_rate_decay_steps,
```
lastmansleeping, line 70: @Ullimague Make sure you tweak these when you train your models.

python/ml4ir/applications/classification/config/parse_args.py
Ullimague: At the beginning you mentioned MRR and ACR wouldn't work for non-ranking (as it needs a rank info in the dataset). Currently we can live without it.

python/ml4ir/applications/classification/model/losses/categorical_cross_entropy.py
Ullimague: Good point, at some point we will want to make it parametrizable, not this highest priority for this pr.

python/ml4ir/applications/classification/pipeline.py
Ullimague: Good catch, it's already set as default actually, we should use self.tfrecord_type here.

python/ml4ir/applications/classification/pipeline.py
Ullimague: Agree, I propose to do a follow-up pr for optimizations.

python/ml4ir/applications/classification/pipeline.py
Ullimague: Agree, I propose to do a follow-up pr for optimizations.

python/ml4ir/applications/classification/config/parse_args.py
lastmansleeping: Okay. Yes, there is a bug with the shapes and need for `rank` to use MRR and ACR. I'm asking because we can prioritize fixing it if you need these metrics. Looks like it's not urgent. Sounds good.

python/ml4ir/applications/classification/pipeline.py
```diff
@@ -0,0 +1,131 @@
...
+    def get_relevance_dataset(self, preprocessing_keys_to_fns={}) -> RelevanceDataset:
+        """
+        Creates RelevanceDataset
+
+        NOTE: Override this method to create custom dataset objects
+        """
+
+        # Adding one_hot_vectorizer needed for classification.
+        preprocessing_keys_to_fns = {
+            "one_hot_vectorize_label": get_one_hot_vectorizer(self.feature_config.get_label(), self.file_io)
```
lastmansleeping, line 98: @Ullimague You shouldn't be doing this. Add it to `PreprocessingMap` instead and use it from feature config YAML.

lastmansleeping: If you want to pass in `file_io`, then you can change the function definition of the preprocessing functions to take in `file_io` or `**kwargs`

python/ml4ir/applications/ranking/tests/test_feature_layer.py
```diff
@@ -44,6 +44,97 @@ def test_bytes_sequence_to_encoding_bilstm(self):
         assert sequence_encoding.shape[1] == 1
         assert sequence_encoding.shape[2] == encoding_size
 
+    def test_categorical_embedding_to_encoding_bilstm(self):
+        """
+        Asserts the conversion of a string tensor to its corresponding sequence encoding
+        obtained through the bytes_sequence_to_encoding_bilstm function
```
lastmansleeping, line 50: Wrong function in the comment @Ullimague

python/ml4ir/applications/ranking/tests/test_feature_layer.py
```diff
@@ -44,6 +44,97 @@ def test_bytes_sequence_to_encoding_bilstm(self):
...
+        # Strings 0 and 2 should result in the same embedding because they are the same
+        assert tf.reduce_all(tf.equal(sequence_encoding[0], sequence_encoding[2]))
+        assert not tf.reduce_all(tf.equal(sequence_encoding[0], sequence_encoding[1]))
+        assert not tf.reduce_all(tf.equal(sequence_encoding[3], sequence_encoding[4]))
+        assert not tf.reduce_all(tf.equal(sequence_encoding[1], sequence_encoding[4]))
+
+    def test_categorical_embedding_to_encoding_bilstm_file_truncation(self):
+        """
+        Asserts the conversion of a string tensor to its corresponding sequence encoding
+        obtained through the bytes_sequence_to_encoding_bilstm function
```
lastmansleeping, line 95: @Ullimague Same here.

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -369,3 +300,117 @@ def categorical_indicator_with_vocabulary_file(feature_tensor, feature_info, fil
     categorical_one_hot = tf.expand_dims(categorical_one_hot, axis=1)
 
     return categorical_one_hot
+
+
+def get_categorical_indices_from_vocabulary(feature_info, feature_tensor, file_io):
```
lastmansleeping, line 305: @Ullimague Can we rename this to `get_categorical_indices_from_vocabulary_file`?

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -119,60 +122,51 @@ def categorical_embedding_with_indices(feature_tensor, feature_info, file_io: Fi
...
-                "vocabulary_ids": self.vocabulary_ids,
-                "vocabulary_size": self.vocabulary_size,
-                "num_oov_buckets": self.num_oov_buckets,
-                "feature_name": self.feature_name,
-            }
-        )
-        return config
+    categorical_embeddings = tf.squeeze(categorical_embeddings, axis=1)
+    encoding = get_bilstm_encoding(categorical_embeddings, int(args["encoding_size"] / 2))
+    return encoding
```
lastmansleeping, line 169: @Ullimague Can we move this to a separate function as well under `sequence.py` perhaps? The reason is I want it to be reused for `bytes_sequence_to_encoding_bilstm` instead of defining the same in two places. I like the refactoring you've done until now. Good place for ml4ir.

python/ml4ir/base/model/relevance_model.py
```diff
@@ -331,7 +331,7 @@ def evaluate(
         Override this method to implement your own evaluation metrics.
         """
         if self.is_compiled:
-            self.model.evaluate(test_dataset)
+            return self.model.evaluate(test_dataset)
```
lastmansleeping, line 334: @Ullimague Thanks!

python/ml4ir/applications/classification/pipeline.py
Ullimague: Implemented, this makes only sense when a label is a categorical feature, hence only for classification thought..

python/ml4ir/base/features/preprocessing.py
```diff
@@ -53,6 +58,71 @@ def preprocess_text(feature_tensor, remove_punctuation=False, to_lower=False):
...
+        feature_info: Dictionary representing the configuration parameters for the specific feature from the FeatureConfig.
+                      See categorical_indicator_with_vocabulary_file, here it is used to read a vocabulary file to
+                      create the one hot encoding.
+        file_io: FileIO required to load the vocabulary file.
+
+    Returns:
+        processed float tensor
+    """
+    # Avoid initialization if there is no vocabulary file stated in feature config wrt to label definition.
+    if not ("args" in feature_info.get("feature_layer_info")) or not (
```
lastmansleeping, line 75: @Ullimague Why are you using `feature_layer_info` here? Try using `preprocessing_info` and pass args using that.

python/ml4ir/base/features/preprocessing.py
```diff
@@ -53,6 +58,71 @@ def preprocess_text(feature_tensor, remove_punctuation=False, to_lower=False):
...
+            "vocabulary_file" in feature_info.get("feature_layer_info")["args"]):
+        return
+
+    label_str = tf.keras.Input(shape=(1,), dtype=tf.string)
+    label_one_hot = categorical_indicator_with_vocabulary_file(label_str, feature_info, file_io)
+    one_hot_vectorizer = tf.keras.Model(inputs=label_str, outputs=label_one_hot)
```
lastmansleeping, line 81: @Ullimague Is it necessary to wrap it as a Keras model? I had to do this in the notebook as a quick workaround. Did you try directly calling the function itself? There might be issues with saving the model(I think). But try it out.

lastmansleeping:
@Ullimague Can you also add a test for serving similar to ranking? I think that is important to make sure that the trained model can be shipped to production without issues.

python/ml4ir/base/features/preprocessing.py
lastmansleeping: Otherwise you will also be accidentally running this in the feature layer step.

Ullimague: Because if there is no vocabulary file stated in `feature_layer_info`, we won't be able to initialize the `get_one_hot_label_vectorizer` to `PreprocessingMap`. The vocabulary file was passed via the `feature_layer_info`, and not the `preprocessing_info`.

Ullimague: For instance a config could have a preprocessing_info, but no feature_layer_info. I imagine we still want to authorize this case? @lastmansleeping

lastmansleeping:
> @lastmansleeping added a test for serving, currently the RelevanceModel.save() only supports SequenceExample, and we are facing this issue while doing get_tfrecord_signature for inference. [tensorflow/tensorflow#31686](https://github.com/tensorflow/tensorflow/issues/31686)
> Comment tests with info. We will need to create a dedicated pr for model evaluation and signature inference.

I had already added a workaround for this, right? Use a vocablookup to convert vocabulary to indices and then use embedding layer or categorical_column_with_identity with embedding feature_column. Check `categorical_embedding_with_vocabulary_file` -> I had added documentation of issues as well.

python/ml4ir/applications/classification/tests/test_classification_serving.py
```diff
@@ -0,0 +1,119 @@
...
+        feature_config: FeatureConfig = FeatureConfig.get_instance(
+            tfrecord_type=self.args.tfrecord_type,
+            feature_config_dict=self.feature_config,
+            logger=self.logger,
+        )
+
+        return feature_config
+
+####################################################################################################
+# Test below disabled as we currently face: https://github.com/tensorflow/tensorflow/issues/31686. #
```
lastmansleeping, line 91: @Ullimague Can you please paste your stacktrace? The issue you linked has nothing to do with this scenario. Where are you using categorical_column_with_vocabulary_list?

python/ml4ir/applications/classification/pipeline.py
lastmansleeping: @Ullimague and I discussed offline. Adding this to PreprocessingMap causes issues downstream while saving the model. So we decided to keep this as on-demand that way we can make sure all functions in PreprocessingMap are save-able. @Ullimague will be reverting this to its original implementation and using it from `pipeline.py`. But make sure you file an issue for fixing `one_hot_vectorizer` so that it does not need to be wrapped as a Keras model. That's the real issue here.

PR #52, opened 2020-07-31
Adding categorical embedding with dropout to OOV token @rev calliebradley@

lastmansleeping:
Also fixed vocabulary file and updated tests accordingly to use `domain..` instead of `group..`


YetAnotherUselessJake:
> @Ullimague have you been looped into this? I'm hoping we're using similar techniques for OOV-handling in Ranking + EP?
> 
> @lastmansleeping I haven't read in detail yet, but did you end up having to do your custom dropout thing, or did you just pool together low-traffic orgs in the training set like @Ullimague does?

(arg, wrong account, deleting and reposting :P )

python/ml4ir/base/features/feature_config.py
```diff
@@ -283,6 +283,38 @@ def create_dummy_protobuf(self, num_records=1, required_only=False):
         """
         raise NotImplementedError
 
+    def get_wandb_config(self):
```
Ullimague, line 286: Would probably be better to call it get_weight_bias_config() or create_hyper_parameters().
Reading the function on its own I didn't get what wandb meant.

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -215,6 +220,83 @@ def categorical_embedding_with_vocabulary_file(feature_tensor, feature_info, fil
...
+    where the key is mapped to one id thereby resulting in a
+    many-to-one vocabulary mapping.
+    If id field is absent, a unique natural number id is assigned by default
+    resulting in a one-to-one mapping
+
+    OOV index will be set to 0
+    num_oov_buckets will be 0
+    """
+    feature_layer_info = feature_info.get("feature_layer_info")
+    vocabulary_df = file_io.read_df(feature_layer_info["args"]["vocabulary_file"])
```
Ullimague, line 255: Why can't we just call ml4ir/base/features/feature_fns/categorical.py#get_vocabulary_info() or adapt it if needed?

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -215,6 +220,83 @@ def categorical_embedding_with_vocabulary_file(feature_tensor, feature_info, fil
...
+    vocabulary_size = len(set(vocabulary_keys)) + 1
+    lookup_table = VocabLookup(
+        vocabulary_keys=vocabulary_keys,
+        vocabulary_ids=vocabulary_ids,
+        default_value=0,  # Setting default value to 0 as we will use dropout to mask keys to this OOV value
+        feature_name=feature_info.get("node_name", feature_info["name"]),
+    )
+    feature_tensor_indices = lookup_table(feature_tensor)
+
+    # Mask indices to OOV key of 0 at dropout_rate
```
lastmansleeping, line 277: @jakemannix Check out this. This "drops" org IDs at a `dropout_rate` to the index 0, which I have explicitly assigned to the OOV token. We can get creative on top of this but I just wanted to start off with something.

python/ml4ir/applications/ranking/tests/test_feature_layer.py
```diff
@@ -282,10 +282,81 @@ def test_categorical_embedding_with_vocabulary_file(self):
...
+                string_tensor, feature_info, self.file_io
+            )
+        except ValueError:
+            # Should throw error as method does not work with IDs containing 0
+            assert True
+
+        ########################################################
+        # Test for vocabulary file with no ids mapping specified
+        ########################################################
+        feature_info["feature_layer_info"]["args"][
```
lastmansleeping, line 328: @jakemannix Check out this unittest for a better view of what is happening at training vs inference time.

lastmansleeping: I've set the `dropout_rate = 0.999` to a really high value to test that every org gets mapped to OOV token at training time.

python/ml4ir/applications/ranking/tests/test_feature_layer.py
```diff
@@ -282,10 +282,81 @@ def test_categorical_embedding_with_vocabulary_file(self):
...
+        assert not tf.reduce_all(tf.equal(categorical_embedding[0], categorical_embedding[1]))
+        assert tf.reduce_all(tf.equal(categorical_embedding[4], categorical_embedding[5]))
+
+        # Strings domain_0 and domain_2 should NOT result in the same embedding because they use a default one-to-one vocabulary mapping
+        assert not tf.reduce_all(tf.equal(categorical_embedding[0], categorical_embedding[3]))
+        assert not tf.reduce_all(tf.equal(categorical_embedding[3], categorical_embedding[4]))
+
+        categorical_embedding = model(string_tensor, training=True)
+
+        # Since dropout_rate is set to 0.999, all categorical indices
```
lastmansleeping, line 356: @jakemannix I've set the `dropout_rate = 0.999` to a really high value to test that every org gets mapped to OOV token at training time.

lastmansleeping:
@jakemannix @Ullimague Thanks for jumping onto this. I'm going to be making small updates as I need for training. So I will be tagging this as WIP. But feel free to continue reviewing.

python/ml4ir/base/features/feature_fns/categorical.py
lastmansleeping: Yeah. I will do that. I'm focusing on training models first. Will address this next week if that is okay with you.

python/ml4ir/base/features/feature_config.py
lastmansleeping: Sure. I just went with `wandb` because that's their package name. But makes sense to do that for more readability.

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -220,6 +220,33 @@ def categorical_embedding_with_vocabulary_file(feature_tensor, feature_info, fil
...
+        """
+        Args:
+            dropout_rate: fraction of units to drop, i.e., set to OOV token 0
+            seed: random seed for sampling
+        """
+        super(CategoricalDropout, self).__init__(**kwargs)
+        self.dropout_rate = dropout_rate
+        self.seed = seed
+
+    def call(self, inputs, training=None):
```
lastmansleeping, line 236: No unnecessary scaling and de-scaling anymore. Just a simple mask. Can even rename to `CategoricalMask` to avoid confusion in the future.

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -215,6 +220,107 @@ def categorical_embedding_with_vocabulary_file(feature_tensor, feature_info, fil
...
 
 
+class CategoricalDropout(layers.Layer):
+    """Custom Dropout class for categorical indices"""
+
+    def __init__(self, dropout_rate, seed=None, **kwargs):
+        """
+        Args:
+            dropout_rate: fraction of units to drop, i.e., set to OOV token 0
+            seed: random seed for sampling
```
Ullimague, line 230: It's not about sampling here, is it? For sampling what?

python/ml4ir/base/features/feature_fns/categorical.py
Ullimague: What the latest on this?

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -215,6 +220,107 @@ def categorical_embedding_with_vocabulary_file(feature_tensor, feature_info, fil
...
+        self.dropout_rate = dropout_rate
+        self.seed = seed
+
+    def get_config(self):
+        config = super(CategoricalDropout, self).get_config()
+        config.update({"dropout_rate": self.dropout_rate, "seed": self.seed})
+        return config
+
+    def call(self, inputs, training=None):
+        # At training time, mask indices to 0 at dropout_rate
```
Ullimague, line 242: This isn't a trivial operation to understand why we need it.
Could you add a python doc about why we need to do it, and an example of input output?

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -360,11 +477,26 @@ def build(self, input_shape):
...
             value_dtype=tf.int64,
         )
-        self.lookup_table = lookup.StaticVocabularyTable(
-            initializer=table_init,
-            num_oov_buckets=self.num_oov_buckets,
-            name="{}_lookup_table".format(self.feature_name),
-        )
+
+        """
+        NOTE:
```
Ullimague, line 482: +1
Could we add:
1. A global doc above NOTE (about what build does)
2. Why do we choose to build a StaticVocabularyTable or StaticHashTable depending on the num_oov_buckets?

python/ml4ir/base/features/preprocessing.py
```diff
@@ -109,15 +110,22 @@ def split_and_pad_string(feature_tensor, split_char=",", max_length=20):
...
+        target_height=1,
+        target_width=max_length,
+    )
     padded_tokens = tf.squeeze(padded_tokens, axis=-1)
     return padded_tokens
 
 
+@tf.function
+def signed_log(feature_tensor, shift=1.0):
+    """Signed log"""
```
Ullimague, line 126: We need to add extensive doc.
Mostly about why and what are the advantage, + typical usage for signed log in our use-case.

python/ml4ir/base/features/preprocessing.py
```diff
@@ -109,15 +110,22 @@ def split_and_pad_string(feature_tensor, split_char=",", max_length=20):
...
+    return tf.math.log(tf.add(feature_tensor, tf.cast(tf.constant(shift), tf.float32)))
```
Ullimague, line 127: Don't we need to take the absolute value and check the sign to be in line with the definition?
https://rdrr.io/github/MyKo101/mutils/man/signed_log.html

python/ml4ir/base/model/relevance_model.py
```diff
@@ -207,6 +208,7 @@ def fit(
         monitor_metric: str = "",
         monitor_mode: str = "",
         patience=2,
+        use_wandb_tracking=False,
```
Ullimague, line 211: Need to update the Args doc accordingly.

Ullimague: (and precise what we mean by "tracking")

python/ml4ir/base/model/relevance_model.py
```diff
@@ -516,6 +520,10 @@ def on_predict_end(self, logs):
 
         callbacks_list.append(DebuggingCallback())
 
+        # Weights and Biases Callback
```
Ullimague, line 523: "Add callback to log Weights and Biases"?
(let's make is a phrase instead of a title :))

python/ml4ir/base/pipeline.py
```diff
@@ -167,11 +182,41 @@ def validate_args(self):
 
         return self
 
+    def setup_wandb_config(self):
```
Ullimague, line 185: Could we add a small pydoc?

python/ml4ir/base/features/feature_config.py
lastmansleeping: Renamed to `get_hyperparameter_dict` to be more generic as we might move away from wandb.

python/ml4ir/base/features/feature_fns/categorical.py
lastmansleeping: Fixed.

python/ml4ir/applications/ranking/tests/test_feature_layer.py
lastmansleeping: No. It just doesn't take a value of 1.0 unfortunately. :/

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -57,6 +58,7 @@ def evaluate(
         group_metrics_min_queries: int = 50,
         logs_dir: Optional[str] = None,
         logging_frequency: int = 25,
+        use_wandb_tracking: bool = False,
```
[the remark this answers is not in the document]
lastmansleeping: Done.

python/ml4ir/applications/ranking/tests/test_feature_layer.py
```diff
@@ -282,10 +282,80 @@ def test_categorical_embedding_with_vocabulary_file(self):
...
+        try:
+            categorical_embedding = categorical_fns.categorical_embedding_with_vocabulary_file_and_dropout(
+                string_tensor, feature_info, self.file_io
+            )
+        except ValueError:
+            # Should throw error as method does not work with IDs containing 0
+            assert True
+
+        ########################################################
+        # Test for vocabulary file with no ids mapping specified
```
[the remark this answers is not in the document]
lastmansleeping: But it's still the same code path, right?

python/ml4ir/base/config/parse_args.py
```diff
@@ -151,6 +151,17 @@ def define_args(self):
             help="Run ID for the current training. Autogenerated if not specified.",
         )
 
+        self.add_argument(
+            "--run_group",
+            type=str,
+            default="general",
+            help="Run group defining a meta grouping for the current training run.",
```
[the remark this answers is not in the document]
lastmansleeping: I'll elaborate it. It is just a name to identify a group of model training runs. `run_id` is the unique identifier for each model training run. `run_group` is one level above. You can later use it to group similar runs and select best model.

python/ml4ir/base/config/parse_args.py
```diff
@@ -151,6 +151,17 @@ def define_args(self):
...
+        )
+
+        self.add_argument(
+            "--run_notes", type=str, default="", help="Notes for the current training run.",
```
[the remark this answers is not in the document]
lastmansleeping: They are not really used in any way. This is just additional model training metadata. Say you want to add some description about the models you are training and later want to come back and check on it - this is the place to do it.

python/ml4ir/base/config/parse_args.py
```diff
@@ -281,6 +292,13 @@ def define_args(self):
             help="String specifying the file handler to be used. Should be one of FileHandler keys in ml4ir/base/config/keys.py",
         )
 
+        self.add_argument(
+            "--use_wandb_tracking",
+            type=bool,
+            default=False,
+            help="Whether to track model performance and hyperparameters using Weights and Biases.",
```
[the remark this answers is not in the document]
lastmansleeping: I don't understand your question here. This tracks hyperparameters, architecture config, train, val, test metrics, etc. I can rename it to `use_experiment_tracking` as a more generic name.

python/ml4ir/base/data/tfrecord_reader.py
```diff
@@ -206,7 +206,7 @@ def crop_fn():
                     )
 
                 mask = tf.cond(
-                    tf.shape(mask)[1] < max_sequence_size,
+                    tf.shape(mask)[1] <= max_sequence_size,
```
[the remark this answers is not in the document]
lastmansleeping: This was a bug from earlier. There is no test. Do you have an idea for how we would test this?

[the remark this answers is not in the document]
lastmansleeping: This is used to check that maximum number of records per query condition is met. Earlier it was set to `<` which is not the right meaning of it.

python/ml4ir/base/features/feature_config.py
```diff
@@ -283,6 +283,38 @@ def create_dummy_protobuf(self, num_records=1, required_only=False):
...
+        """Create hyperparameter configs to track in weights and biases"""
```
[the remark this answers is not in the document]
lastmansleeping: This is experiment tracking - hyperparameters, architecture, train, val, test metrics, etc. Think about how you would use ml4ir to train multiple models and then select the best model.

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -215,6 +220,107 @@ def categorical_embedding_with_vocabulary_file(feature_tensor, feature_info, fil
     return embedding
 
 
+class CategoricalDropout(layers.Layer):
+    """Custom Dropout class for categorical indices"""
+
+    def __init__(self, dropout_rate, seed=None, **kwargs):
+        """
+        Args:
+            dropout_rate: fraction of units to drop, i.e., set to OOV token 0
```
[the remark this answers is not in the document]
lastmansleeping: We define OOV index to be 0 for this function and when dropout is applied, it converts p% of the values to 0(which is the OOV index). This allows us to train a more average embedding for the OOV token as opposed to only using low traffic labels.

python/ml4ir/base/features/feature_fns/categorical.py
lastmansleeping: Sampling to select which values to mask/drop

python/ml4ir/base/features/feature_fns/categorical.py
lastmansleeping: It works similar to this without the scaling part -> https://www.tensorflow.org/api_docs/python/tf/keras/layers/Dropout

The example is in the test, right?

lastmansleeping: I've added an example in the comments now.

python/ml4ir/base/features/feature_fns/categorical.py
lastmansleeping: Because StaticHashTable doesn't have num_oov_buckets argument and StaticVocabularyTable does not have default_value argument. For most cases, StaticVocabularyTable is sufficient. But when we want to use a custom default_value(like in the case of the dropout function), we need to use StaticHashTable.

lastmansleeping: Added the doc above note regarding what build does.

python/ml4ir/base/features/preprocessing.py
lastmansleeping: I added the documentation. I don't think I meant to use signed_log. Reverting the name to unsigned_log

python/ml4ir/base/features/preprocessing.py
lastmansleeping: Changed to natural_log

python/ml4ir/base/model/relevance_model.py
lastmansleeping: I've updated this to mean experiment tracking everywhere.

python/ml4ir/base/model/relevance_model.py
lastmansleeping: Done.

python/ml4ir/base/pipeline.py
lastmansleeping: Added.

python/ml4ir/applications/ranking/tests/test_feature_layer.py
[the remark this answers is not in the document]
Ullimague: Yes, but it helps to not test everything in a single test.
- So that when a test fail, we know why (the reason is limited), and can segment more easily the issue.
- Here, we won't know if a change breaks everything after the first failure.
- It help from a CI perspective to easily track the amount of failures.
- And it's good practice :)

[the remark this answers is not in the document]
Ullimague: Here the test could fail for 10 different asserts.

python/ml4ir/applications/ranking/tests/test_feature_layer.py
```diff
@@ -282,10 +282,80 @@ def test_categorical_embedding_with_vocabulary_file(self):
...
+        string_tensor = tf.constant(
+            ["domain_0", "domain_1", "domain_0", "domain_2", "domain_10", "domain_11"]
+        )
+        try:
+            categorical_embedding = categorical_fns.categorical_embedding_with_vocabulary_file_and_dropout(
+                string_tensor, feature_info, self.file_io
+            )
+        except ValueError:
+            # Should throw error as method does not work with IDs containing 0
+            assert True
```
Ullimague, line 323: We should do:

```
exception_raised = False
try:
....
except ValueError:
     exception_raised = True

assert exception_raised
```

Ullimague: Because here, if the exception isn't raised, the test will still pass.

python/ml4ir/base/data/tfrecord_reader.py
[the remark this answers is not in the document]
Ullimague: We should add dedicated test about how and whether the mask is applied correctly.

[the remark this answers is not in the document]
Ullimague: How did you assess the masking before? Manually?

python/ml4ir/base/features/feature_fns/categorical.py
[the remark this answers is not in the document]
Ullimague: We should exactly add what you said as part of the doc here :)

python/ml4ir/base/features/feature_fns/categorical.py
Ullimague: We should add this context :)

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -215,6 +220,105 @@ def categorical_embedding_with_vocabulary_file(feature_tensor, feature_info, fil
...
+        inputs: [[1, 2, 3], [4, 1, 2]]
+        dropout_rate = 0.5
+
+        When training, output: [[0, 0, 3], [0, 1, 2]]
+        When testing, output: [[1, 2, 3], [4, 1, 2]]
+        """
+        if training:
+            return tf.math.multiply(
+                tf.cast(
+                    tf.random.uniform(shape=tf.shape(inputs)) >= self.dropout_rate, dtype=tf.int64
```
Ullimague, line 258: We don't need to use a seed here?

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -215,6 +220,105 @@ def categorical_embedding_with_vocabulary_file(feature_tensor, feature_info, fil
...
+
+        This works similar to tf.keras.layers.Dropout without the scaling
+        Ref: https://www.tensorflow.org/api_docs/python/tf/keras/layers/Dropout
+
+        Example:
+        inputs: [[1, 2, 3], [4, 1, 2]]
+        dropout_rate = 0.5
+
+        When training, output: [[0, 0, 3], [0, 1, 2]]
+        When testing, output: [[1, 2, 3], [4, 1, 2]]
```
Ullimague, line 253: 👍 👍

Ullimague: Could we add a unit-test exactly for this?
with dropout_rate=1, 0.5 and 0?

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -317,11 +423,20 @@ def categorical_indices_from_vocabulary_file(feature_info, feature_tensor, file_
         vocabulary_keys: values of the vocabulary stated in the vocabulary_file.
     """
     vocabulary_keys, vocabulary_ids = get_vocabulary_info(feature_info, file_io)
-    num_oov_buckets = feature_info.get("feature_layer_info")["args"].get("num_oov_buckets", 1)
+    args = feature_info.get("feature_layer_info")["args"]
+
+    if "dropout_rate" in args:
```
Ullimague, line 428: Do we have test that goes through both path?

python/ml4ir/base/features/feature_fns/categorical.py
Ullimague: We should also add a form of what you just said for the context as doc:
_Because StaticHashTable doesn't have num_oov_buckets argument and StaticVocabularyTable does not have default_value argument. For most cases, StaticVocabularyTable is sufficient. But when we want to use a custom default_value(like in the case of the dropout function), we need to use StaticHashTable._
Mostly adding the _why_ :)

Ullimague: ie that we need it for instance to add a custom default_value, like in the case of the dropout function.

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -406,11 +549,25 @@ def get_vocabulary_info(feature_info, file_io):
...
-        vocabulary_df["id"].values if "id" in vocabulary_df else list(range(len(vocabulary_keys)))
-    )
-    return vocabulary_keys, vocabulary_ids
\ No newline at end of file
+    if "dropout_rate" in args:
+        # NOTE: If a dropout_rate is specified, then reserve 0 as the OOV index
+        vocabulary_ids = (
+            vocabulary_df["id"].values
+            if "id" in vocabulary_df
+            else list(range(1, len(vocabulary_keys) + 1))
```
Ullimague, line 561: We should have:
```
OOV = 1
list(range(1, len(vocabulary_keys) + OOV))
```

python/ml4ir/base/features/preprocessing.py
Ullimague: why did you change to natural_log? Did it change something?

python/ml4ir/base/data/tfrecord_reader.py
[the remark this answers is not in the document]
Ullimague: The question to have in mind would be basically: _as a new dev to ml4ir, if I revert this change tomorrow will there be something preventing me to do so, and could I get the reason easily?_

python/ml4ir/base/config/parse_args.py
```diff
@@ -151,6 +151,20 @@ def define_args(self):
             help="Run ID for the current training. Autogenerated if not specified.",
         )
 
+        self.add_argument(
+            "--run_group",
+            type=str,
+            default="general",
+            help="Run group defining a meta grouping to filter different model training runs for best model selection as a post step.",
```
Ullimague, line 158: _Ullimague: I'm not sure I understand the help without extra context.
Could we namely state: why should we set it, what are the options, where are they stated, what is the outcome of it?_

_lastmansleeping: I'll elaborate it. It is just a name to identify a group of model training runs. run_id is the unique identifier for each model training run. run_group is one level above. You can later use it to group similar runs and select best model._

Ullimague: We should say so, maybe in a less abstract way.
Could we just add:
- it is an identifier used to group the different run_id.
- can later use it to group similar runs and select best model.
- where is it output? (basically where can we access it back)?

python/ml4ir/base/config/parse_args.py
```diff
@@ -151,6 +151,20 @@ def define_args(self):
...
+        )
+
+        self.add_argument(
+            "--run_notes",
+            type=str,
+            default="",
+            help="Notes for the current training run. Use this argument to add short description of the model training run that helps in identifying the run later.",
```
Ullimague, line 165: +1, could we also say: where do the note end up? And how can the user get back to it?

python/ml4ir/base/config/parse_args.py
```diff
@@ -281,6 +295,13 @@ def define_args(self):
             help="String specifying the file handler to be used. Should be one of FileHandler keys in ml4ir/base/config/keys.py",
         )
 
+        self.add_argument(
+            "--track_experiment",
+            type=bool,
+            default=False,
+            help="Whether to track model performance and hyperparameters.",
```
Ullimague, line 302: Can we say where the user can access to the stats?

python/ml4ir/base/features/feature_fns/categorical.py
lastmansleeping: I can't test dropout because it doesn't promise 50% will be dropped. It is random. So the only way I could test it was set the rate to 0.999

python/ml4ir/base/features/feature_fns/categorical.py
lastmansleeping: We have unit tests for `categorical_embedding_from_vocabulary_file` and `categorical_embedding_from_vocabulary_file_and_dropout`. I can add something for this.

python/ml4ir/base/config/parse_args.py
lastmansleeping: I can add that. You can access it in the experiment tracking data.

python/ml4ir/base/config/parse_args.py
lastmansleeping: It is just going to be added to the experiment tracking data. Currently it is in the weights and biases summary file. But we might choose to write a JSON of our own later.

python/ml4ir/base/config/parse_args.py
lastmansleeping: I left this out on purpose. Currently we use weights and biases, but we might move out of this and just use a simple JSON file.

Ullimague: So in the help we need to say what we currently do, and update it when we change it :)

python/ml4ir/base/config/parse_args.py
Ullimague: So in the help we need to say what we currently do, and update it when we change it :)

python/ml4ir/applications/ranking/tests/test_feature_layer.py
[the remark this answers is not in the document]
lastmansleeping: I will break it up.

python/ml4ir/base/data/tfrecord_reader.py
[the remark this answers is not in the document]
lastmansleeping: I will take a shot at this. But in our current design, it is painfully complicated to test this.

python/ml4ir/base/features/feature_fns/categorical.py
lastmansleeping: Had missed it. Added now.

lastmansleeping:
@Ullimague Updated the documentation and split the unit tests. For breaking up the SequenceExample parsing function, I have created a new Issue/User Story -> https://github.com/salesforce/ml4ir/issues/58

@jakemannix Will you help approve this review as Guillaume is out currently? I've worked with Guillaume to resolve most of his comments.

python/ml4ir/applications/ranking/tests/test_feature_layer.py
```diff
@@ -282,27 +299,114 @@ def test_categorical_embedding_with_vocabulary_file(self):
         assert not tf.reduce_all(tf.equal(categorical_embedding[0], categorical_embedding[1]))
         assert tf.reduce_all(tf.equal(categorical_embedding[4], categorical_embedding[5]))
 
-        # Strings group_0 and group_2 should NOT result in the same embedding because they use a default one-to-one vocabulary mapping
+        # Strings domain_0 and domain_2 should NOT result in the same embedding because they use a default one-to-one vocabulary mapping
         assert not tf.reduce_all(tf.equal(categorical_embedding[0], categorical_embedding[3]))
         assert not tf.reduce_all(tf.equal(categorical_embedding[3], categorical_embedding[4]))
 
-    def test_categorical_indicator_with_vocabulary_file(self):
+    def test_categorical_embedding_with_vocabulary_file_with_ids_and_dropout(self):
```
lastmansleeping, line 306: @mohazahran This is a unittest that captures what is happening in the function. The goal is to learn categorical embeddings, but "drop" the categorical tokens at a `dropout_rate` to the OOV token so that the OOV embedding is learned across all categorical values.

python/ml4ir/base/config/parse_args.py
```diff
@@ -148,7 +148,21 @@ def define_args(self):
...
-            help="Run ID for the current training. Autogenerated if not specified.",
+            help="Unique string identifier for the current training run. Used to identify logs and models directories. Autogenerated if not specified.",
+        )
+
+        self.add_argument(
+            "--run_group",
+            type=str,
+            default="general",
+            help="Unique string identifier to group multiple model training runs. Allows for defining a meta grouping to filter different model training runs for best model selection as a post step.",
+        )
```
calliebradley, line 159: [very minor] Would be great to have an example here of what syntax is expected in order to use this arg.

python/ml4ir/base/config/parse_args.py
calliebradley: @lastmansleeping Ya, it would be helpful to add that this note goes in the weights and biases summary file

python/ml4ir/base/config/parse_args.py
calliebradley: @lastmansleeping +1 Very minor but would be helpful to track what we currently do and update when we change it.

python/ml4ir/base/features/preprocessing.py
calliebradley: @lastmansleeping +1 to Guillaume's question, actually is this used anywhere?

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -57,16 +58,19 @@ def evaluate(
...
         Args:
             test_dataset: an instance of tf.data.dataset
             inference_signature: If using a SavedModel for prediction, specify the inference signature
+            additional_features: Additional post processing feature functions as key value pairs
+            group_metrics_min_queries: Minimum number of queries per group to be used for group aggregate metrics
+            logs_dir: Directory to log the predictions and metrics
             logging_frequency: integer representing how often(in batches) to log status
-            metric_group_keys: list of fields to compute group based metrics on
-            save_to_file: set to True to save predictions to file like self.predict()
+            track_experiment: Boolean to determine if weights and biases tracking should be used
```
calliebradley, line 73: @lastmansleeping Should we also take the mentions of `metric_group_keys` and `save_to_file` out of `base/model/relevance_model.py` ? I don't see them implemented anywhere [anymore?].

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -57,16 +58,19 @@ def evaluate(
...
         logging_frequency: int = 25,
+        track_experiment: bool = False,
     ):
         """
         Evaluate the ranking model
 
         Args:
             test_dataset: an instance of tf.data.dataset
             inference_signature: If using a SavedModel for prediction, specify the inference signature
+            additional_features: Additional post processing feature functions as key value pairs
```
calliebradley, line 69: @lastmansleeping It's a little confusing to me that some of these args are accessible through `parse_args` and some are not. Is that intentional/what's the reasoning?

python/ml4ir/base/config/parse_args.py
lastmansleeping: It's just a string actually. Not sure I understand what you mean by example.

python/ml4ir/applications/ranking/model/ranking_model.py
lastmansleeping: Yep. I will do that.

python/ml4ir/applications/ranking/model/ranking_model.py
lastmansleeping: Yeah, this is intentional. These arguments are for using ml4ir as a library with custom functions. So you'll be plugging them in here as a dictionary of functions. More documentation coming soon.

python/ml4ir/base/data/tfrecord_reader.py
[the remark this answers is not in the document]
calliebradley: @lastmansleeping @Ullimague If it's complicated to test this and it gets deprioritized right now, at least is there a document we can point to explaining why the partial inequality is correct? That would help a new dev understand.

python/ml4ir/base/features/feature_config.py
```diff
@@ -283,6 +283,46 @@ def create_dummy_protobuf(self, num_records=1, required_only=False):
...
+    def get_hyperparameter_dict(self):
+        """
+        Create hyperparameter configs to track model metadata for best model selection
+        Unwraps the feature config for each of the features to add
+        preprocessing_info and feature_layer_info as key value pairs
+        that can be tracked across the experiment. This can be used to
+        identify the values that were set for the different feature layers
+        in a given experiment. Will be used during best model selection and
+        Hyper Parameter Optimization.
+        """
```
calliebradley, line 295: @lastmansleeping Great description, very clear! 👍

python/ml4ir/base/features/feature_fns/categorical.py
calliebradley: @lastmansleeping Thoughts on Guillaume's suggestion here? Would we want this 00V index customizable here?

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -2,6 +2,7 @@
 from tensorflow.keras import layers
 from tensorflow import feature_column
 from tensorflow import lookup
+from tensorflow.keras import backend as K
```
calliebradley, line 5: @lastmansleeping I don't see this used, do we need the import?

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -215,6 +220,109 @@ def categorical_embedding_with_vocabulary_file(feature_tensor, feature_info, fil
...
 
+class CategoricalDropout(layers.Layer):
+    """Custom Dropout class for categorical indices"""
+
+    def __init__(self, dropout_rate, seed=None, **kwargs):
+        """
+        Args:
+            dropout_rate: fraction of units to drop, i.e., set to OOV token 0
+            seed: random seed for sampling to mask/drop categorical labels
+
```
calliebradley, line 231: @lastmansleeping Super minor question, but why are you using zeros instead of o's for "00V" references? Is this an industry standard (and if so, curious what the reasoning is)?

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -215,6 +220,109 @@ def categorical_embedding_with_vocabulary_file(feature_tensor, feature_info, fil
...
+        Note:
+        We define OOV index to be 0 for this function and when dropout is applied, it converts p% of the values to 0(which is the OOV index). This allows us to train a good average embedding for the OOV token.
```
calliebradley, line 233: @lastmansleeping [EXTREMELY minor] Would be nice to put some PEP-8 length returns on this so one can see the whole note without scrolling :)

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -215,6 +220,109 @@ def categorical_embedding_with_vocabulary_file(feature_tensor, feature_info, fil
...
+        config = super(CategoricalDropout, self).get_config()
+        config.update({"dropout_rate": self.dropout_rate, "seed": self.seed})
+        return config
+
+    def call(self, inputs, training=None):
+        """
+        At training time, mask indices to 0 at dropout_rate
+
+        This works similar to tf.keras.layers.Dropout without the scaling
+        Ref: https://www.tensorflow.org/api_docs/python/tf/keras/layers/Dropout
```
calliebradley, line 249: Great to see the reference, very helpful!

python/ml4ir/base/features/feature_fns/categorical.py
calliebradley: EDIT Ah never mind, I see that it is o's, was just the font throwing me off! :)

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -317,11 +427,20 @@ def categorical_indices_from_vocabulary_file(feature_info, feature_tensor, file_
         vocabulary_keys: values of the vocabulary stated in the vocabulary_file.
     """
     vocabulary_keys, vocabulary_ids = get_vocabulary_info(feature_info, file_io)
-    num_oov_buckets = feature_info.get("feature_layer_info")["args"].get("num_oov_buckets", 1)
+    args = feature_info.get("feature_layer_info")["args"]
+
+    if "dropout_rate" in args:
+        default_value = 0  # Default OOV index when using dropout
+        num_oov_buckets = None
```
calliebradley, line 434: @lastmansleeping It concerns me to see a default value defined here (and for `num_oov_buckets`) instead of in `parse_args`, but is the reasoning because we want to customize it depending on whether we're using a dropout rate? I guess I'll be easy to convince if you can explain a little bit ;)

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -345,26 +465,55 @@ class VocabLookup(layers.Layer):
...
-            num_oov_buckets=self.num_oov_buckets,
-            name="{}_lookup_table".format(self.feature_name),
-        )
+
+        """
+        NOTE:
+        If num_oov_buckets are specified, use a StaticVocabularyTable
+        otherwise, use a StaticHashTable with the specified default_value
+
+        For most cases, StaticVocabularyTable is sufficient. But when we want to use a custom default_value(like in the case of the dropout function), we need to use StaticHashTable.
```
calliebradley, line 501: [VERY MINOR] Same comment on using PEP-8 line breaks for the note

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -345,26 +465,55 @@ class VocabLookup(layers.Layer):
...
+                name="{}_lookup_table".format(self.feature_name),
+            )
+        elif self.default_value is not None:
+            self.lookup_table = lookup.StaticHashTable(
+                initializer=table_init,
+                default_value=self.default_value,
+                name="{}_lookup_table".format(self.feature_name),
+            )
+        else:
+            raise KeyError("You must specify either num_oov_buckets or default_value")
```
calliebradley, line 516: @lastmansleeping Technically this error looks impossible to happen to me, since one of them is defined as none in lines 433 and following. Right? But I suppose good to keep this, as a safeguard in case someone changes that code without realizing the waterfall error it would cause.

python/ml4ir/base/features/feature_layer.py
```diff
@@ -8,6 +8,9 @@
 from ml4ir.base.features.feature_fns.categorical import categorical_embedding_with_hash_buckets
 from ml4ir.base.features.feature_fns.categorical import categorical_embedding_with_indices
 from ml4ir.base.features.feature_fns.categorical import categorical_embedding_with_vocabulary_file
+from ml4ir.base.features.feature_fns.categorical import (
+    categorical_embedding_with_vocabulary_file_and_dropout,
+)
 from ml4ir.base.io.file_io import FileIO
```
calliebradley, line 14: @lastmansleeping Sorry so many minor comments, but why import it this way? Do you expect this to become a list in the future?

python/ml4ir/applications/ranking/tests/test_metrics.py
```diff
@@ -11,29 +11,29 @@
 GOLD_METRICS = {
     "query_count": 1408.0,
```
calliebradley, line 12: @lastmansleeping So to be clear, did these and the numbers in `test_ranking_model.py` change because of the categorical embeddings addition?

python/ml4ir/base/config/parse_args.py
calliebradley: @lastmansleeping Oh I understand. It sounded like it did the grouping somehow. Ok, then it's ok as it is.

python/ml4ir/applications/ranking/model/ranking_model.py
calliebradley: Ok. Good enough for me.

calliebradley:
Ok, I've gone through this and made some minor comments, also checked to see what we still need to address from Guillaume. After my comments are addressed we should be ok to merge. I'm not listed as a reviewer so @lastmansleeping I think you'd need to add me before I can approve.

calliebradley (approved):
Looks good. I commented on a few things to check we've finished addressing. A few comments are marked as minor, those are optional to address.

python/ml4ir/base/features/feature_fns/categorical.py
lastmansleeping: We need a numeric representation of OOV index. So any text that is not in our vocabulary will get mapped to 0th index.

python/ml4ir/base/features/feature_fns/categorical.py
lastmansleeping: Yes. You are right. You choose which function you want to use from the feature config as opposed to the command line arguments.

python/ml4ir/base/features/feature_layer.py
lastmansleeping: Oh nice. I didn't do this. I think the PEP-8 plugin we have on the repo auto-formated it. Pretty cool really.

python/ml4ir/applications/ranking/tests/test_metrics.py
lastmansleeping: Not really. I think it changed slightly because I fixed a bug in the data loading code.

PR #53, opened 2020-08-05
Script for converting ranklib format to ml4ir

mohazahran:


lastmansleeping:
@mohazahran Can you modify this script to be usable as a utility in ml4ir? Check `csv_reader.py` for example.

Additionally, might be useful to add a unit test for data load for this format.

python/ml4ir/base/data/ranklib_to_ml4ir.py
```diff
@@ -0,0 +1,42 @@
...
+    feature_values[0] = relevace_name+':'+feature_values[0]
+    r={}
+    for fv in feature_values:
+        feat = fv.split(':')[0].strip()
+        val = fv.split(':')[1].strip()
+        if feat == relevace_name or feat == query_id_name:
+            r[feat] = val
+        else:
+            r['f_'+feat] = val
+    rows.append(r)
```
jakemannix, line 30: This should be a function, so it's reusable.  Ideally something which takes in an Iterator[str] and perhaps spits out an Iterator[Dict[str, str]].

python/ml4ir/base/data/ranklib_to_ml4ir.py
```diff
@@ -0,0 +1,42 @@
...
+        #r = {'f_'+fv.split(':')[0].strip() : fv.split(':')[1].strip() for fv in feature_values}
+
+f.close()
+df = pd.DataFrame(rows)
+if args.GL2Clicks == 1:
+    groups = df.groupby(query_id_name)
+    for gname, group in groups:
+        df.loc[df[query_id_name] == gname, relevace_name] = (df.loc[df[query_id_name] == gname].relevance == max(group.relevance)).astype(int)
+df.to_csv(args.outputfilePath)
+print('DONE!')
```
jakemannix, line 40: should probably wire in the same Logger which the rest of ml4ir uses

python/ml4ir/base/data/ranklib_to_ml4ir.py
jakemannix: (and in fact, probably that method will call a function which parses each line: str -> Dict[str,str] )

jakemannix:
> @mohazahran Can you modify this script to be usable as a utility in ml4ir? Check csv_reader.py for example.
> 
> Additionally, might be useful to add a unit test for data load for this format.

+1 to both of these :)

python/ml4ir/base/data/ranklib_to_ml4ir.py
```diff
@@ -0,0 +1,42 @@
+import pandas as pd
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--inputfilePath', type=str, default='', help='ranklib input file path')
+parser.add_argument('--outputfilePath', type=str, default='', help='output converted file path')
+parser.add_argument('--keepAdditionalInfo', type=int, default=1, help='Option to keep additional info (All info after the "#") 1 to keep, 0 to ignore')
```
balikasg, line 7: You may want to check [this answer for booleans](https://stackoverflow.com/a/15008806/1881878). Essentially here we want a bool and using an int is a hack!

python/ml4ir/base/data/ranklib_to_ml4ir.py
```diff
@@ -0,0 +1,42 @@
+import pandas as pd
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--inputfilePath', type=str, default='', help='ranklib input file path')
```
balikasg, line 5: Consider renaming this `input-file`, `inputFile`, or `inputFilePath` or input-filepath etc. I mean camel-case or no came-case, a consistent pattern! If you use `input-file` (my preferred solution as it is less verbose) `argparse` will transform it to `parser.input_file` for free. You don't need the default! Same below!

python/ml4ir/base/data/ranklib_to_ml4ir.py
balikasg: Same for `GL2Clicks` below!

python/ml4ir/base/data/ranklib_to_ml4ir.py
balikasg: +1

python/ml4ir/base/data/ranklib_to_ml4ir.py
```diff
@@ -0,0 +1,42 @@
...
+parser.add_argument('--outputfilePath', type=str, default='', help='output converted file path')
+parser.add_argument('--keepAdditionalInfo', type=int, default=1, help='Option to keep additional info (All info after the "#") 1 to keep, 0 to ignore')
+parser.add_argument('--GL2Clicks', type=int, default=1, help='Convert graded relevance to clicks (only max relevant document is considered clicked) 1 to convert')
+args = parser.parse_args()
+query_id_name = 'qid'
```
balikasg, line 10: Consider moving these two (`query_id_name` and `relevance_name`) below the imports and upper-casing them; they seem user choices and this would enable other users to easily change them.

python/ml4ir/base/data/ranklib_to_ml4ir.py
lastmansleeping: I prefer [snake case](https://en.wikipedia.org/wiki/Snake_case#:~:text=Snake%20case%20(stylized%20as%20snake_case,%2C%20function%2C%20and%20file%20names.) for all python code.

python/ml4ir/base/data/ranklib_to_ml4ir.py
```diff
@@ -0,0 +1,42 @@
...
+relevace_name = 'relevance'
+
+f = open(args.inputfilePath, 'r')
+cntr = 0
```
balikasg, line 14: Is this used elsewhere?

python/ml4ir/base/data/ranklib_to_ml4ir.py
```diff
@@ -0,0 +1,42 @@
...
+    r={}
+    for fv in feature_values:
+        feat = fv.split(':')[0].strip()
+        val = fv.split(':')[1].strip()
+        if feat == relevace_name or feat == query_id_name:
+            r[feat] = val
+        else:
+            r['f_'+feat] = val
+    rows.append(r)
+        #r = {'f_'+fv.split(':')[0].strip() : fv.split(':')[1].strip() for fv in feature_values}
```
balikasg, line 31: this commented line should be removed!

python/ml4ir/base/data/ranklib_to_ml4ir.py
balikasg: yes! I prefer it too for inline code, for args i prefer "-" which results in snake-case with argparser. Anyway! Consistency is the key!

mohazahran:
I modified the scrip to account for the comments and suggestions. I also added a unit test and a sample ranklib file for testing.

python/ml4ir/applications/ranking/tests/test_ranklib_to_ml4ir.py
```diff
@@ -0,0 +1,40 @@
...
+warnings.filterwarnings("ignore")
+
+INPUT_FILE = "ml4ir/applications/ranking/tests/data/sample.txt"
+OUTPUT_FILE = "ml4ir/applications/ranking/tests/data/sample_ml4ir.csv"
+QUERY_ID_NAME = 'qid'
+RELEVANCE_NAME = 'relevance'
+KEEP_ADDITIONAL_INFO = 1
+GL_2_CLICKS = 1
+
+class Test_ranklib_to_ml4ir(unittest.TestCase):
```
lastmansleeping, line 17: [Minor] @mohazahran People generally use CapWords convention for class names -> https://www.python.org/dev/peps/pep-0008/#id41

python/ml4ir/applications/ranking/tests/test_ranklib_to_ml4ir.py
```diff
@@ -0,0 +1,40 @@
...
+    def setUp(self):
+        ranklib_to_ml4ir.convert(INPUT_FILE, OUTPUT_FILE, KEEP_ADDITIONAL_INFO, GL_2_CLICKS)
```
lastmansleeping, line 19: @mohazahran Might be better to do this within the `test()` method itself.

python/ml4ir/applications/ranking/tests/test_ranklib_to_ml4ir.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def test(self):
```
lastmansleeping, line 21: @mohazahran Can you rename this to something along the lines of `test_ranklib_to_tfrecord`?

python/ml4ir/base/data/ranklib_to_ml4ir.py
```diff
@@ -0,0 +1,60 @@
+import pandas as pd
+import argparse
+import logging
+
+#Constanst for the column names given to the query id and graded relevance.
```
lastmansleeping, line 5: [Minor] @mohazahran type on Constant

python/ml4ir/base/data/ranklib_to_ml4ir.py
```diff
@@ -0,0 +1,60 @@
...
+QUERY_ID_NAME = 'qid'
+RELEVANCE_NAME = 'relevance'
+
+
+def process_line(line, keep_additional_info):
```
lastmansleeping, line 10: @mohazahran Can you add docstrings for the methods explaining the parameters and what is returned? Doesn't have to be too detailed to get us started off.

python/ml4ir/base/data/ranklib_to_ml4ir.py
```diff
@@ -0,0 +1,60 @@
...
+    if keep_additional_info == 1:
+        feature_values = line.replace('#', '').replace(' = ', ':').strip().split()
+    else:
+        feature_values = line.split('#')[0].strip().split()
+    feature_values[0] = RELEVANCE_NAME + ':' + feature_values[0]
+    r = {}
+    for fv in feature_values:
+        feat = fv.split(':')[0].strip()
+        val = fv.split(':')[1].strip()
+        if feat == RELEVANCE_NAME or feat == QUERY_ID_NAME:
```
lastmansleeping, line 20: @mohazahran Does using this hardcoded keys, limit us to only using ranklib? Would this work even if we want to use ml4ir for libsvm format in general? ml4ir could be used for other ML tasks as well.

python/ml4ir/base/data/ranklib_to_ml4ir.py
```diff
@@ -0,0 +1,60 @@
...
+            r[feat] = val
+        else:
+            r['f_' + feat] = val
+    return r
+
+def convert(input_file, output_file, keep_additional_info, gl_2_clicks):
```
lastmansleeping, line 26: @mohazahran Can you create an API similar to this one here? https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/base/data/csv_reader.py#L14:L27

python/ml4ir/base/data/ranklib_to_ml4ir.py
```diff
@@ -0,0 +1,60 @@
...
+    f = open(input_file, 'r')
+    rows = []
+
+    for line in f:
+        rows.append(process_line(line, keep_additional_info))
+    f.close()
+
+    df = pd.DataFrame(rows)
+    if gl_2_clicks == 1:
```
lastmansleeping, line 35: @mohazahran What is `gl_2_clicks`?

lastmansleeping: Might be useful to add a few comments about what is happening here.

python/ml4ir/base/data/ranklib_to_ml4ir.py
```diff
@@ -0,0 +1,60 @@
...
+        groups = df.groupby(QUERY_ID_NAME)
```
lastmansleeping, line 36: @mohazahran Try using this method to convert to TFRecord. I think we should abstract the CSV data format here and jump to TFRecord as part of the `read()` method.

https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/base/data/tfrecord_writer.py#L80

lastmansleeping: This is a good reference -> https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/base/data/csv_reader.py

Basically, write the dataframe to TFRecord and then read a TFRecordDataset

python/ml4ir/base/data/ranklib_to_ml4ir.py
mohazahran: I believe libsvm uses ranklib format as well which specifies the features as "name:value" so it shouldn't be limiting.
However, I noticed some datasets as "Yahoo LETOR" stores only non zero features (for space efficiency). I will handle that as well

python/ml4ir/base/data/ranklib_to_ml4ir.py
mohazahran: it's one of the script's arguments. It's a flag to whether to convert the graded relevance to clicks or not.

python/ml4ir/base/data/ranklib_to_ml4ir.py
mohazahran: Personally, I think converting to csv is more natural to me, plus it makes the script usable in more scenarios.
But, I can write another function that does that.

lastmansleeping: Another function sounds good to me.

python/ml4ir/base/data/ranklib_to_ml4ir.py
lastmansleeping: Right. I just wanted us to be able to use this script to convert data that is not necessarily limited to ranking(ranklib) alone. So I was wondering if using `query_id` and `relevance` was a good idea or not. May be we can have two paths based on an argument to specify if the dataset is using ranklib? A better design would be creating libsvm parser and then ranklib parser which inherits the libsvm parser. That way most of the functionality can be shared.

mohazahran: I'm not sure that I fully understand your point. You see, ranklib and libsvm shares the same format (with minor differences). Both require a query Id and a target relevance present. So explicitly using both should not be limiting the script.

PR #54, opened 2020-08-14
[WIP] Porting to tensorflow2.3.0 and Keras model subclassing API

lastmansleeping:


lastmansleeping:
TODO:
- Fix metrics
- Verify model saving loading end-to-end

python/ml4ir/applications/classification/model/losses/categorical_cross_entropy.py
```diff
@@ -18,18 +18,15 @@ def get_loss(loss_key) -> RelevanceLossBase:
         raise NotImplementedError
 
 
-class CategoricalCrossEntropy(RelevanceLossBase):
-
-    def get_loss_fn(self, **kwargs):
+class CategoricalCrossEntropy(losses.CategoricalCrossentropy, RelevanceLossBase):
```
balikasg, line 21: what is the advantage of defining it here? Consistency with the rest of loss definitions? Wouldn't it work if we imported from keras directly?

lastmansleeping: The way we have defined loss functions in ml4ir, they are coupled with the final activation layer. You can make the case that we shouldn't do this as default and that's a fair argument. But it will need some minor redesign.

PR #56, opened 2020-09-01
Classification model - inferring number of units for last layer

Ullimague:
This PR is to allow not specifying a number of units for the last layer of the model_config.yaml, in which case we infer it from the vocabulary files passed as part of label defined in the feature_config.yaml.
@rev lastmansleeping@


python/ml4ir/base/model/architectures/dnn.py
```diff
@@ -11,12 +17,30 @@ class DNNLayer:
...
+        """
         def get_op(layer_type, layer_args):
             if layer_type == DNNLayer.DENSE:
+                if not "units" in layer_args or layer_args["units"] == -1:
+                    try:
+                        label_feature_info = feature_config.get_label()
+                        vocabulary_keys, vocabulary_ids = get_vocabulary_info(label_feature_info, self.file_io)
+                        layer_args["units"] = len(vocabulary_keys) + OOV
+                    except:
+                        raise KeyError("units can only not be specified for the last layer, in which can a vocabulary"
```
balikasg, line 42: I would suggest a Keras alike message, a bit more actionable
"We were not able to find information for the output layer of your DNN. Try specifying the number of output units either by passing "units" in the model configuration yaml file or units in the feature configuration file."

python/ml4ir/base/model/architectures/dnn.py
```diff
@@ -11,12 +17,30 @@ class DNNLayer:
...
+            label vocabulary defined in the feature_config to deduce the number of units.
+        :return: List[layers]: list of keras layer corresponding to each of the layers defined in the model_config.
+        """
         def get_op(layer_type, layer_args):
             if layer_type == DNNLayer.DENSE:
+                if not "units" in layer_args or layer_args["units"] == -1:
+                    try:
+                        label_feature_info = feature_config.get_label()
+                        vocabulary_keys, vocabulary_ids = get_vocabulary_info(label_feature_info, self.file_io)
+                        layer_args["units"] = len(vocabulary_keys) + OOV
```
balikasg, line 40: This is nice!

lastmansleeping: Although this looks very neat and is handy for the k-class classification problem, I am not a fan of wiring in feature config and vocabulary here as it does not make sense for other ML applications. If I want to train k-class classifier without a vocabulary file, how would I do it? I think the vocabulary logic should be outside of this architecture function.

Ullimague: If I want to train k-class classifier without a vocabulary file, how would I do it? ==> just add the "units" param in the model_config.

Where the confusion may come from is the name "feature config", which isn't actually only "feature config" in the sense of input, but it also contains the labels/output definition.

lastmansleeping: @Ullimague and I synced offline and decided that this is okay for now and we will plan a redesign of feature config and model config in the future.

python/ml4ir/applications/classification/tests/test_model_config.py
```diff
@@ -0,0 +1,106 @@
+import yaml
+from ml4ir.applications.classification.tests.test_base import ClassificationTestBase
+from ml4ir.base.features.feature_config import FeatureConfig
+from ml4ir.base.model.architectures.dnn import DNN
+
+
+class DNNModelTest(ClassificationTestBase):
+
+    def test_dnn_last_layer_units(self):
```
lastmansleeping, line 9: @Ullimague Probably better to call it `num_classes`?

python/ml4ir/applications/classification/tests/test_model_config.py
```diff
@@ -0,0 +1,106 @@
...
+              - type: dropout
+                name: second_dropout
+                rate: 0.0
+              - type: dense
+                name: final_dense
+                activation: null
+        ''')
+
+        dnn = DNN(model_info, feature_config, self.file_io)
+        assert(len(dnn.layer_ops) == 5)
```
lastmansleeping, line 106: @Ullimague We should probably assert to see if each of the layers are there, no?

python/ml4ir/applications/classification/tests/test_model_config.py
```diff
@@ -0,0 +1,106 @@
...
+                name: first_dense
+                units: 256
+                activation: relu
+              - type: dense
+                name: final_dense
+                activation: null
+        ''')
+
+        dnn = DNN(model_info, feature_config, self.file_io)
+        assert(len(dnn.layer_ops) == 2)
```
lastmansleeping, line 61: @Ullimague This doesn't really test the number of hidden units. Would be good to assert on it.

You can get it using `layer.get_config()['units']`

python/ml4ir/base/model/architectures/dnn.py
```diff
@@ -2,6 +2,14 @@
 from tensorflow.keras import layers
 from typing import List
 
+from ml4ir.base.features.feature_config import FeatureConfig
+from ml4ir.base.features.feature_fns.categorical import get_vocabulary_info
+from ml4ir.base.io.file_io import FileIO
+
+
+OOV = 1
```
lastmansleeping, line 10: @Ullimague Is it okay to hard code this?

Ullimague: Yes! It basically provides free "documentation", and we directly understand:
`layer_args["units"] = len(vocabulary_keys) + OOV`
whereas
`layer_args["units"] = len(vocabulary_keys) + 1`
we don't know what the + 1 is about. 

On top of that OOV could be reuse elsewhere if needed.

python/ml4ir/applications/classification/tests/test_model_config.py
Ullimague: Good point, added that.

lastmansleeping (dismissed):
@Ullimague Thanks for the changes!

python/ml4ir/applications/classification/tests/test_model_config.py
```diff
@@ -25,7 +25,7 @@ def test_num_classes_from_units(self):
         dnn = DNN(model_info, feature_config, self.file_io)
         assert(len(dnn.layer_ops) == 2)
         assert(dnn.layer_ops[0].get_config()['units'] == 256)
-        assert(dnn.layer_ops[4].get_config()['units'] == 9)
+        assert(dnn.layer_ops[1].get_config()['units'] == 9)
```
lastmansleeping, line 28: You should just use `[-1]`

PR #57, opened 2020-09-01
Updating the version in pom - housekeeping

darshshah:
We are using 0.0.2 internally and hence updating this to 0.0.3-SNAPSHOT.


darshshah:
@jakemannix Can you review this? This is housekeeping work so that we can checkin the correct ml4ir jar internally

jakemannix (approved):
LGTM!

PR #59, opened 2020-09-07
Adding demo notebook and clean up

lastmansleeping:
This PR covers the following changes:

- activate notebook
- saving model layers separately for transfer learning
- removing weights and biases from ml4ir
- general cleanup


python/ml4ir/base/io/local_io.py
```diff
@@ -251,3 +252,39 @@ def rm_file(self, file_path: str):
         if os.path.isfile(file_path):
             os.remove(file_path)
             self.log("File deleted : {}".format(file_path))
+
+    def save_numpy_array(self, np_array, file_path: str, allow_pickle=True, zip=True, **kwargs):
```
lastmansleeping, line 256: @jakemannix I had to do this because there is an issue with saving a list of numpy arrays directly and tensorflow layer weights are lists of arrays unfortunately. `savez` internally saves each list item in separate files and abstracts this out for us.

python/ml4ir/base/model/callbacks/debugging.py
```diff
@@ -0,0 +1,52 @@
+from tensorflow.keras import callbacks
+
+
+class DebuggingCallback(callbacks.Callback):
```
lastmansleeping, line 4: Moving this out because `mypy` wouldn't let me commit due to code complexity.

python/ml4ir/base/model/optimizer.py
```diff
@@ -7,17 +7,20 @@
 def get_optimizer(
     optimizer_key: str,
     learning_rate: float,
-    learning_rate_decay: float,
-    learning_rate_decay_steps: int,
-    gradient_clip_value: float,
+    learning_rate_decay: float = 1.0,
```
lastmansleeping, line 10: Setting default values here.

python/ml4ir/base/model/relevance_model.py
```diff
@@ -234,16 +246,25 @@ def fit(
...
+
+            # Write metrics for experiment tracking
+            train_metrics = dict()
+            for metric, value in history.history.items():
+                if not metric.startswith("val_"):
+                    train_metrics["train_{}".format(metric)] = value[0]
+                else:
+                    train_metrics[metric] = value[0]
+
+            return train_metrics
```
lastmansleeping, line 267: Returning metrics to be stored for experiment tracking. Since we are not using weights and biases, I am creating a dict manually.

python/ml4ir/base/model/relevance_model.py
```diff
@@ -381,6 +404,16 @@ def save(
                 max_sequence_size=self.max_sequence_size,
             ),
         )
+
+        # Save individual layer weights
+        self.file_io.make_directory(os.path.join(model_file, "layers"), clear_dir=True)
```
lastmansleeping, line 409: Saving weights of each layer so that we can load it later in a new model for transfer learning.

python/ml4ir/base/pipeline.py
```diff
@@ -182,54 +179,6 @@ def validate_args(self):
 
         return self
 
-    def setup_experiment_tracking_config(self):
```
lastmansleeping, line 185: Not needed anymore. This is wandb related.

python/ml4ir/base/model/optimizer.py
Ullimague: One of the drawback here is that we are overriding default tf values (with the same values, which makes things not robust/ideal).

python/ml4ir/base/model/optimizer.py
```diff
@@ -7,17 +7,20 @@
...
+    gradient_clip_value: float = 1000000,
 ) -> tf_optimizers.Optimizer:
     # Define an exponential learning rate decay schedule
-    learning_rate_schedule = ExponentialDecay(
-        learning_rate,
-        decay_steps=learning_rate_decay_steps,
-        decay_rate=learning_rate_decay,
-        staircase=True,
-    )
+    if learning_rate_decay <= 1.0:
```
Ullimague, line 15: Why do we decide on this? We also need to add a doc for this.

python/ml4ir/base/model/relevance_model.py
```diff
@@ -234,16 +246,25 @@ def fit(
...
                 validation_data=dataset.validation,
                 epochs=num_epochs,
                 verbose=True,
                 callbacks=callbacks_list,
             )
+
+            # Write metrics for experiment tracking
+            train_metrics = dict()
+            for metric, value in history.history.items():
+                if not metric.startswith("val_"):
```
Ullimague, line 262: Seems weird, can't we have `if metric.startswith("train_")`?

python/ml4ir/base/model/relevance_model.py
Ullimague: This info needs to be part of python doc. Ie explicitly define what we are returning.

python/ml4ir/base/model/relevance_model.py
Ullimague: Can you add this info as pydoc?

python/ml4ir/base/pipeline.py
```diff
@@ -351,17 +301,62 @@ def run(self):
...
+
+        job_info = pd.DataFrame.from_dict(
+            experiment_tracking_dict, orient="index", columns=["value"]
+        ).to_csv()
+
+        # Finish
+        self.finish(job_status, job_info)
+
+    def finish(self, job_status, job_info):
+        """
```
Ullimague, line 331: We need to add pydoc about what finish does.

python/ml4ir/base/model/relevance_model.py
```diff
@@ -325,20 +346,22 @@ def evaluate(
         Args:
```
Ullimague, line 346: line 323: "Evaluate the ranking model" it's actually the base model. And we would need to provide more info, because we do more than just evaluating it (returning metrics, saving layers, etc.)

python/ml4ir/base/model/relevance_model.py
```diff
@@ -325,20 +346,22 @@ def evaluate(
...
+            additional_features: Additional post processing feature functions as key value pairs
+            group_metrics_min_queries: Minimum number of queries per group to be used for group aggregate metrics
+            logs_dir: Directory to log the predictions and metrics
             logging_frequency: integer representing how often(in batches) to log status
-            metric_group_keys: list of fields to compute group based metrics on
-            save_to_file: set to True to save predictions to file like self.predict()
 
         Returns:
-            metrics and groupwise metrics as pandas DataFrames
+            metrics and groupwise metrics as pandas DataFrames and flattened metrics as a dictionary
```
Ullimague, line 355: We should define here what is expected as output. Reading the code we just see: `returns None, None, dict`
Ideal we should add an example. And what of each value in the Triple means.

python/ml4ir/base/model/scoring/scoring_model.py
```diff
@@ -72,11 +72,11 @@ def final_activation_op(self, scores, metadata_features):
 
 class RelevanceScorer(ScorerBase):
```
Ullimague, line 73: We need to add pydoc.

python/ml4ir/base/model/scoring/scoring_model.py
```diff
@@ -72,11 +72,11 @@ def final_activation_op(self, scores, metadata_features):
...
     def architecture_op(self, train_features, metadata_features):
```
Ullimague, line 74: Same here: need to add pydoc. What is an architecture_op?

Ullimague (commented):
Early comment to help iterate. Waiting for the tests for continuing the review :)

python/ml4ir/applications/classification/tests/test_classification_model.py
```diff
@@ -17,22 +17,18 @@ def run_default_pipeline(self, data_format: str):
         # Fix random seed values for repeatability
         self.set_seeds()
 
-        args: Namespace = self.get_overridden_args(data_format)
```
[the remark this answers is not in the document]
lastmansleeping: I will add it. It was giving me errors while committing because you had not imported Namespace in this module. I think you use an IDE and it doesn't catch these errors because imports from all modules are in same context.

python/ml4ir/base/features/feature_config.py
```diff
@@ -336,26 +369,29 @@ class SequenceExampleFeatureConfig(FeatureConfig):
     """Feature config overrides for data containing SequenceExample protos"""
 
     def __init__(self, features_dict, logger):
+        super(SequenceExampleFeatureConfig, self).__init__(features_dict, logger)
```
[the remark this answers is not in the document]
lastmansleeping: There's a command line argument called `tfrecord_type` which can be set to `sequence_example` or `example`

python/ml4ir/base/model/optimizer.py
lastmansleeping: Yeah, I agree. Thoughts on how we should go about this? If I don't specify defaults, then it will expect us to pass this argument to the `get_optimizer` function each time. I can wrap it in if-else clauses, but it will be very dirty.

python/ml4ir/base/model/relevance_model.py
lastmansleeping: This is not a check though. It's the key for the dictionary. How can I do `.startswith`?

python/ml4ir/base/model/relevance_model.py
lastmansleeping: I've added what each value in the triple means, right? It is (metrics, groupwise_metrics, metrics_dict)
I can add example.

python/ml4ir/base/model/scoring/scoring_model.py
lastmansleeping: I have a work item to add pydocs for all methods. Will do it.

python/ml4ir/base/model/optimizer.py
balikasg: Needs to be less than one to be decay. I agree with all your comments on doc.

python/ml4ir/base/model/optimizer.py
balikasg: The code here looks for different optimizers based on keys that are configured on `ml4ir.base.config.keys`. We can come up with a way that skips this, and given the users choice directly looks at tensorflow.keras.optimizers.USER_CHOICE, if exists it returns it, if not it returns SGD with a message that it was not found (or simply fails with the message). This removes the need to maintain `ml4ir.base.config.keys`, fizes the problem of default arguments as they will be maintained. Arguments can be passed as done here, all being optional apart from the optimizer_key. Further args, spacific to each optimizer, can be passed with the (ugly) **kwards. The last advantage is that we expose the full functionality of the optimizers to a user.

python/ml4ir/base/model/scoring/interaction_model.py
```diff
@@ -39,8 +39,8 @@ class UnivariateInteractionModel(InteractionModel):
     def __init__(
         self,
         feature_config: FeatureConfig,
-        feature_layer_keys_to_fns: dict,
         tfrecord_type: str,
```
balikasg, line 42: A naive ask here: what is exactly an Interaction model? why  do we call it like this? Is it a family of feedforward NNs or something in these lines? I have seen it for ranking and ep but I do not grasp the term well

lastmansleeping: We can work on the term. But it is the first layer of the model where you define the feature functions for each feature and then combine them into a dense layer.

Use this layer to define embeddings, LSTMs, etc.

python/ml4ir/applications/ranking/tests/test_transfer_learning.py
```diff
@@ -0,0 +1,130 @@
...
+
+from ml4ir.base.config.keys import DataFormatKey
+from ml4ir.applications.ranking.config.keys import MetricKey
+from ml4ir.applications.ranking.tests.test_base import RankingTestBase
+from ml4ir.base.data.relevance_dataset import RelevanceDataset
+from ml4ir.applications.ranking.model.ranking_model import RankingModel
+from ml4ir.base.features.feature_config import FeatureConfig
+
+
+class RankingTransferLearningTest(RankingTestBase):
```
lastmansleeping, line 15: @Ullimague Added the tests here. I tried to cover all scenarios.

python/ml4ir/base/features/feature_config.py
```diff
@@ -102,29 +108,28 @@ def get_instance(feature_config_dict: dict, tfrecord_type: str, logger: Logger):
         else:
             return SequenceExampleFeatureConfig(feature_config_dict, logger=logger)
 
-    def extract_features(self, features_dict, logger: Optional[Logger] = None):
-
+    def extract_features(self):
```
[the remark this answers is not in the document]
lastmansleeping: Added.

python/ml4ir/base/io/local_io.py
lastmansleeping: Added.

python/ml4ir/base/model/optimizer.py
lastmansleeping: @balikasg I'm not sure I know how to do this -> "looks at tensorflow.keras.optimizers.USER_CHOICE, if exists it returns it". I need an enum because when using ml4ir as a script people will enter a string. I need to a way to map that string to a subclass of `tensorflow.keras.optimizers`

python/ml4ir/base/model/relevance_model.py
lastmansleeping: Added.

python/ml4ir/base/model/relevance_model.py
lastmansleeping: Added.

python/ml4ir/base/model/relevance_model.py
lastmansleeping: Added.

python/ml4ir/base/pipeline.py
lastmansleeping: Added.

lastmansleeping:
@Ullimague I've added the unit test, pydocs and addressed all your review comments. Let me know if this works. I have a parallel work item (WIP) to add and consolidate the pydocs across ml4ir.

Ullimague (approved):
Thanks for the follow-ups @lastmansleeping!

PR #61, opened 2020-09-14
Cleaning up requirements and docker @rev calliebradley@

lastmansleeping:
This PR covers:
- cleaning up requirements.txt
- cleaning up Dockerfile for faster docker builds
- restructuring docker-compose to `ml4ir/python`
- updating `.dockerignore` for faster builds


python/.dockerignore
```diff
@@ -1,9 +1,10 @@
...
-./python/models
-./python/logs
-target/
+./data/
+./models/
+./logs/
+./env/
+./target/
 **/test_output/
 **/test_output/
```
calliebradley, line 10: [Extremely minor] Looks like we have a duplicated line here, can we remove one of the `**/test_output/` lines?

python/requirements.txt
```diff
@@ -1,91 +1,35 @@
-# external requirements
```
calliebradley, line 1: @lastmansleeping Why were there so many unneeded packages listed in the requirements? Did we copy this from another repo?

calliebradley:
@lastmansleeping Looks like references to `run_docker.sh` will need to be updated in `/applications/classification/scripts/example_run.sh` and `/applications/ranking/scripts/example_run.sh`

calliebradley:
Replacing my previous comment about `tools/run_docker.sh` after a closer look. With the rearrangement of files, looks like just the README also needs to be updated for locations of the `example_run.sh`. Works when I `cd` into the `python` directory and use `\bin\bash ml4ir/applications/ranking/scripts/example_run.sh`

python/requirements.txt
lastmansleeping: > Replacing my previous comment about `tools/run_docker.sh` after a closer look. With the rearrangement of files, looks like just the README also needs to be updated for locations of the `example_run.sh`. Works when I `cd` into the `python` directory and use `/bin/bash ml4ir/applications/ranking/scripts/example_run.sh`

Will update readme. Thanks!

python/.dockerignore
lastmansleeping: Good catch.

python/requirements.txt
lastmansleeping: They were not unnecessary really. What I did at ml4ir's inception was take a dump of the packages in my virtualenv packages. So some of the packages are nested dependencies of the packages you see here(for example, keras might be a dependency on tensorflow, which will now get automatically installed anyway). There could have been some from my other repositories as well.

calliebradley: That makes sense. Thanks!

python/README.md
```diff
@@ -20,6 +20,11 @@ This will install ml4ir-0.0.1 (the current version) from PyPI.
 #### Docker (Recommended)
 We have set up a `docker-compose.yml` file for building and using docker containers to train models.
 
+First, change working directory to the python package
+```
+cd python/
```
calliebradley, line 25: @lastmansleeping Great. Then we also just need to change line this line of the README:
`/bin/bash python/ml4ir/applications/ranking/scripts/example_run.sh`

to this (remove the `python/` directory from the command, since it will not run from the directory above):
`/bin/bash ml4ir/applications/ranking/scripts/example_run.sh`

calliebradley (approved):
Looks good (after that one last README change I commented, see last comment before this). Tested all tests and example runs with docker and venv. Great to see this kind of cleanup work for keeping good quality code, thanks @lastmansleeping !

PR #64, opened 2020-09-24
Fix incorrect input dimension computation

mbrette:
Fix incorrect input dimension computation with categorical_embedding_to_encoding_bilstm.
When this function is used with dropout_rate, it does not use oov_buckets but have a specific value for out-of-vocabulary input. This extra value must be accounted for in the embedding layer dimension.


python/ml4ir/applications/classification/pipeline.py
```diff
@@ -92,9 +92,10 @@ def get_relevance_dataset(self, parse_tfrecord=True, preprocessing_keys_to_fns={
 
         NOTE: Override this method to create custom dataset objects
         """
-        # Adding one_hot_vectorizer needed for classification.
+        # Adding one_hot_vectorizer needed for classification
         preprocessing_keys_to_fns = {
-            "one_hot_vectorize_label": get_one_hot_label_vectorizer(self.feature_config.get_label(), self.file_io)
+            "one_hot_vectorize_label": get_one_hot_label_vectorizer(self.feature_config.get_label(), self.file_io),
+            "split_and_pad_string": split_and_pad_string
```
mbrette, line 98: This was missing from classification hook (although it was used in ep models).

python/ml4ir/base/data/relevance_dataset.py
```diff
@@ -4,7 +4,7 @@
 from logging import Logger
 import tensorflow as tf
 
-from ml4ir.base.config.keys import DataFormatKey, DataSplitKey, DefaultDirectoryKey
```
mbrette, line 7: Was not used

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -220,6 +219,70 @@ def categorical_embedding_with_vocabulary_file(feature_tensor, feature_info, fil
     return embedding
 
 
+def word_sequence_embedding_with_vocabulary_file(feature_tensor, feature_info, file_io: FileIO):
```
mbrette, line 222: I added this code by mistake, will remove.

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -431,10 +425,15 @@ def categorical_indices_from_vocabulary_file(feature_info, feature_tensor, file_
 
     if "dropout_rate" in args:
         default_value = 0  # Default OOV index when using dropout
+        if args.get("num_oov_buckets"):
+            raise RuntimeError("Cannot have both dropout_rate and num_oov_buckets set. "
+                               "OOV buckets are not supported with dropout")
```
mbrette, line 430: Enforce that we don't have both at the same time.

python/ml4ir/base/features/feature_fns/categorical.py
```diff
@@ -444,7 +443,8 @@ def categorical_indices_from_vocabulary_file(feature_info, feature_tensor, file_
         feature_name=feature_info.get("node_name", feature_info["name"]),
     )
     categorical_indices = lookup_table(feature_tensor)
-    return categorical_indices, vocabulary_keys, num_oov_buckets
+
+    return categorical_indices, vocabulary_size, num_oov_buckets
```
mbrette, line 447: I refactored to return vocabulary_size. vocabulary_keys was used by callers only for computing the vocabulary size

python/ml4ir/applications/classification/pipeline.py
lastmansleeping: You don't have to add predefined functions. ml4ir will pick it up automatically.

python/ml4ir/applications/ranking/tests/test_feature_layer.py
```diff
@@ -135,6 +135,91 @@ def test_categorical_embedding_to_encoding_bilstm_file_truncation(self):
...
+            "tfrecord_type": SequenceExampleTypeKey.CONTEXT,
+        }
+
+        string_tensor = tf.constant([[['AAA', 'BBB', 'out_of_vocabulary']]])
+
+        categorical_fns.categorical_embedding_to_encoding_bilstm(
+            string_tensor, feature_info, self.file_io
+        )
+
+    def test_categorical_embedding_to_encoding_bilstm_oov_mapping_with_dropout(self):
```
lastmansleeping, line 167: @mbrette Wasn't this test already there? Why do I see it in the diff?

python/ml4ir/base/data/relevance_dataset.py
```diff
@@ -65,7 +65,7 @@ def create_dataset(self, parse_tfrecord=True):
         elif self.data_format == DataFormatKey.TFRECORD:
             data_reader = tfrecord_reader
         else:
-            raise NotImplementedError
+            raise NotImplementedError("Unsupported data format: {}".format(self.data_format))
```
balikasg, line 68: Looks good to me, but if you do another pass, please add ("Unsupported data format: {}. We currenty support {} and {}.".format(self.data_format, DataFormatKey.CSV, DataFormatKey. TFRECORD)) so that a user know what to do

python/ml4ir/base/features/feature_fns/categorical.py
balikasg: Do we know why? I had the impression that there was an issue with this Ashish was working on, but is it confirmed?

python/ml4ir/base/features/feature_layer.py
```diff
@@ -87,9 +87,10 @@ def feature_layer_op(inputs):
...
 
             if "fn" in feature_layer_info:
-                feature_tensor = feature_layer_map.get_fn(feature_layer_info["fn"])(
-                    feature_tensor=feature_tensor, feature_info=feature_info, file_io=file_io
-                )
+                feature_fn = feature_layer_map.get_fn(feature_layer_info["fn"])
+                if not feature_fn:
+                    raise RuntimeError("Unsupported feature function: {}".format(feature_layer_info["fn"]))
+                feature_tensor = feature_fn(feature_tensor=feature_tensor, feature_info=feature_info, file_io=file_io)
             elif feature_info["trainable"]:
```
balikasg, line 94: why not `else` ?

balikasg: super minor though

python/ml4ir/applications/classification/pipeline.py
mbrette: Great. Why do we have to add one_hot_vectorize_label ? Because the tf.function is nested inside ?

python/ml4ir/base/features/feature_fns/categorical.py
mbrette: I don't know. This just reflect what is already in the code.

python/ml4ir/applications/ranking/tests/test_feature_layer.py
mbrette: I don't think so. I added 2 tests. One that use a sequence of words as input (all the other tests were using batches of single word).
The second one, comprehensively go through all the vocabulary terms, which was not there either.

python/ml4ir/base/features/feature_layer.py
mbrette: I disagree with that. 

Here I raise an error in the middle of the function, it will skip the rest of the function.

If I put an else for the next row, it feels like 'either we raise an exception or (else) we assign feature_tensor', when actually the behavior is: raise an exception or continue evaluate the rest of the code (feature_tensor=.. but also everything that would not be in the else statement.

balikasg: My comment was simpler, it does not have anything to do with the exception, the comment was for the outer level of the code. I just suggested to change the `elif` of line 94 to `else` to avoid the pattern if-elif without the else..  I checked again, this was not code you touched though.

mbrette: ha ha, sorry. I did not change it, right.

Ullimague (commented):
Looks good to me, could we also update https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/applications/classification/tests/data/configs/feature_config.yaml to add the word feature as example in the config @mbrette?

python/ml4ir/applications/classification/tests/test_classification_model.py
```diff
@@ -45,7 +45,9 @@ def test_csv(self):
         metrics = self.run_default_pipeline(data_format="csv")
 
         # Check if the loss and accuracy on the test set is the same
-        assert np.isclose(metrics["loss"], 1.966392993927002, rtol=0.01)
-        assert np.isclose(metrics["categorical_accuracy"], 0.18229167, rtol=0.01)
-        assert np.isclose(metrics["Precision"], 0.2, rtol=0.01)
```
mbrette, line 50: I removed Precision because it is not relevant for our model.
Basically, what Precision do by default is to consider 'true' a prediction that is above 0.5 (0.5 is the default threshold).
There is a parameter that allows to instead consider the best score (top_k=1), but then the metric will be the same as CategoricalAccuracy I think.

mbrette:
> Looks good to me, could we also update https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/applications/classification/tests/data/configs/feature_config.yaml to add the word feature as example in the config @mbrette?

Ullimague (dismissed):
Approve to merge it already, could we create a follow-up to (as part of this story):
- lower case + use regex for word split?

Ullimague (dismissed):

python/ml4ir/applications/classification/pipeline.py
lastmansleeping: @Ullimague should be able to answer this better.

python/ml4ir/base/features/feature_fns/categorical.py
lastmansleeping: Nice.

python/ml4ir/applications/ranking/tests/test_feature_layer.py
lastmansleeping: Ah. Okay. I see. I misread. My bad. Good check to have.

python/ml4ir/base/features/feature_fns/categorical.py
lastmansleeping: By the nature of how "dropout" is defined on categorical labels, you can only assign it to only 1 bucket. The reason I introduced this notion of a dropout on categorical labels is to make sure that each category contributes to the OOV embedding(instead of only low count categories). (I'm trying to be really generic in my wording here haha)

But yeah, we can modify it to use multiple buckets. Feel free to do it if you want.
Check out how the VocabLoopkup differs for the two cases.

lastmansleeping: Essentially, I needed a neat way to "fix" an OOV token. I can't do it easily with `StaticVocabularyTable` as it depends on number of unique keys and such. Again, can be done in some way. But wasn't really critical.

python/ml4ir/applications/classification/tests/data/configs/feature_config.yaml
```diff
@@ -65,6 +65,10 @@ features:
     log_at_inference: true
     is_group_metric_key: true
     preprocessing_info:
+      - fn: preprocess_text
```
mbrette, line 68: This combination of preprocessing effectively perform the tokenization we have with the legacy EP model.

balikasg (approved):
thank you for all the effort here

PR #65, opened 2020-09-24
Prototyping factory for optimizers

balikasg:
A common pattern is to use `ml4ir.base.config.keys` to use some keywords to control functionality. This is error prone, difficult to maintain and can cause limitations. 

This PR starts this discussion proposing an alternative way to do it. Happy to hear back your thought. 

It is not finished work, it just points the direction. But if we were to agree on something like this it means we automatically support everything tf does. And we can make the `ml4ir.base.config.keys` smaller. There are also other places that will benefit (discussed offline with Ashish).

### DIFF
--- a/python/ml4ir/applications/classification/pipeline.py
+++ b/python/ml4ir/applications/classification/pipeline.py
@@ -69,7 +69,7 @@ class ClassificationPipeline(RelevancePipeline):
             learning_rate=self.args.learning_rate,
             learning_rate_decay=self.args.learning_rate_decay,
             learning_rate_decay_steps=self.args.learning_rate_decay_steps,
-            gradient_clip_value=self.args.gradient_clip_value,
+            clipvalue=self.args.gradient_clip_value,
         )
 
         # Combine the above to define a RelevanceModel
--- a/python/ml4ir/applications/ranking/tests/test_base.py
+++ b/python/ml4ir/applications/ranking/tests/test_base.py
@@ -126,7 +126,7 @@ class RankingTestBase(unittest.TestCase):
             learning_rate=self.args.learning_rate,
             learning_rate_decay=self.args.learning_rate_decay,
             learning_rate_decay_steps=self.args.learning_rate_decay_steps,
-            gradient_clip_value=self.args.gradient_clip_value,
+            clipvalue=self.args.gradient_clip_value,
         )
 
         # Combine the above to define a RelevanceModel
--- a/python/ml4ir/base/config/parse_args.py
+++ b/python/ml4ir/base/config/parse_args.py
@@ -59,7 +59,7 @@ class RelevanceArgParser(ArgumentParser):
         self.add_argument(
             "--optimizer_key",
             type=str,
-            default="adam",
+            default="Adam",
             help="Optimizer to use. Has to be one of the optimizers in OptimizerKey under "
             "ml4ir/config/keys.py",
         )
--- a/python/ml4ir/base/model/optimizer.py
+++ b/python/ml4ir/base/model/optimizer.py
@@ -1,13 +1,10 @@
 import tensorflow.keras.optimizers as tf_optimizers
 from tensorflow.keras.optimizers.schedules import ExponentialDecay
 
-from ml4ir.base.config.keys import OptimizerKey
-
 
 def get_optimizer(
-    optimizer_key: str = "SGD",
-    with_exp_decay: bool = False,
-    initial_learning_rate: float = 0.01,
+    optimizer_key: str = "Adam",
+    learning_rate: float = 0.01,
     learning_rate_decay: float = 1.0,
     learning_rate_decay_steps: int = 1000000,
     **kwargs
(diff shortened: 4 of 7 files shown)


python/ml4ir/base/model/optimizer.py
```diff
@@ -5,58 +5,61 @@
...
-        return tf_optimizers.SGD(
-            learning_rate=learning_rate_schedule, clipvalue=gradient_clip_value
-        )
-    elif optimizer_key == OptimizerKey.RMS_PROP:
-        return tf_optimizers.RMSprop(
-            learning_rate=learning_rate_schedule, clipvalue=gradient_clip_value
-        )
-    else:
-        raise ValueError("illegal Optimizer key: " + optimizer_key)
+    try:
```
balikasg, line 57: This gist of the PR is here. 
These lines of code enable you to use any tf optimizer and pass any keyword arguments as long as you respect the tf names. So to use tensorflow.keras.optimizers.Adam, the key needs to be 'Adam' ('adam') will fail. Same goes for any specific keyword args of the optimizer.

python/ml4ir/base/pipeline.py
```diff
@@ -149,13 +149,6 @@ def validate_args(self):
                 )
             )
 
-        if self.optimizer_key not in OptimizerKey.get_all_keys():
```
balikasg, line 152: I removing this validation, because this is inherent to the `get_optimizer` function.

python/ml4ir/applications/ranking/tests/test_base.py
```diff
@@ -126,7 +126,7 @@ def get_ranking_model(
             learning_rate=self.args.learning_rate,
             learning_rate_decay=self.args.learning_rate_decay,
             learning_rate_decay_steps=self.args.learning_rate_decay_steps,
-            gradient_clip_value=self.args.gradient_clip_value,
+            clipvalue=self.args.gradient_clip_value,
```
balikasg, line 129: this needs to be clipvalue now, because this is the tf variable name, and its name + value will be passed directly there. I know you prefer 'clip_value', but this is TF not me.

python/ml4ir/base/config/parse_args.py
```diff
@@ -59,7 +59,7 @@ def define_args(self):
         self.add_argument(
             "--optimizer_key",
             type=str,
-            default="adam",
+            default="Adam",
```
balikasg, line 62: Change needed because we need to match exactly the name of the class at `tf.keras.optimizers`

python/ml4ir/base/model/optimizer.py
```diff
@@ -5,58 +5,61 @@
...
+    Examples:
+        get_optimizer() -> SGD optimizer
+        get_optimizer('Adam', learning_rate=0.002) -> Adam, with lr=0.002
     FIXME:
         Define all arguments overriding tensorflow defaults in a separate file
         for visibility with ml4ir users
+
     """
     # Define an exponential learning rate decay schedule
     learning_rate_schedule = ExponentialDecay(
```
balikasg, line 50: I think that schedulers should in their own `get_scheduler` function. But good for now.

python/ml4ir/base/model/optimizer.py
balikasg: If you like this, we should do similarly with metrics, layers etc. 
Some pointers where we could improve a lot: 
- https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/base/model/architectures/dnn.py#L48
- https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/applications/ranking/model/metrics/metric_factory.py

mbrette (approved):

Ullimague (approved):

python/ml4ir/applications/ranking/tests/test_base.py
lastmansleeping: Haha. No worries.

python/ml4ir/base/config/parse_args.py
lastmansleeping: Can you update config keys?

python/ml4ir/applications/classification/pipeline.py
```diff
@@ -69,7 +69,7 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
             learning_rate=self.args.learning_rate,
             learning_rate_decay=self.args.learning_rate_decay,
             learning_rate_decay_steps=self.args.learning_rate_decay_steps,
-            gradient_clip_value=self.args.gradient_clip_value,
+            clipvalue=self.args.gradient_clip_value,
```
lastmansleeping, line 72: @balikasg Why are you not passing `with_exp_decay` here?

Also, looks like there is mismatch between `learning_rate` and `initial_learning_rate`. Maybe I am missing something actually.

python/ml4ir/base/config/parse_args.py
balikasg: Yes. I need to do this.

balikasg:
Closing as part of it is in https://github.com/salesforce/ml4ir/pull/89/files

PR #67, opened 2020-09-28
trying to add new metric (TopKCategoricalAccuracy)

balikasg:
This PR tries to add TOP_K_CATEGORICAL_ACCURACY (TopKCategoricalAccuracy from tf.keras.metrics)

Unfortunately, the tests fail with incopatible tensor shapes, but I do not understand why as CategoricalAccuracy and TopKCategoricalAccuracy have the same APIs.

The error I get: 
```
ValueError: Shape must be rank 2 but is rank 3 for ‘in_top_k/InTopKV2’ (op: ‘InTopKV2’) with input shapes: [?,1,9], [?,?], [].”
The thing is TopKClassificationAccuracy has exactly the same API with Classification Accuracy. 
```

I can not understand how one can succeed and the other fail. 
As a matter of fact, my implementation follows exactly the implementation of `CategoricalAccuracy`

If anyone can help, it is appreciated..


balikasg:
@Ullimague @lastmansleeping if you have any hints I would happily take them. I have spent several hours on this (seemingly) simple task

balikasg:
update: no progress here from my side. If you have any suggestion to look at, I will use them

lastmansleeping:
@balikasg Sorry, I missed this. Feel free to ping me on slack if you see no progress from my side.

Let's set up a quick meeting next week to discuss this. I will take a look and try to add comments before then.

jakemannix:
> This PR tries to add TOP_K_CATEGORICAL_ACCURACY (TopKCategoricalAccuracy from tf.keras.metrics)
> 
> Unfortunately, the tests fail with incopatible tensor shapes, but I do not understand why as CategoricalAccuracy and TopKCategoricalAccuracy have the same APIs.
> 
> The error I get:
> 
> ```
> ValueError: Shape must be rank 2 but is rank 3 for ‘in_top_k/InTopKV2’ (op: ‘InTopKV2’) with input shapes: [?,1,9], [?,?], [].”
> The thing is TopKClassificationAccuracy has exactly the same API with Classification Accuracy. 
> ```
> 
> I can not understand how one can succeed and the other fail.
> As a matter of fact, my implementation follows exactly the implementation of `CategoricalAccuracy`
> 
> If anyone can help, it is appreciated..

@Ullimague this looks like the issue we were seeing with JVM inference.  

@balikasg  I wonder if the tensors are wrapping one extra layer on the outside?  So in this case, we just need to grab the "first" (equivalently "last") entry in the tensor we have, via a simple `tf.squeeze` on it, before computing the metric?

balikasg:
@jakemannix @Ullimague  I just fixed this. The tests should pass now
the issue comes from an extra dimension that exists to accommodate list-wise losses

jakemannix:
> @jakemannix @Ullimague I just fixed this. The tests should pass now
> the issue comes from an extra dimension that exists to accommodate list-wise losses

Does it lead to a situation where that new metric will only work with Example input (not SequenceExample)?  Which might be fine, but do we have any way to ensure users don't accidentally add a metric of the "wrong" kind to a model?

Ullimague:
@balikasg thanks for tackling this!

Question question:
1. Could we add a test for k=1-3-5?
2. How do we pass k as agrument?

lastmansleeping:
> @balikasg thanks for tackling this!
> 
> Quick question:
> 
> 1. Could we add a test for k=1-3-5?
> 2. How do we pass k as agrument?

1. Should be easy, although I don't think it is super important
2. You can pass k when you create the `TopKCategoricalAccuracy(k=10)`

lastmansleeping:
> > @jakemannix @Ullimague I just fixed this. The tests should pass now
> > the issue comes from an extra dimension that exists to accommodate list-wise losses
> 
> Does it lead to a situation where that new metric will only work with Example input (not SequenceExample)? Which might be fine, but do we have any way to ensure users don't accidentally add a metric of the "wrong" kind to a model?

I like the idea of adding a check for which metrics can be used with which model(or data format).

python/ml4ir/applications/classification/config/parse_args.py
```diff
@@ -18,7 +18,7 @@ def set_default_args(self):
             tfrecord_type=TFRecordTypeKey.EXAMPLE,
             loss_key=LossKey.CATEGORICAL_CROSS_ENTROPY,
             metrics_keys="['" + MetricKey.CATEGORICAL_ACCURACY + "'"
-                         + ", '" + MetricKey.PRECISION + "'"
+                         + ", " + "'top_k_categorical_accuracy'"
```
lastmansleeping, line 21: @balikasg Why are you not adding this to the MetricKey?

lastmansleeping: I don't know if you want to change the default value though. Up to you.

python/ml4ir/applications/ranking/model/metrics/metrics_impl.py
```diff
@@ -219,3 +219,27 @@ def __init__(
...
+    def __init__(
+        self,
+        feature_config: FeatureConfig,
+        metadata_features: Dict,
+        name="top_k_categorical_accuracy",
+        state=MetricState.NEW,
+        **kwargs
+    ):
+        super(TopKCategoricalAccuracy, self).__init__(name=name)
+    def update_state(self, y_true, y_pred, sample_weight=None):
```
lastmansleeping, line 242: [Minor] new line?

python/ml4ir/applications/ranking/model/metrics/metrics_impl.py
```diff
@@ -219,3 +219,27 @@ def __init__(
...
+        return super(TopKCategoricalAccuracy, self).update_state(
```
lastmansleeping, line 243: @balikasg Add a comment here for why you need to do this. Basically input data shape and expected data shape by metric.

lastmansleeping (dismissed):
LGTM. @balikasg and I synced offline over getting this to work.

python/ml4ir/applications/ranking/model/metrics/metrics_impl.py
balikasg: yes, I am doing this now.

balikasg:
@lastmansleeping please take a look again. I simplified things a bit using your suggestions, I will start the work on the other parts now

python/ml4ir/applications/ranking/model/metrics/metrics_impl.py
```diff
@@ -219,3 +219,31 @@ def __init__(
...
+        metadata_features: Dict,
+        name="top_k_categorical_accuracy",
+        state=MetricState.NEW,
+        **kwargs
+    ):
+        super(Top5CategoricalAccuracy, self).__init__(name=name)
+
+    def update_state(self, y_true, y_pred, sample_weight=None):
+        """Input shape is a 3 dimensional tensor of size
+        (batch_size, 1, prediction_cardinality). We are squeezing
```
lastmansleeping, line 245: [Minor] `num_classes`?

python/ml4ir/applications/ranking/tests/test_metrics.py
```diff
@@ -94,3 +97,15 @@ def test_model_training(self):
         for gold_metric_name, gold_metric_val in GOLD_METRICS.items():
             assert gold_metric_name in metrics
             assert np.isclose(metrics[gold_metric_name], gold_metric_val, rtol=0.01)
+
+
+def test_top_5_categorical_accuracy():
```
lastmansleeping, line 102: @balikasg Can probably put this in the class itself, no? I understand why you didn't. We should restructure these tests out of ranking and into the base package at some point. I think I have an issue filed for it.

python/ml4ir/applications/classification/config/parse_args.py
```diff
@@ -18,7 +18,7 @@ def set_default_args(self):
             tfrecord_type=TFRecordTypeKey.EXAMPLE,
             loss_key=LossKey.CATEGORICAL_CROSS_ENTROPY,
             metrics_keys="['" + MetricKey.CATEGORICAL_ACCURACY + "'"
-                         + ", '" + MetricKey.PRECISION + "'"
+                         + ", " + "'top_5_categorical_accuracy'"
```
lastmansleeping, line 21: Checking back on why we are not using `MetricKey.TOP_5_CATEGORICAL_ACCURACY`

lastmansleeping, line 21:
```suggestion
                         + ", " + "'{}'".format(MetricKey.TOP_5_CATEGORICAL_ACCURACY)
```

python/ml4ir/applications/ranking/tests/test_metrics.py
```diff
@@ -94,3 +97,15 @@ def test_model_training(self):
...
+    """Test for calculating top_5_categorical_accuracy metric"""
+    feature_config = os.path.join(ROOT_DATA_DIR, "configs", "feature_config.yaml")
+    feature_config = LocalIO().read_yaml(feature_config)
+    metric = Top5CategoricalAccuracy(feature_config=feature_config, metadata_features={})
```
lastmansleeping, line 106: I recommend passing in a None object for feature_config here instead of loading one. @balikasg

python/ml4ir/applications/ranking/model/metrics/metrics_impl.py
```diff
@@ -219,3 +219,31 @@ def __init__(
...
+
+    Currently a wrapper around tf.keras.metrics.TopKCategoricalAccuracy that
+    squeezes one dimension.
+    It maintains consistency of arguments to __init__
+    """
+
+    def __init__(
+        self,
+        feature_config: FeatureConfig,
+        metadata_features: Dict,
```
lastmansleeping, line 236:
```suggestion
        feature_config: Optional[FeatureConfig]=None,
        metadata_features: Dict={},
```

lastmansleeping (approved):
LGTM. Thanks for the addition @balikasg :)

PR #68, opened 2020-09-28
Tokenize punctuation

mbrette:
Add support to replace punctuation by whitespace, as well as custom value for punctuation.


mbrette:
I should not try to have modification from 2 branch ahead of master. Let's kill this PR until I merged the previous oun.

PR #69, opened 2020-09-28
Tokenize punctuation

mbrette:
Support for replacing punctuation with whitespace and provide custom punctuation character list.


python/ml4ir/base/features/preprocessing.py
```diff
@@ -37,7 +32,8 @@ def pop_fn(self, key):
 
 
 @tf.function
-def preprocess_text(feature_tensor, remove_punctuation=False, to_lower=False):
+def preprocess_text(feature_tensor, remove_punctuation=False, to_lower=False,
+                    punctuation=string.punctuation, replace_with_whitespace=False):
```
mbrette, line 36: I chose to have 'replace_with_whitespace' instead of specifying your own replacement because it is simpler. With your own replacement, I would have to go through some length to also strip leading and trailing such replacement (which would not be so straightforward to understand from an api perspective).

python/ml4ir/base/features/preprocessing.py
```diff
@@ -46,12 +42,23 @@ def preprocess_text(feature_tensor, remove_punctuation=False, to_lower=False):
...
+        punctuation: punctuation characters to replace (a single string containing the character to remove
+        replace_with_whitespace: if True punctuation will be replaced by whitespace (i.e. used as separator), note that
+                               leading and trailing whitespace will also be removed, as well as consecutive whitespaces.
 
     Returns:
         processed float tensor
     """
     if remove_punctuation:
-        feature_tensor = tf.strings.regex_replace(feature_tensor, PUNCTUATION_REGEX, "")
+        replacement = ""
```
balikasg, line 53: we could make this more pythonic as `replacement = " " if replace_with_whitespace else ""`

balikasg (approved):
LGTM. For the new changes to take effect, we need to publish the ml4ir from MLConfigs. When merging, you can ping me (I have a bug open on this)

PR #70, opened 2020-09-30
Add dataset prefetch to improve performance.

mbrette:
Improved training by 30% in one of my entity prediction test.

### DIFF
--- a/python/ml4ir/base/data/relevance_dataset.py
+++ b/python/ml4ir/base/data/relevance_dataset.py
@@ -116,7 +116,7 @@ class RelevanceDataset:
                 parse_tfrecord=parse_tfrecord,
                 file_io=self.file_io,
                 logger=self.logger,
-            ).prefetch(tf.data.experimental.AUTOTUNE)
+            )
             self.validation = data_reader.read(
                 data_dir=os.path.join(self.data_dir, DataSplitKey.VALIDATION),
                 feature_config=self.feature_config,
@@ -129,7 +129,7 @@ class RelevanceDataset:
                 parse_tfrecord=parse_tfrecord,
                 file_io=self.file_io,
                 logger=self.logger,
-            ).prefetch(tf.data.experimental.AUTOTUNE)
+            )
             self.test = data_reader.read(
                 data_dir=os.path.join(self.data_dir, DataSplitKey.TEST),
                 feature_config=self.feature_config,
@@ -142,7 +142,7 @@ class RelevanceDataset:
                 parse_tfrecord=parse_tfrecord,
                 file_io=self.file_io,
                 logger=self.logger,
-            ).prefetch(tf.data.experimental.AUTOTUNE)
+            )
 
     def balance_classes(self):
         """
--- a/python/ml4ir/base/data/relevance_dataset.py
+++ b/python/ml4ir/base/data/relevance_dataset.py
@@ -100,6 +100,9 @@ class RelevanceDataset:
                 ├── data_file
                 ├── ...
                 └── data_file
+                
+            We also apply prefetch(tf.data.experimental.AUTOTUNE) as it improved train/test/validation throughput 
+            by 30% in some real model training.
             """
             self.train = data_reader.read(
                 data_dir=os.path.join(self.data_dir, DataSplitKey.TRAIN),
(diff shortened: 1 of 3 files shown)


python/ml4ir/base/data/relevance_dataset.py
```diff
@@ -113,7 +116,7 @@ def create_dataset(self, parse_tfrecord=True):
                 parse_tfrecord=parse_tfrecord,
                 file_io=self.file_io,
                 logger=self.logger,
-            )
+            ).prefetch(tf.data.experimental.AUTOTUNE)
```
mbrette, line 119: I don't think this is worth adding a unit test (note that the code is exercised in integration tests).

lastmansleeping: @mbrette I recommend making the change here instead -> https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/base/data/tfrecord_reader.py#L352

lastmansleeping: Thoughts?

mbrette: I considered this, but then I was worried that it would not be exercised by tests, or not all tests.

lastmansleeping: It should be. Everything that hits this code line will hit the same in tfrecord_reader as well. Luckily or unluckily.

mbrette: No, I meant that some tests use CSV dataset. It will not be exercised there

lastmansleeping: It should be. The way ml4ir handles CSV is by converting it into TFRecords. Check here -> https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/base/data/csv_reader.py#L70

lastmansleeping: Did you see any tests failing?

balikasg: the point where you suggest to put it is conditional to the user passing a `batch_size`. I would be OK with any choice. I would even suggest [this line](https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/base/data/tfrecord_reader.py#L361), but I don't think this is a blocker.

balikasg (approved):
I am to approve. I think it is better to move `prefetch` inside the function, to avoid also replicating this for train/validation/test. 
But we should unblock this PR

python/ml4ir/base/data/relevance_dataset.py
lastmansleeping: @balikasg That was intentional. `prefetch` can only work if you have batched your dataset? What does prefetch even do otherwise? It will probably just use batch size = 1. That is not really what we are trying to do, right?

mbrette: I followed both your suggestions. prefetch is now performed as a last step in tfrecord_reader.

python/ml4ir/base/data/tfrecord_helper.py
```diff
@@ -55,6 +55,8 @@ def get_example_proto(row, features):
         # that breaks this part of the code. Example:
         # https://stackoverflow.com/questions/47143631/
         # how-do-i-preserve-datatype-when-using-apply-row-wise-in-pandas-dataframe
+        if feature_name not in row:
```
mbrette, line 58: Got bitten by this one, making the error explicit.

python/ml4ir/base/data/tfrecord_reader.py
```diff
@@ -345,7 +345,10 @@ def read(
     # Parse the protobuf data to create a TFRecordDataset
     dataset = data.TFRecordDataset(tfrecord_files)
     if parse_tfrecord:
-        dataset = dataset.map(parse_fn).apply(data.experimental.ignore_errors())
+        # Parallel calls set to AUTOTUNE: improved training performance by 40% with a classification model
```
mbrette, line 348: Added this additional improvement. So now with both improvement, training time is improved by 60%

balikasg: this is amazing!

balikasg (approved):

python/ml4ir/base/data/tfrecord_reader.py
```diff
@@ -358,4 +361,7 @@ def read(
             )
         )
 
+    # We apply prefetch as it improved train/test/validation throughput by 30% in some real model training.
+    dataset = dataset.prefetch(tf.data.experimental.AUTOTUNE)
```
lastmansleeping, line 365: @mbrette I don't think we should have prefetch when we are not batching. I'm going to fix this in a follow up PR.

mbrette: @lastmansleeping Why ? (I guess this may not be useful, but does it hurt ?and if not, we should not make a special case)
Also when do we not batch ?

PR #74, opened 2020-10-04
Exception improvement

mbrette:


python/ml4ir/base/features/feature_config.py
```diff
@@ -116,32 +116,27 @@ def extract_features(self):
         - label
         - all_features
         """
-        try:
-            self.query_key = self.features_dict.get(FeatureConfigKey.QUERY_KEY)
-            self.all_features.append(self.query_key)
-        except KeyError:
```
mbrette, line 122: I think KeyError would never happen with this code. features_dict[foo] would raise KeyError, but features_dict.get(foo) just return None.

So I'm now checking upfront and explicitly raising an error.

balikasg: you are right. `dict.get` returns `None` be default. Good catch!

python/ml4ir/base/features/feature_config.py
```diff
@@ -116,32 +116,27 @@ def extract_features(self):
...
-            self.query_key = None
+        for mandatory_key in [FeatureConfigKey.LABEL, FeatureConfigKey.FEATURES]:
```
balikasg, line 119: from the code I see that the `mandatory` should be 3 ( I count 3 `keyError` exceptions): 
`FeatureConfigKey.LABEL, FeatureConfigKey.FEATURES, FeatureConfigKey.QUERY_KEY` no?

balikasg (commented):
The PR LGTM. 
I am wondering if we should have 3 values in the `mandatory` you defined.

python/ml4ir/base/features/feature_config.py
mbrette: In the previous code, for QUERY_KEY, we catched a KeyError, but did not raise a KeyError.
In the new code, we check 'not in self.features_dict' 3 times too (2 for mandatory keys, one for QUERY_KEY line 124).

lastmansleeping: There is also `RANK` but only for SequenceExample

lastmansleeping (approved):

PR #82, opened 2020-10-15
Allowing the user to save a model bundle and vocabulary files

balikasg:
I am adding code to create the necessary files when training a model for deployment. 
Essentially, apart from the SUCCESS/FAILURE files, we create a model bundle and we pack also the files used for vocabulary lookups.

### DIFF
--- a/python/ml4ir/applications/classification/pipeline.py
+++ b/python/ml4ir/applications/classification/pipeline.py
@@ -1,24 +1,20 @@
 import sys
 import ast
-import os
-import shutil
 from argparse import Namespace
-from pathlib import Path
-from tensorflow.keras.metrics import Metric, Precision
+from tensorflow.keras.metrics import Metric
 from tensorflow.keras.optimizers import Optimizer
 
 from ml4ir.applications.classification.config.parse_args import get_args
 from ml4ir.applications.classification.model.losses import categorical_cross_entropy
 from ml4ir.applications.classification.model.metrics import metrics_factory
 from ml4ir.base.data.relevance_dataset import RelevanceDataset
-from ml4ir.base.features.preprocessing import get_one_hot_label_vectorizer, split_and_pad_string
+from ml4ir.base.features.preprocessing import get_one_hot_label_vectorizer
 from ml4ir.base.model.losses.loss_base import RelevanceLossBase
 from ml4ir.base.model.optimizer import get_optimizer
 from ml4ir.base.model.relevance_model import RelevanceModel
 from ml4ir.base.model.scoring.scoring_model import ScorerBase, RelevanceScorer
 from ml4ir.base.model.scoring.interaction_model import InteractionModel, UnivariateInteractionModel
 from ml4ir.base.pipeline import RelevancePipeline
-from ml4ir.base.config.keys import FileHandlerKey
 from typing import Union, List, Type
 
 
@@ -169,31 +165,6 @@ class ClassificationPipeline(RelevancePipeline):
         )
 
         return relevance_dataset
-
-    def finish(self, job_status, job_info):
-        # This will write the _SUCCESS/_FAILURE files
-        super().finish(job_status, job_info)
-
-        # Create model_bundle.zip depending on user args
-        if self.args.write_tf_model_dir:
-            os.makedirs(self.args.write_tf_model_dir)
-            shutil.copytree(os.path.join(self.models_dir_local,
-                                         "final/tfrecord/"), "model_files")
-            shutil.make_archive("{}/model_bundle".format(self.args.write_tf_model_dir),
-                                "zip", "model_files")
-
-            if self.args.write_config_csvs_by_name:
-                for config_key in self.args.write_config_csvs_by_name:  # a list can be provided
-                    for lookup_file in self.feature_config.get_all_values(key=config_key):
-                        out_path = Path(lookup_file).stem + ".csv"
-                        (self.file_io
-                         .read_df(lookup_file)  # needs a str here, not a Path object
-                         .to_csv(out_path))
-                        shutil.make_archive("{}/parameters/{}".format(self.args.write_tf_model_dir,
-                                                                      out_path), "zip", ".", out_path)
-            if self.args.file_handler == FileHandlerKey.SPARK:
-                self.file_io.copy_to_hdfs(self.args.write_tf_model_dir,
-                                          self.models_dir, overwrite=False)
 
 
 def main(argv):
--- a/python/ml4ir/applications/classification/pipeline.py
+++ b/python/ml4ir/applications/classification/pipeline.py
@@ -1,5 +1,6 @@
 import sys
 import ast
+import shutil
 from argparse import Namespace
 
 from tensorflow.keras.metrics import Metric, Precision
@@ -16,7 +17,7 @@ from ml4ir.base.model.relevance_model import RelevanceModel
 from ml4ir.base.model.scoring.scoring_model import ScorerBase, RelevanceScorer
 from ml4ir.base.model.scoring.interaction_model import InteractionModel, UnivariateInteractionModel
 from ml4ir.base.pipeline import RelevancePipeline
-
+from ml4ir.base.config.keys import FileHandlerKey
 from typing import Union, List, Type
 
 
@@ -167,6 +168,25 @@ class ClassificationPipeline(RelevancePipeline):
         )
 
         return relevance_dataset
+
+    def finish(self, job_status, job_info):
+        # This will write the _SUCCESS/_FAILURE files
+        super().finish(job_status, job_info)
+
+        # Create model_bundle.zip
+        if self.args.file_handler == FileHandlerKey.SPARK:
+            # Copy files in model_files
+            self.local_io.make_directory('model_files', clear_dir=True)
+            shutil.copytree(self.models_dir_local+"/final/tfrecord/", "model_files")
+            # Zip model_files to model_bundle.zip
+            shutil.make_archive("model_bundle", "zip", "model_files")
+            # Put them to HDFS
+            self.file_io.copy_to_hdfs("model_bundle.zip", self.models_dir, overwrite=True)
+
+        # Create org_id_to_key_prefix.csv.zip
+        shutil.make_archive("org_id_to_key_prefix.csv", "zip", dir_name)
+
+        # Create parameters/key_prefix_labels.csv
 
 
 def main(argv):
--- a/python/ml4ir/applications/classification/pipeline.py
+++ b/python/ml4ir/applications/classification/pipeline.py
@@ -1,5 +1,6 @@
 import sys
 import ast
+import os
 import shutil
 from argparse import Namespace
 
@@ -175,18 +176,26 @@ class ClassificationPipeline(RelevancePipeline):
 
         # Create model_bundle.zip
         if self.args.file_handler == FileHandlerKey.SPARK:
-            # Copy files in model_files
-            self.local_io.make_directory('model_files', clear_dir=True)
-            shutil.copytree(self.models_dir_local+"/final/tfrecord/", "model_files")
-            # Zip model_files to model_bundle.zip
-            shutil.make_archive("model_bundle", "zip", "model_files")
-            # Put them to HDFS
-            self.file_io.copy_to_hdfs("model_bundle.zip", self.models_dir, overwrite=True)
+            # Copy tfrecord files in model_files
+            os.makedirs("epmm")
+            shutil.copytree(os.path.join(self.models_dir_local, "final/tfrecord/"), "model_files")
+            shutil.make_archive("./epmm/model_bundle", "zip", "model_files")
+
+            (self.file_io
+             .read_df(self.feature_config.get_label()['feature_layer_info']['args']['vocabulary_file'])
+             .to_csv("key_prefix_labels.csv"))
+            shutil.make_archive("./epmm/key_prefix_labels.csv", "zip", ".", "key_prefix_labels.csv")
 
         # Create org_id_to_key_prefix.csv.zip
-        shutil.make_archive("org_id_to_key_prefix.csv", "zip", dir_name)
+            org_id = [x for x in self.feature_config.get_all_features() if x['name'] == 'organization_id'][0]
+            org_mapping = org_id['feature_layer_info']['args']['vocabulary_file']
+            (self.file_io
+             .read_df(org_mapping)
+             .to_csv("org_id_to_key_prefix.csv"))
+            shutil.make_archive("./epmm/org_id_to_key_prefix.csv", "zip", ".", "org_id_to_key_prefix.csv" )
+            self.file_io.copy_to_hdfs("epmm", self.models_dir, overwrite=False)
 
-        # Create parameters/key_prefix_labels.csv
+        # This will write the _SUCCESS/_FAILURE files
 
 
 def main(argv):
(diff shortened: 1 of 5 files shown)


python/ml4ir/base/pipeline.py
```diff
@@ -231,24 +231,6 @@ def validate_args(self):
 
         return self
 
-    def finish(self):
```
balikasg, line 234: this version of `finish` is not used anywhere, correct me if I a wrong. Lower in the same file there is another definition of it. Removing it to clean code.

python/ml4ir/base/pipeline.py
```diff
@@ -413,13 +394,15 @@ def run(self):
                 )
 
             # Finish
-            self.finish(job_status, job_info)
+            # self.finish(job_status, job_info)
```
balikasg, line 397: I am planning to remove this call to finish too, if you are OK @lastmansleeping. In the same function, we call `finish` again lower so I did not find the purpose of doing it several times and overwriting files.

python/ml4ir/base/pipeline.py
```diff
@@ -413,13 +394,15 @@ def run(self):
...
 
         except Exception as e:
             self.logger.error("!!! Error Training Model: !!!\n{}".format(str(e)))
             traceback.print_exc()
             job_status = "_FAILURE"
             job_info = "{}\n{}".format(str(e), traceback.format_exc())
+            self.finish(job_status, job_info)
+            return self
```
balikasg, line 405: the code should exit here by calling finish to populate the 'FAILURE' file with the exception. Otherwise, it continues and tries to calculate scores.

python/ml4ir/base/pipeline.py
```diff
@@ -471,8 +454,8 @@ def finish(self, job_status, job_info):
 
         if self.args.file_handler == FileHandlerKey.SPARK:
             # Copy logs and models to HDFS
-            self.file_io.copy_to_hdfs(self.models_dir_local, self.models_dir, overwrite=True)
-            self.file_io.copy_to_hdfs(self.logs_dir_local, self.logs_dir, overwrite=True)
+            self.file_io.copy_to_hdfs(self.models_dir_local, self.models_dir, overwrite=False)
```
balikasg, line 457: Do we want to overwrite the files by default?

python/ml4ir/applications/classification/pipeline.py
```diff
@@ -168,6 +170,33 @@ def get_relevance_dataset(
...
+        super().finish(job_status, job_info)
+
+        # Create model_bundle.zip
+        if self.args.file_handler == FileHandlerKey.SPARK:
+            # Copy tfrecord files in model_files
+            os.makedirs("epmm")
+            shutil.copytree(os.path.join(self.models_dir_local, "final/tfrecord/"), "model_files")
+            shutil.make_archive("./epmm/model_bundle", "zip", "model_files")
+
+            (self.file_io
```
balikasg, line 184: while this solution is good for the actual problem at hand (let's call it a v0), I am not comfortable with selecting the vocabularies to put in the bundle. I think it would be preferable to copy all vocabulary files.

python/ml4ir/applications/classification/pipeline.py
```diff
@@ -168,6 +170,33 @@ def get_relevance_dataset(
...
         return relevance_dataset
 
+    def finish(self, job_status, job_info):
+        # This will write the _SUCCESS/_FAILURE files
+        super().finish(job_status, job_info)
+
+        # Create model_bundle.zip
+        if self.args.file_handler == FileHandlerKey.SPARK:
+            # Copy tfrecord files in model_files
+            os.makedirs("epmm")
```
arvindsrikantan, line 180: Let's name this something more generic?

python/ml4ir/base/pipeline.py
arvindsrikantan, line 405: I suggest reverting these two lines. The metrics are not being computed. It'll just create an empty dict.
```suggestion
```

python/ml4ir/base/pipeline.py
arvindsrikantan: Pros and cons on both ends. Overwriting is risky as we might lose important previous run. Not doing so, we risk discarding the whole run. Best middle ground would be to create a temporary path in HDFS, write it there and communicate the path to the user

For now, I agree we should stick with `overwrite=False`

arvindsrikantan (changes requested):
One mandatory correction we will need is to copy the entity_id csv file without compressing. Looks good otherwise.

python/ml4ir/applications/classification/pipeline.py
jakemannix: yes please.  Probably have it configured by the commandline. 

("epmm" means nothing in the open source / outside of Salesforce usage of this codebase)

python/ml4ir/applications/classification/pipeline.py
```diff
@@ -168,6 +170,33 @@ def get_relevance_dataset(
...
+            shutil.copytree(os.path.join(self.models_dir_local, "final/tfrecord/"), "model_files")
+            shutil.make_archive("./epmm/model_bundle", "zip", "model_files")
+
+            (self.file_io
+             .read_df(self.feature_config.get_label()['feature_layer_info']['args']['vocabulary_file'])
+             .to_csv("key_prefix_labels.csv"))
+            shutil.make_archive("./epmm/key_prefix_labels.csv", "zip", ".", "key_prefix_labels.csv")
```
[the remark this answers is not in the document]
jakemannix: there should be no explicit mention of "key_prefix_labels" in this library - that's a name/choice for a company-internal specific thing.

We can have files which are specified in the feature_config all be packed in somewhere, but this shouldn't be hard coded, for sure.

python/ml4ir/applications/classification/pipeline.py
```diff
@@ -168,6 +170,33 @@ def get_relevance_dataset(
...
+
+        # Create org_id_to_key_prefix.csv.zip
+            org_id = [x for x in self.feature_config.get_all_features() if x['name'] == 'organization_id'][0]
+            org_mapping = org_id['feature_layer_info']['args']['vocabulary_file']
```
jakemannix, line 191: what does "org_id" mean to an open source project like this?  Remember, this is *generic* "ml4ir" - the code in here cannot specify any SFDC-internal terminology, and file names need to come in via cmdline args or in the config files.

jakemannix (changes requested):
Please remove all references to company internal terminology, and hard-coded file names.  All should come through configuration (cmdline args and/or config files that don't live in this repo)

python/ml4ir/applications/classification/pipeline.py
balikasg: yes, I will do.

python/ml4ir/applications/classification/pipeline.py
```diff
@@ -168,6 +170,31 @@ def get_relevance_dataset(
 
         return relevance_dataset
 
+    def finish(self, job_status, job_info):
+        # This will write the _SUCCESS/_FAILURE files
+        super().finish(job_status, job_info)
+
+        # Create model_bundle.zip depending on user args
```
lastmansleeping, line 177: @balikasg Should this even be in ml4ir? Probably worth doing it internally? Also I'm interested in understanding why you need to save vocabulary files? I thought you were using the `VocabLookup` in the model itself?

CC: @jakemannix

python/ml4ir/base/features/feature_config.py
```diff
@@ -368,6 +368,40 @@ def get_all_features(self, key: str = None, include_label: bool = True):
                     if fdict["name"] != self.get_label("name")
                 ]
 
+    def get_all_values(self, key: str = None, include_label: bool = True):
```
lastmansleeping, line 371: @balikasg I'm trying to understand what is the need for this. Are you trying to use this get all the vocab files at once?

python/ml4ir/base/pipeline.py
lastmansleeping: Yeah. I think this is an artifact of rebase and merge. I'm fixing it in my follow up PR. But you can do it here as well.

python/ml4ir/base/pipeline.py
lastmansleeping: Yes! Again, bad rebase.

python/ml4ir/base/pipeline.py
lastmansleeping: Why do you not want to overwrite by default? I don't get the logic. Do you want your training jobs that have run for hours to fail because there existed something already?

lastmansleeping: I'm fine if we implement the temp path, but it will be a nightmare to track if we have 200 models training in parallel. I don't think we should remove the `overwrite` without an alternative.

python/ml4ir/applications/classification/pipeline.py
balikasg: @lastmansleeping , we need the labels at this point. The model predicts a list of ids, we want to be able to convert to the actual label and post process.

python/ml4ir/applications/classification/pipeline.py
```diff
@@ -168,6 +170,31 @@ def get_relevance_dataset(
...
+        if self.args.write_tf_model_dir:
+            os.makedirs(self.args.write_tf_model_dir)
```
lastmansleeping, line 179: @balikasg I recommend moving all IO to `ml4ir/base/io/file_io` if possible and only keep the function call here.

python/ml4ir/applications/classification/pipeline.py
balikasg: We are planning to work on a post-processing function tf.function, that will create the format needed e.g., a list of ordered predictions with their probabilities. But we are not there yet.

lastmansleeping:
@balikasg Interested in knowing why you are adding functionality only to the classification pipeline and not the base pipeline itself.

python/ml4ir/base/features/feature_config.py
balikasg: yes. exactly.

python/ml4ir/applications/classification/pipeline.py
lastmansleeping: Why not do it in ml4ir itself?

python/ml4ir/applications/classification/pipeline.py
balikasg: I will do that.

python/ml4ir/applications/classification/pipeline.py
lastmansleeping: https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/base/model/serving.py#L128

@balikasg @jakemannix

balikasg:
"Interested in knowing why you are adding this functionality only to the classification pipeline and not the base pipeline itself" 
We did not sync offline, so I started with the classification pipeline. Having something that works now, I can easily move it to the pipeline.

python/ml4ir/base/config/parse_args.py
```diff
@@ -331,6 +331,27 @@ def define_args(self):
             help="[Ranklib format only] Option to keep additional info (All info after the '#' in the format [key = val]). 1 to keep, 0 to ignore",
         )
 
+        self.add_argument(
+            "--write_tf_model_dir",
```
lastmansleeping, line 335: @balikasg Can we maybe add "bundle" or "rich_model_bundle" something like that to differentiate?

python/ml4ir/base/config/parse_args.py
```diff
@@ -331,6 +331,27 @@ def define_args(self):
...
+            help="When specified zips the Tensorflow Record Signature of the model"
+                 " to model_bundle.zip and writes it to the specified directory.",
+        )
+
+        self.add_argument(
+            "--write_config_csvs_by_name",
+            type=str,
+            nargs='+',
+            required=False,
+            default='vocabulary_file',
```
lastmansleeping, line 348:
```suggestion
            default="vocabulary_file",
```

python/ml4ir/base/config/parse_args.py
```diff
@@ -331,6 +331,27 @@ def define_args(self):
...
+            default=".",
+            required=False,
+            help="When specified zips the Tensorflow Record Signature of the model"
+                 " to model_bundle.zip and writes it to the specified directory.",
+        )
+
+        self.add_argument(
+            "--write_config_csvs_by_name",
+            type=str,
+            nargs='+',
```
lastmansleeping, line 346:
```suggestion
            nargs="+",
```

python/ml4ir/base/pipeline.py
```diff
@@ -474,6 +472,26 @@ def finish(self, job_status, job_info):
             self.file_io.copy_to_hdfs(self.models_dir_local, self.models_dir, overwrite=True)
             self.file_io.copy_to_hdfs(self.logs_dir_local, self.logs_dir, overwrite=True)
 
+        if self.args.write_tf_model_dir:
```
mbrette, line 475: Could be a good idea to move all that to a separate method (rules of thumb for me: a function should do a single thing).

lastmansleeping (changes requested):
@balikasg (Adding here for visibility and completeness, I know we've discussed some of this offline)

_I can approve once these questions are answered:_

1. Why are we adding this logic to ml4ir and not something internal?
2. Why are we adding this logic at all when we can use ml4ir `postprocessing_fn` hook in the serving signature to convert scores back to labels?
3. If the models are already trained, do we even need to add this idiosyncratic (to Salesforce) model bundling/saving logic to ml4ir? Do you expect it to be used by others?
4. [Already addressed] If we go ahead with adding it, it should be on the base pipeline IMO as this does not uniquely apply to classification only per se.

CC: @mbrette @arvindsrikantan

balikasg:
@lastmansleeping, as I said offline:

Why are we adding this logic to ml4ir and not something internal?
- Right now the training process is built around ml4ir. We do not differentiate between internal/external versions of the project. I did the most of what I could think to make things generic as it is now controlled by arguments passed. The pipelines are e2e mechanisms, from data to models. 

Why are we adding this logic at all when we can use ml4ir postprocessing_fn hook in the serving signature to convert scores back to labels?
- Right now it requires more and non-trivial work to do this. I do not say we will not do this in the future. I also think it is fair to be able to save the model and the vocabulary files you used to get this model, even for testing and analysis purposes. This PR lets you do this in an easier way. In the analysis I have done with the models, I needed these files. In the future, perhaps this need will cease to exist, we can consider removing it. 

If the models are already trained, do we even need to add this idiosyncratic (to Salesforce) model bundling/saving logic to ml4ir? Do you expect it to be used by others?
- We trained some models. All for a particular org. I do not claim we will not train models in the coming weeks etc. I expect that anyone who is deploying a model can use this output. 

[Already addressed] If we go ahead with adding it, it should be on the base pipeline IMO as this does not uniquely apply to classification only per se.
- This is now done.

lastmansleeping (approved):
@balikasg Looks good for now. Approving with the caveat that this is a temporary way to unblock training and deploying models. We can move it to something internal soon.

jakemannix:
> @lastmansleeping, as I said offline:
> 
> Why are we adding this logic to ml4ir and not something internal?
> 
> * Right now the training process is built around ml4ir. We do not differentiate between internal/external versions of the project. I did the most of what I could think to make things generic as it is now controlled by arguments passed. The pipelines are e2e mechanisms, from data to models.

Just a note: there's not a plan to have "an internal version" of the the project.  It's "using ml4ir as a library": we have _code_ inside the company which calls into the ml4ir code to run training.  This is what we mean by "migrating the pipeline.py to internal".  Does that make sense?

balikasg:
I am closing this PR. We now have an internal solution and we can avoid adding this functionality.

PR #83, opened 2020-10-16
let argument checks happen in argument parsing step and not in a custom validation step

balikasg:
Today, in the pipelines we have a custom `validate_args` function. This, for example, does not allow as to set `default=None` and there are workarounds setting empty strings, etc. This PR tries to fix this.


python/ml4ir/base/config/parse_args.py
```diff
@@ -130,16 +138,17 @@ def define_args(self):
 
         self.add_argument(
             "--compute_intermediate_stats",
-            type=bool,
+            type=ast.literal_eval,
```
balikasg, line 141: See here: https://stackoverflow.com/questions/15008758/parsing-boolean-values-with-argparse

python/ml4ir/base/pipeline.py
```diff
@@ -188,48 +175,6 @@ def set_seeds(self, reset_graph=True):
         tf.random.set_seed(self.args.random_state)
         random.seed(self.args.random_state)
 
-    def validate_args(self):
-        """
-        Validate the arguments to be used with RelevancePipeline
-        """
-        unset_arguments = {key: value for (key, value) in vars(self.args).items() if value is None}
```
balikasg, line 195: we should be able to have None in arguments. This blocks this. Now every check is enforced in the argparser. Another benefit is we get more detailed help messages that are dynamic (we add something to the metrics, the metrics will directly appear).

python/ml4ir/base/config/parse_args.py
```diff
@@ -21,30 +25,31 @@ def define_args(self):
...
             type=str,
-            default=None,
             help="Path to YAML file or YAML string with feature metadata for training.",
         )
 
         self.add_argument(
             "--model_file",
             type=str,
-            default="",
+            default=None,
```
jakemannix, line 52: should we have "required" args, that don't allow empty string or None? 

cc: @lastmansleeping

balikasg: setting this to `not-required` @jakemannix as in the pipeline docstring: 
```
model_file : str, optional
            Path to pretrained model file to be loaded for evaluation or retraining
```

balikasg: @jakemannix if you have some time, please take a look at this PR.

python/ml4ir/base/config/parse_args.py
```diff
@@ -196,13 +211,14 @@ def define_args(self):
...
             type=str,
+            choices=ServingSignatureKey.get_all_keys(),
             default="serving_default",
             help="SavedModel signature to be used for inference",
         )
 
         self.add_argument(
             "--use_part_files",
-            type=bool,
+            type=ast.literal_eval,
```
jakemannix, line 221: does this require True/False or will it work with lowercase as well?

jakemannix (dismissed):
Booleans and None are better than than int's and empty strings.  Looks good to me!

jakemannix:
> Booleans and None are better than than int's and empty strings. Looks good to me!

but if you have any concerns @lastmansleeping , raise them before @balikasg merges ;)

We probably need checks against commandline usage in the CI... :\

python/ml4ir/base/config/parse_args.py
```diff
@@ -59,24 +65,25 @@ def define_args(self):
         self.add_argument(
             "--loss_key",
             type=str,
-            default=None,
-            help="Loss to optimize. Has to be one of the losses in LossKey under ml4ir/config/keys.py",
```
lastmansleeping, line 63: Still worth having the comment, no? @balikasg

python/ml4ir/base/config/parse_args.py
```diff
@@ -59,24 +65,25 @@ def define_args(self):
...
+            choices=RankingLoss.get_all_keys() + ClassificationLoss.get_all_keys(),
+            help="Loss to optimize."
         )
 
         self.add_argument(
             "--metrics_keys",
             type=str,
+            nargs="+",
```
lastmansleeping, line 75: @balikasg I don't think this will work on MLS/sip-pycli.

lastmansleeping (dismissed):
I have to block this merge until we have verified the `nargs="+"` works on MLS for single model and grid search. My guess is that the addition of spaces, will break it.

lastmansleeping:
@balikasg Can we merge/close this?

python/ml4ir/base/config/parse_args.py
arvindsrikantan: Works fine on MLS @lastmansleeping , see SearchAssistantModeling/configs/train_ner.yaml (line 13)

balikasg: I tested with EP too.

arvindsrikantan (approved):
Looks good!

lastmansleeping (approved):
Thanks for the work @balikasg!

PR #87, opened 2020-10-28
implementing the `evaluation` function for the classification pipeline

balikasg:
This PR implements the evaluation function for the classification pipeline.


python/ml4ir/applications/classification/model/metrics/metrics_impl.py
```diff
@@ -55,7 +55,7 @@ def __init__(
         self,
         feature_config: Optional[FeatureConfig] = None,
         metadata_features: Dict = {},
-        name="top_k_categorical_accuracy",
+        name="top_5_categorical_accuracy",
```
balikasg, line 58: a miss from an older PR.

python/ml4ir/applications/classification/model/metrics/metrics_impl.py
```diff
@@ -107,5 +107,5 @@ def update_state(self, y_true, y_pred, sample_weight=None):
         will throw error.
         """
         return super(Top5CategoricalAccuracy, self).update_state(
-            tf.squeeze(y_true, axis=1), tf.squeeze(y_pred, axis=1), sample_weight=sample_weight
+            tf.squeeze(y_true), tf.squeeze(y_pred), sample_weight=sample_weight
```
balikasg, line 110: tf.sqeeze will only remove dimensions of 1. this makes the code here robust, keeping the standard behaviour of the metrics (according to TF) as well as making it compatible with the extra dimension we have.

python/ml4ir/applications/classification/tests/data/configs/feature_config.yaml
```diff
@@ -63,7 +63,7 @@ features:
       - 1
       - 20
     log_at_inference: true
-    is_group_metric_key: true
+    is_group_metric_key: false
```
balikasg, line 66: it does not make sense to have `query_text` as a `is_group_metric_key`

python/ml4ir/applications/classification/tests/data/configs/feature_config.yaml
```diff
@@ -121,7 +115,7 @@ features:
...
```
balikasg, line 118: same as above.

python/ml4ir/applications/classification/tests/test_base.py
```diff
@@ -72,7 +75,29 @@ def setUp(
         # Setup logging
         outfile: str = os.path.join(self.args.logs_dir, "output_log.csv")
 
-        self.logger = setup_logging(reset=True, file_name=outfile, log_to_file=True)
+        self.logger = setup_logging(reset=True,
+                                    file_name=outfile,
+                                    log_to_file=True)
+        self.run_default_pipeline(data_format="csv")
```
balikasg, line 81: I preferred to run the training pipeline as a part of `setUp`. This way, it is run only once and then in the `test_classification_model.py` we can just evaluate the training results.

python/ml4ir/applications/classification/tests/test_base.py
```diff
@@ -72,7 +75,29 @@ def setUp(
...
+
+    def run_default_pipeline(self, data_format: str):
```
balikasg, line 83: simply copied this function from the other test file.

python/ml4ir/base/model/relevance_model.py
```diff
@@ -416,14 +418,18 @@ def fit(
                     to differentiate from validation and test metrics
                     in the final experiment results
                     """
-                    train_metrics["train_{}".format(metric)] = value[0]
+                    # History is a dict of key: list(values per epoch)
+                    # We are capturing the metrics of the last epoch (-1)
```
balikasg, line 422: `history.history` looks like 
```
{"categorical_accuracy": [1.80, 1.65, 1.60, ...]} 
```
where the dim of the values is the number of epochs. The previous version uses the value of the first epoch. I find this misleading when reporting results. I am changing the behavior to get the info of the last epoch. 

To be honest, I would prefer returning the actual history here so that the user has the complete information. `train_metrics` is a subset of information and lots of interesting insights are missing.

python/ml4ir/applications/classification/pipeline.py
```diff
@@ -164,6 +166,102 @@ def get_relevance_dataset(
...
+                        metric.update_state(group[label_name].values.tolist(),
+                                            group[output_name].values.tolist())
+                        grouped_metrics.append({"group_name": group_['name'],
+                                                "group_key": name,
+                                                "metric": metric.name,
+                                                "value": metric.result().numpy(),
+                                                "size": group.shape[0]})
+        global_metrics = pd.DataFrame(global_metrics)
+        grouped_metrics = pd.DataFrame(grouped_metrics)
+        if logs_dir:
```
balikasg, line 254: results (`global_metrics` and `group_metrics`) are written to the logs_dir as CSV fles.
cc: @Ullimague

python/ml4ir/applications/classification/pipeline.py
```diff
@@ -164,6 +166,102 @@ def get_relevance_dataset(
...
 
+    def evaluate(
+        self,
+        test_dataset: data.TFRecordDataset,
+        inference_signature: str = None,
+        additional_features: dict = {},
+        group_metrics_min_queries: int = 50,
+        logs_dir: Optional[str] = None,
+        logging_frequency: int = 25,
+    ):
```
balikasg, line 177: I kept the same signature with the ranking and the base version of the function.

Ullimague (commented):
Looks good to me, just 2 questions:
- Is the model evaluated on each epoch? Or only the last one?
- What about creating a test where we use another column (not trainable) to base the metric grouping on?

Ullimague:
Also, do we explain somewhere in any readme how to select a column to compute the grouped metrics on?

balikasg:
- Is the model evaluated on each epoch? Or only the last one?
Only after finishing training, once. This is how ml4ir orchestrates things. 
- What about creating a test where we use another column (not trainable) to base the metric grouping on?
Yes, done. 
- Also, do we explain somewhere in any readme how to select a column to compute the grouped metrics on?
https://ml4ir.readthedocs.io/en/latest/quickstart/feature_config.html 
I think it is sufficient for the moment.

python/ml4ir/base/model/relevance_model.py
```diff
@@ -407,6 +408,7 @@ def fit(
             )
 
             # Write metrics for experiment tracking
+            # Returns
```
Ullimague, line 411: Returns + ?

python/ml4ir/applications/classification/tests/generate_test_data.py
```diff
@@ -30,6 +30,7 @@
 ]
 VOCABULARY_FEATURE_DOMAIN_ID = [char for char in string.ascii_lowercase.upper()]
 VOCABULARY_FEATURE_ENTITY = ["AAA", "BBB", "CCC", "DDD", "EEE", "FFF", "GGG", "HHH"]
+VOCABULARY_FEATURE_CODE = ["local", "production", "canary"]
```
Ullimague, line 33: Thanks for adding this, could we comment it
 VOCABULARY_FEATURE_CODE = ... # untrainable feature, used to compute grouped metric based on this category 
or something like that to provide context?

Ullimague (commented):
Thanks for the follow-up @balikasg! Only decorative suggestions added.

.circleci/config.yml
```diff
@@ -53,16 +53,48 @@ jobs:
           command: |
             . venv/bin/activate
             cd python
-            python3 -m pytest --html=pytest_report.html
-          no_output_timeout: 30m  # Time for running the tests before we timeout
+            python3 -m pytest ml4ir/applications/classification/tests --html=pytest_report.html
+          no_output_timeout: 45m  # Time for running the tests before we timeout
```
Ullimague, line 57: Ouch, we can pass +30m for the tests? Could be interesting to see which test run too long. cc @lastmansleeping

balikasg: In some cases I went up to ~25. Moved this to ensure it is not a blocker.

Ullimague (approved):
Thanks @balikasg!

.circleci/config.yml
GhislainBrun: @balikasg @Ullimague I don't think we should do that. Let's identify the tests taking too long and fix them. Tests should take a handful of minutes.

python/ml4ir/applications/classification/tests/test_base.py
arvindsrikantan: Too late to the party, but `setUp` and `tearDown` run before and after **each test** in the class. We need to change this to [setUpClass](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass) for it to run only once. cc: @lastmansleeping @jakemannix 

I'll try this locally to check if it helps with the memory issue and create a PR

PR #89, opened 2020-10-30
Cyclic lr

mohazahran:
Adding unit tests for cyclic lr in training
Adding unit tests for constant and exp learning rate
Using tf optimizer get function
Grouping default values for optimizer and lr schedule into one class.


python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -103,12 +103,15 @@ def test_cyclic_exp_lr(self):
...
+        gold = [0.10000000149011612,0.09486832469701767,0.08999999612569809,0.08538150042295456,0.08099999278783798,
+                0.07684334367513657,0.07289998978376389,0.06915900856256485,0.06560999155044556,0.062243103981018066,
+                0.05904899165034294,0.05601879581809044,0.05314409360289574,0.050416912883520126,0.04782968387007713,
+                0.045375220477581024,0.04304671287536621,0.04083769768476486,0.03874203935265541,0.036753926426172256,
+                0.03486783429980278,0.03307853266596794,0.03138105198740959,0.029770677909255028,0.028242945671081543]
+
         lrs = ExponentialDecay(
             initial_learning_rate=0.1,
-            decay_steps=10000000,
+            decay_steps=2,
```
balikasg, line 114: Nice!

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -0,0 +1,217 @@
+import sys; sys.path.insert(0, '/Users/mohamed.m/Documents/work/projects/ml4ir/python/')
```
balikasg, line 1: This is a red flag.. 
I am assuming it also means that you have not properly installed ml4ir in your local? You should try to follow the README, if it fails we should sync.

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -0,0 +1,217 @@
...
+        """Test a triangular cyclic learning rate"""
+        gold = [0.1,0.28000003,0.45999995,0.64000005,0.81999993,1.0,0.81999993,0.64000005,0.45999995,0.28000003,
+                0.1,0.28000003,0.46000007,0.6399999,0.81999993,1.0,0.81999993,0.6399999,0.46000007,0.28000003,
+                0.1,0.27999982,0.46000007,0.6399999,0.8200002]
+        lrs = cyclic_learning_rate.TriangularCyclicalLearningRate(
+            initial_learning_rate=0.1,
+            maximal_learning_rate=1.0,
+            step_size=5,
+        )
+        for i in range(25):
```
balikasg, line 69: Consider using [`np.isclose`](https://het.as.utexas.edu/HET/Software/Numpy/reference/generated/numpy.isclose.html). It offers tolerance and array comparison out of the box. 

If you want to get it a step further, you will notice that the loop is repeated in every test below. You can abstract it in a function: 
```python
def compare_lr_values(scheduler, expected_values):
    """Expects a learning rate scheduler `scheduler` and 
    a list of values. Compares ...."""
    actual_values = [scheduler[i].numpy() for i in range(len(expeceted_values))]
    return np.all(np.isclose(actual_values, expected_values, rtol=0.001)) 
```

Then in the parent function you can `self.assertTrue(compare_lr_values(scheduler, expected_values))`
You can adapt it to have nice messages etc but this is the idea.

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -0,0 +1,217 @@
...
+    def setUp(self):
+        self.feature_config_yaml_convert_to_clicks = INPUT_DIR + \
+            'ranklib_test_data/feature_config_convert_to_clicks.yaml'
+        self.model_config_file = MODEL_CONFIG
+
+    def parse_config(self, tfrecord_type: str, feature_config, io) -> SequenceExampleFeatureConfig:
+        if feature_config.endswith(".yaml"):
+            feature_config = io.read_yaml(feature_config)
+        else:
+            feature_config = yaml.safe_load(feature_config)
```
balikasg, line 55: when do you expect this part of the `if/else` to be used (since fileio below is the LocalIO by definition?

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -0,0 +1,217 @@
...
+    def __init__(self, lr_list):
+        self.lr_list = lr_list
+
+    def on_train_batch_begin(self, batch, logs=None):
+        self.lr_list.append(self.model.optimizer._decayed_lr(tf.float32).numpy())
+
+    def on_train_batch_end(self, batch, logs=None):
+        pass
+
+class TestCyclicLr(unittest.TestCase):
```
balikasg, line 44: I see two types of tests in this class. The first, generically tests the code. The second, is more an integration type test (not exactly, but resembles). Why not split them in two separate classes?

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -0,0 +1,217 @@
...
+        callbacks_list.append(my_calback_object)
+
+        history = relevance_model.model.fit(
+            x=dataset.train,
+            validation_data=dataset.validation,
+            epochs=2,
+            verbose=True,
+            callbacks=callbacks_list,
+        )
+        lr_gold = [0.001,0.020800006,0.040599994,0.0604,0.080199994,0.1,0.080199994,0.0604,0.040599994,0.020800006,0.001,
```
balikasg, line 203: Something to keep in mind is that our tests are taking a lot of time. It is good to have tests of course. Do we need 2 epochs here? Are the data small enough? I am asking myself such questions too..

balikasg: Another question would be if we can avoid initializing the whole model stuff, and just defined the optimizer (in line 179) and be able to get its values directly without passing from fit (I get there are pros and cons everywhere).

python/ml4ir/base/model/optimizers/optimizer.py
```diff
@@ -31,28 +44,8 @@ def choose_optimizer(model_config, learning_rate_schedule):
...
-                learning_rate=learning_rate_schedule, clipvalue=gradient_clip_value if 'gradient_clip_value' in model_config['optimizer'] else 5.0
-            )
-        elif optimizer_key == OptimizerKey.RMS_PROP:
-            return tf_optimizers.RMSprop(
-                learning_rate=learning_rate_schedule, clipvalue=gradient_clip_value if 'gradient_clip_value' in model_config['optimizer'] else 5.0
-            )
-        else:
-            raise ValueError("Unsupported Optimizer: " + optimizer_key)
+        config = {'learning_rate':learning_rate_schedule, 'clipvalue': gradient_clip_value if 'gradient_clip_value' in model_config['optimizer'] else DefaultValues.GRADIENT_CLIP_VALUE}
+        return tf.keras.optimizers.get({'class_name':optimizer_key, 'config':config})
```
balikasg, line 48: Nice! I tried to do the same [in this PR](https://github.com/salesforce/ml4ir/pull/65); I like your solution too (i am closing the other). A shortcoming though is that there is manual care for two things (gradient_clip_val and learning_rate). Also, even if the user does not ask for `clipvalue` we apply it under the hood! We should re-think this in another PR.

python/ml4ir/base/model/optimizers/optimizer.py
```diff
@@ -2,6 +2,19 @@
...
+import tensorflow as tf
+
+class DefaultValues(object):
+  """Default values for unspecified parameters of the optimizer and learning rate schedule."""
+  GRADIENT_CLIP_VALUE = 5.0
+  CONSTANT_LR = 0.01
+  EXP_DECAY_STEPS = 100000
+  EXP_DECAY_RATE = 0.96
+  CYCLIC_INITIAL_LEARNING_RATE = 0.001
+  CYCLIC_MAXIMAL_LEARNING_RATE = 0.01
```
balikasg, line 14: why do we need default values for all these? I appreciate the work, but why not simply fail the code when they are not specified? We should not be doing this under-the-hood. There should be a minimal amount of values that the user specifies. IMO, `initial` and `maximal_lr` should belong in this family.

python/ml4ir/base/model/optimizers/optimizer.py
```diff
@@ -2,6 +2,19 @@
 from tensorflow.keras.optimizers.schedules import ExponentialDecay
 from ml4ir.base.model.optimizers import cyclic_learning_rate
 from ml4ir.base.config.keys import OptimizerKey, LearningRateScheduleKey, CyclicLearningRateType
+import tensorflow as tf
+
+class DefaultValues(object):
```
balikasg, line 7: Your class-string (aka docstring) is very good! Consider propagating part of the info in the class name e.g., `OptimizerDefaultValues`

python/ml4ir/base/model/optimizers/optimizer.py
```diff
@@ -90,15 +83,15 @@ def choose_scheduler(model_config):
 
         if lr_schedule_key == LearningRateScheduleKey.EXPONENTIAL:
             learning_rate_schedule = ExponentialDecay(
-                initial_learning_rate=lr_schedule['learning_rate'] if 'learning_rate' in lr_schedule else 0.01,
-                decay_steps=lr_schedule['learning_rate_decay_steps'] if 'learning_rate_decay_steps' in lr_schedule else 100000,
-                decay_rate=lr_schedule['learning_rate_decay'] if 'learning_rate_decay' in lr_schedule else 0.96,
+                initial_learning_rate=lr_schedule['learning_rate'] if 'learning_rate' in lr_schedule else DefaultValues.CONSTANT_LR,
```
balikasg, line 86: Consider using the `get` method of a dictionary: 
```python
initial_learning_rate=lr_schedule.get('learning_rate', DefaultValues.CONSTANT_LR)
```
Same for the usecases below.

balikasg (changes requested):
Some changes requested.. Perhaps the PR was WIP but I jumped on it. 
A general comment (I have made this to @lastmansleeping too). There are benefits is smaller PRs. This PR (i) adds tests (ii) add defaults on optimizers. It could be two PRs ;-) The advantage is easier reviews, less time to merge. But we are working on this as a team all of us are trying to improve.

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
mohazahran: Oh this is my bad. I included the wrong file. Nice catch.

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
mohazahran: I set to 2 epochs to verify the behavior of scheduler for more than one epoch. The dataset I'm used is small, so time shouldn't be an issue I think.
As for using only the optimizer with no real data, I'm not sure how this would work. With that in mind, I'm not experienced with tensorflow.

python/ml4ir/base/model/optimizers/optimizer.py
mohazahran: I agree. We can set the default gradient_clip value to be 'unbounded'.

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -0,0 +1,217 @@
...
+from tensorflow.keras.optimizers.schedules import ExponentialDecay
+
+warnings.filterwarnings("ignore")
+
+INPUT_DIR = "ml4ir/applications/ranking/tests/data/"
+MODEL_CONFIG = "ml4ir/applications/ranking/tests/data/config/model_config_cyclic_lr.yaml"
+KEEP_ADDITIONAL_INFO = 0
+NON_ZERO_FEATURES_ONLY = 0
+
+class lr_callback(keras.callbacks.Callback):
```
lastmansleeping, line 33: @mohazahran I think PEP8 recommends use of [CapWords case](https://www.python.org/dev/peps/pep-0008/#class-names) for classes and typically we name test classes like `TestLearningRate`

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -0,0 +1,217 @@
...
+    def __init__(self, lr_list):
+        self.lr_list = lr_list
+
+    def on_train_batch_begin(self, batch, logs=None):
+        self.lr_list.append(self.model.optimizer._decayed_lr(tf.float32).numpy())
```
lastmansleeping, line 38: @mohazahran Why are you using `_decayed_lr` here instead of `lr`?

lastmansleeping: Also, add a docstring for this method so that we know what is happening later.

lastmansleeping: and for the Class.

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
lastmansleeping, line 33: @mohazahran PEP8 recommends we use CapWords convention for class names -> https://www.python.org/dev/peps/pep-0008/#class-names

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -0,0 +1,217 @@
...
+            non_zero_features_only=NON_ZERO_FEATURES_ONLY,
+            max_sequence_size=319,
+        )
+
+        # Define interaction model
+        interaction_model: InteractionModel = UnivariateInteractionModel(
+            feature_config=feature_config,
+            feature_layer_keys_to_fns={},
+            tfrecord_type=TFRecordTypeKey.SEQUENCE_EXAMPLE,
+            max_sequence_size=319,
```
lastmansleeping, line 161: @mohazahran Why is the `max_sequence_size` 319 here?

lastmansleeping: Same question for above.

lastmansleeping: Just use `ml4ir/applications/ranking/test/data/tfrecord` and set it to default.

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
lastmansleeping: @balikasg Our tests are taking long because of save and load. I don't think training on this tiny dataset will add a lot of time. We can discuss how we want to solve the save/load tests. IMO those are the most critical tests. But I think reducing the number of parameters will solve this. Can you file a github issue so that we can work on it?

lastmansleeping: @mohazahran I doubt you can test the optimizer without real data. We can look into it, but I like what you have here.

lastmansleeping (changes requested):
@mohazahran This is going to be an annoying request for you. Can you 
- move the gold learning rate values to a separate JSON file so that the code is more readable?
- after doing the above, you can use the same gold values for both the unit tests and your training time callback tests
- add the same training time tests for all learning rate schedules perhaps?

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -0,0 +1,217 @@
...
+            tfrecord_type=TFRecordTypeKey.SEQUENCE_EXAMPLE,
+            scorer=scorer,
+            optimizer=optimizer,
+            model_file=None,
+            file_io=io,
+            logger=Logger,
+        )
+        callbacks_list = []
+        lr_list = []
+        my_calback_object = lr_callback(lr_list)
```
lastmansleeping, line 193: @mohazahran An alternative implementation to consider is have assert statements in your callback itself. Upto you. This looks fine as well. The one change I would recommend though is to fetch the value using a getter method from the callback object after training as opposed to passing a list by reference into the callback like here.

python/ml4ir/base/model/optimizers/optimizer.py
lastmansleeping: The reason I set a clipvalue by default is because I observed constant gradient explosion otherwise. We can remove the default if we all agree that we set it to some sensible value every time we train.

@mohazahran Thanks for this change. See! So clean now. :)

python/ml4ir/base/model/optimizers/optimizer.py
lastmansleeping: Came here to say this.

@mohazahran Think of it like this - you might have users doing `from optimizer import *`, which would import `DefaultValues` into the python context. Since `DefaultValues` is a very generic name, the user will have no idea what it really is. So as a rule of thumb, it is good to be as explicit as possible in class and function names. I am working on improving on this as well.

python/ml4ir/base/model/optimizers/optimizer.py
lastmansleeping: +1

@balikasg You are beating me to all the comments during my PTO. Haha.

python/ml4ir/base/pipeline.py
```diff
@@ -400,6 +400,9 @@ def run(self):
         experiment_tracking_dict.update(train_metrics)
         experiment_tracking_dict.update(test_metrics)
 
+        # Add optimizer and lr schedule
+        experiment_tracking_dict.update(relevance_model.model.optimizer.get_config())
```
lastmansleeping, line 404: @mohazahran Does this add both optimizer and learning rate schedule params? It will be useful to do that.

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
mohazahran: The sample dataset I use has a query with 319 docs

python/ml4ir/base/pipeline.py
mohazahran: Yes it adds both.

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
lastmansleeping: Exactly. Change it to the one I pointed above.

python/ml4ir/base/model/optimizers/optimizer.py
lastmansleeping: I think these are the tensorflow defaults @balikasg

lastmansleeping (approved):
@mohazahran Approving this. Looks good to me. @balikasg do you have any pending blockers?

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -0,0 +1,223 @@
...
+    """Testing different learning rate schedules"""
+
+    def setUp(self):
+        self.feature_config_yaml_convert_to_clicks = INPUT_DIR + \
+            'ranklib/feature_config_convert_to_clicks.yaml'
+        self.model_config_file = MODEL_CONFIG
+
+    def compare_lr_values(self, scheduler, expected_values):
+        """Expects a learning rate scheduler `scheduler` and
+        a list of values. Compares ...."""
```
balikasg, line 55: typo in `....`?

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -0,0 +1,223 @@
...
+        actual_values = [scheduler[i].numpy() for i in range(len(expected_values))]
+        return np.all(np.isclose(actual_values, expected_values, rtol=0.001))
```
balikasg, line 57: excellent!

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
balikasg: Also, the docstring is not accurate, it currently does not get a scheduler but the lrs of the scheduler I think.

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -0,0 +1,223 @@
...
+        """Test a triangular cyclic learning rate"""
+        gold = [0.1,0.28000003,0.45999995,0.64000005,0.81999993,1.0,0.81999993,0.64000005,0.45999995,0.28000003,
+                0.1,0.28000003,0.46000007,0.6399999,0.81999993,1.0,0.81999993,0.6399999,0.46000007,0.28000003,
+                0.1,0.27999982,0.46000007,0.6399999,0.8200002]
+        lrs = cyclic_learning_rate.TriangularCyclicalLearningRate(
+            initial_learning_rate=0.1,
+            maximal_learning_rate=1.0,
+            step_size=5,
+        )
+        scheduler = [lrs(i) for i in range(25)]
```
balikasg, line 77: this repeat in every test. You can move it in the `self.compare_lr_values`. If you do, you don't need the 25 (hard-coded now) you can infer it from the `len(gold)` directly.

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -0,0 +1,223 @@
...
+        lr_list = my_callback_object.get_lr_list()
+        lr_gold = [0.001,0.020800006,0.040599994,0.0604,0.080199994,0.1,0.080199994,0.0604,0.040599994,0.020800006,0.001,
+                   0.010900003,0.020800006,0.030699994,0.040599994,0.050499998,0.040599994,0.030699994,0.020800006,
+                   0.010900003,0.001,0.0059499955,0.010900003,0.015849996,0.020800006,0.02575,0.020800006,0.015849996,
+                   0.010900003,0.0059499955,0.001,0.0034749978,0.0059500015,0.008424998,0.010900003,0.013375,
+                   0.010900003,0.008424998,0.0059500015,0.0034749978,0.001,0.0022374988,0.0034749978,0.0047125025,
+                   0.0059500015,0.0071875,0.0059500015,0.0047125025]
+
+        for i in range(len(lr_list)):
+            assert np.isclose(lr_gold[i], lr_list[i])
```
balikasg, line 217:
```suggestion
np.all(np.isclose(lr_gold, lr_list))
```

python/ml4ir/base/model/optimizers/optimizer.py
```diff
@@ -90,15 +85,15 @@ def choose_scheduler(model_config):
 
         if lr_schedule_key == LearningRateScheduleKey.EXPONENTIAL:
             learning_rate_schedule = ExponentialDecay(
-                initial_learning_rate=lr_schedule['learning_rate'] if 'learning_rate' in lr_schedule else 0.01,
-                decay_steps=lr_schedule['learning_rate_decay_steps'] if 'learning_rate_decay_steps' in lr_schedule else 100000,
-                decay_rate=lr_schedule['learning_rate_decay'] if 'learning_rate_decay' in lr_schedule else 0.96,
+                initial_learning_rate=lr_schedule.get('learning_rate', OptimizerDefaultValues.CONSTANT_LR),
+                decay_steps=lr_schedule.get('learning_rate_decay_steps', OptimizerDefaultValues.EXP_DECAY_STEPS),
+                decay_rate=lr_schedule.get('learning_rate_decay', OptimizerDefaultValues.EXP_DECAY_RATE),
```
balikasg, line 90: thank you for all the effort here @mohazahran

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
lastmansleeping: Actually, I recommended we move all these values to a separate gold file to keep the code readable.

balikasg (approved):
Thank you for the hard work and the changes here! 
I have left some minor comments, but overall i think we should move forward. I also don;t want to block this is it is related to a story etc

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
balikasg: what you suggest is even better. A JSON or something that holds the expected values of its optimizer.

PR #90, opened 2020-11-03
evaluation of classification models is slow

balikasg:
when evaluating a classification model, there is a log message "2020-11-03 00:45:25.599392: W tensorflow/core/grappler/optimizers/implementation_selector.cc:310] Skipping optimization due to error while loading function libraries: Invalid argument: Functions '__inference_standard_lstm_326453' and '__inference_standard_lstm_326453_specialized_for_model_1_bidirectional_backward_lstm_StatefulPartitionedCall_at___inference__predict_score_328540' both implement 'lstm_2e98d781-3305-4503-8a74-a0e21edaacd6' but their signatures do not match." 
This seems to [be a known issue](https://github.com/tensorflow/tensorflow/issues/30263).  

This makes inference very slow.. Local tests found we are at least 15x slower than calling model's predict. This PR re-writes the classification `predict` function making it 15x faster.

### DIFF
--- a/python/ml4ir/applications/classification/model/classification_model.py
+++ b/python/ml4ir/applications/classification/model/classification_model.py
@@ -1,4 +1,5 @@
 import os
+import numpy as np
 import pandas as pd
 from tensorflow import data
 from ml4ir.base.model.relevance_model import RelevanceModelConstants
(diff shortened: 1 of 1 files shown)


Ullimague (approved):
Thanks for the follow-up @balikasg! I assume the existing tests are going through the added automatically.

PR #91, opened 2020-11-09
Changing the TFRecord parsing logic to accept array inputs @rev mohazahran@

lastmansleeping:
- Updating parsing function, default value shapes
- Updating tfrecord parsing unit tests set up in previous PR (other tests should be untouched)
- Updating ranking loss and metrics to handle new shape
- Updating feature layer to work with new shapes
- Updating predictions_helper to handle new scores shape
- Adding new unit tests

After this change:
- Context features or features for Example message will be of shape `[max_len]`
- Sequence features will be of shape `[sequence_size, max_len]`
This requires additional squeeze when computing loss and metrics because the labels are now of shape `[batch_size, sequence_size, 1]`


lastmansleeping:
@mohazahran Can I merge this?

python/ml4ir/applications/ranking/model/losses/listwise_losses.py
```diff
@@ -26,18 +26,28 @@ def get_loss_fn(self, **kwargs):
...
-            y_true = tf.expand_dims(tf.squeeze(y_true), axis=-1)
-            y_pred = tf.expand_dims(tf.squeeze(y_pred), axis=-1)
+            # Reshape the tensors so that we sum the losses from each record
+            y_true = tf.expand_dims(y_true, axis=-1)
+            y_pred = tf.expand_dims(y_pred, axis=-1)
 
+            # Scale the sum of losses down by number of queries in the batch
             return tf.math.divide(bce(y_true, y_pred), batch_size)
 
+            return bce(y_true, y_pred, sample_weight=mask)
```
balikasg, line 49: there seems to be 2 return statements here, same identatation, no if/else. Typo?

python/ml4ir/applications/ranking/model/losses/listwise_losses.py
```diff
@@ -59,20 +69,20 @@ def get_final_activation_op(self, output_name):
             Uses `mask` field to exclude padded records from contributing
             to the softmax activation
         """
-        softmax_op = layers.Softmax(axis=-1, name=output_name)
+        softmax_op = layers.Softmax(axis=-1)
```
balikasg, line 72: are we using this? there is a note that it is more stable, but then we use `tf.where(..)`

python/ml4ir/base/pipeline.py
```diff
@@ -381,28 +382,26 @@ def run(self):
...
-        experiment_tracking_dict.update(self.feature_config.get_hyperparameter_dict())
-
-        # Add train and test metrics
-        experiment_tracking_dict.update(train_metrics)
-        experiment_tracking_dict.update(test_metrics)
-
-        # Add optimizer and lr schedule
-        experiment_tracking_dict.update(relevance_model.model.optimizer.get_config())
-
         job_info = pd.DataFrame.from_dict(
```
balikasg, line 405: If there is an issue, `job_info` here overwrites the exception and put in its place an empty dict. Do you want to move it to line 398, in the try statement?

balikasg (commented):
I left a few comments. The last one on `job_info` is important imho

python/ml4ir/base/pipeline.py
lastmansleeping: Yeah. Let me do that.

python/ml4ir/applications/ranking/model/losses/listwise_losses.py
lastmansleeping: Good catch. Forgot the delete the older one below. I've been playing with a bunch of different ways to do this. Using the `sample_weight` argument is a little more difficult than I expected.

mohazahran (approved):
Looking good. All tests and runs passed

PR #92, opened 2020-11-11
Adding support for command line feature_config and model_config overrides @rev mohazahran@

lastmansleeping:
- Adding support for dynamic arguments for `feature_config` and `model_config`
- Changing pipeline to read feature_config and model_config dict instead of passing file to Scorer
- Adding new nested feature_config and model_config updating utility
- Updating docs
- Adding unit tests

You can now run override feature_config and model_config from CLI like below:
```
python3 ml4ir/applications/ranking/pipeline.py \
--run_id test \
--data_format tfrecord \
--loss_key rank_one_listnet \
--num_epochs 1 \
--feature_config.features.query_text.feature_layer_info.args.encoding_size 512 \
--model_config.layers.first_dense.units 1024
```

For lists, we currently support either integer indices like `model_config.layers.0.units` or match on `name` in subdict like `model_config.layers.first_dense.units`

@jakemannix I will add some tests and docs for this. Let me know your thoughts.


lastmansleeping:
@balikasg You might be interested in this change when training models. We don't have to maintain multiple feature config and model config files. We can just override values from command line. Check my top comment.

python/ml4ir/base/config/parse_args.py
```diff
@@ -299,6 +336,34 @@ def define_args(self):
     def set_default_args(self):
         pass
 
+    def parse_args(self, args: List[str]) -> Namespace:
+        """
+        Parse command line arguments passed a list of strings.
```
mohazahran, line 341: passed as a

mohazahran (commented):
I think it would be beneficial to add a unit test in which a dynamic arg does not exist in the base dict (i.e. adding new parameter not overriding one).

mohazahran (approved):

python/ml4ir/base/config/parse_args.py
```diff
@@ -1,10 +1,47 @@
...
+            Value to be added to dictionary
+        option_string: str
+            Name of command line argument the action is used with
+
+        Returns
+        -------
+        dict
+            Dictionary with the updated key value pair
+        """
+        old_custom_args = getattr(namespace, self.dest)
```
balikasg, line 33: Should we add a check here for the first part of the argument to be only `feature_config` or `model_config`? E.g., return an error when I pass `config.units  52`?

python/ml4ir/base/config/parse_args.py
```diff
@@ -299,6 +336,34 @@ def define_args(self):
     def set_default_args(self):
         pass
 
+    def parse_args(self, args: List[str]) -> Namespace:
+        """
+        Parse command line arguments passed as a list of strings.
+        Additionally, handles dynamic arguments for feature_config
+        and model_config with prefixes feature_config. and model_config.
```
balikasg, line 343: Ok, it is noted here. We do not currently check this I think.

python/ml4ir/base/config/parse_args.py
```diff
@@ -299,6 +336,34 @@ def define_args(self):
...
+            List of command line args to be parsed
+
+        Returns
+        -------
+        Namespace
+            Namespace object obtained by parsing the input
+        """
+        dynamic_args = self.parse_known_args(args)[1]
+
+        for i in range(int(len(dynamic_args) / 2)):
```
balikasg, line 358: this is not clear why we are doing it.. I mean the `range(int(len(dynamic_args) / 2))`
it took me time. 
You could: 
```
args, unknownargs = parser.parse_known_args(args)

# Unknown args, when exist, are a list that consists of pairs of key values. 
# args, unknownargs = parser.parse_known_args('-knowwn bar -unknown test'.split())
# returns unknownargs=['-unknown', 'test']
# Below, we iterate in pairs: 
for key, value in zip(dynamic_args[::2], dynamic_args[1::2]):
   ...
```

balikasg (commented):
I find this a very useful addition. I made a couple of comments about things that were not obvious to me when trying to understand the though process.

python/ml4ir/base/config/parse_args.py
lastmansleeping: I can throw an error here, yes. Let me add that to my current PR.

PR #93, opened 2020-11-12
Fixing OOM in the `evaluate` function of classification

balikasg:


python/ml4ir/applications/classification/model/classification_model.py
```diff
@@ -103,6 +93,56 @@ def evaluate(
...
+
+        Returns
+        -------
+        A dictionary with the metric scores for the given data.
+        """
+        label_name = self.feature_config.get_label()['name']
+        output_name = self.output_name
+        metric.reset_states()
+        for chunk in self.get_chunks_from_df(predictions, batch_size):
+            metric.update_state(tf.constant(chunk[label_name].values.tolist(), dtype=tf.float32),
```
balikasg, line 134: this is important, it gives a guge speedup as there is not tracing needed when using tensors.

python/ml4ir/applications/classification/model/classification_model.py
```diff
@@ -140,19 +180,31 @@ def predict(
             outfile = os.path.join(logs_dir, RelevanceModelConstants.MODEL_PREDICTIONS_CSV_FILE)
             # Delete file if it exists
             self.file_io.rm_file(outfile)
+        predictions_df = self._create_prediction_dataframe(logging_frequency,
+                                                           test_dataset)
+        predictions_ = np.squeeze(self.model.predict(test_dataset))
+        # Below, avoid doing predictions.tolist() as it explodes the memory
+        predictions_df[self.output_name] = [x for x in predictions_]
```
balikasg, line 187: this is also very important. `.tolist()` increases the memory a lot.. Some examples here: https://stackoverflow.com/questions/43122012/why-is-nparray-tolist-taking-this-much-space

python/ml4ir/applications/classification/tests/test_classification_model.py
```diff
@@ -8,9 +8,9 @@ class ClassificationModelTest(ClassificationTestBase):
     :func:`~ml4ir.applications.classification.pipeline.ClassificationPipeline.get_relevance_model`
     """
 
-    def test_csv_loss(self):
+    def test_csv_loss_metrics(self):
```
balikasg, line 11: I followed Arvind's suggestion here and made the 3 functions one, to avoid executing setUp thrice. It speeds up test ~40 sec.

python/ml4ir/applications/classification/model/classification_model.py
```diff
@@ -103,6 +93,56 @@ def evaluate(
             self.logger.info(f"Evaluation Results written at: {logs_dir}")
         return global_metrics, grouped_metrics, metrics_dict
 
+    @staticmethod
```
balikasg, line 96: I extracted these functions hopefully to improve readability

python/ml4ir/applications/classification/model/classification_model.py
```diff
@@ -66,27 +67,16 @@ def evaluate(
                                    additional_features=additional_features,
                                    logs_dir=None,
                                    logging_frequency=logging_frequency)
-
-        label_name = self.feature_config.get_label()['name']
-        output_name = self.output_name
         global_metrics = []  # group_name, metric, value
         grouped_metrics = []
+        batch_size = test_dataset._input_dataset._batch_size.numpy()  # Hacky way to get batch_size
```
Ullimague, line 72: What do you think about commenting why we need batch_size?

Ullimague: Probably as part of the doc in calculate_metric_on_batch()

python/ml4ir/applications/classification/model/classification_model.py
```diff
@@ -103,6 +93,56 @@ def evaluate(
             self.logger.info(f"Evaluation Results written at: {logs_dir}")
         return global_metrics, grouped_metrics, metrics_dict
 
+    @staticmethod
+    def get_chunks_from_df(dataframe, size):
+        """
+        Given a pd.DataFrame, creates batches of size `size`.
```
Ullimague, line 99: Should we say we do this for performances reasons?

python/ml4ir/applications/classification/model/classification_model.py
```diff
@@ -103,6 +93,56 @@ def evaluate(
...
+        """
+        return [dataframe.iloc[pos:pos + size] for pos in range(0, len(dataframe), size)]
```
Ullimague, line 101: Just out of curiosity (may be a dumbish question), would we benefit to use a generator for our use-case here?

python/ml4ir/applications/classification/tests/test_classification_model.py
arvindsrikantan: We should use [subtest](https://docs.python.org/3/library/unittest.html#unittest.TestCase.subTest) here. It'll run all tests in parallel and report it individually

python/ml4ir/applications/classification/model/classification_model.py
balikasg: yes. done.

python/ml4ir/applications/classification/model/classification_model.py
balikasg: I think we should use generators if possible. I have done the change.

python/ml4ir/applications/classification/model/classification_model.py
```diff
@@ -140,19 +180,31 @@ def predict(
...
 
-        predictions_df_list = list()
+    def _create_prediction_dataframe(self, logging_frequency, test_dataset):
+        """
+        Iterates through the test data and collects for each
+        data point the information to be logged.
+        """
+        batch_count = 0
+        predictions_df_list = []
         for (x, y) in test_dataset.take(-1):  # returns (x, y) tuples
```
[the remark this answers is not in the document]
balikasg: yes. It ensures you iterate though the whole dataset. See here: https://www.tensorflow.org/api_docs/python/tf/data/Dataset#take

python/ml4ir/applications/classification/model/classification_model.py
```diff
@@ -140,19 +180,31 @@ def predict(
...
             batch_predictions = pd.DataFrame({key: np.squeeze(x[key].numpy()).tolist() for key in x.keys()})
             batch_predictions[self.feature_config.get_label()["name"]] = np.squeeze(y.numpy()).tolist()
             predictions_df_list.append(batch_predictions)
+            if batch_count % logging_frequency == 0:
+                self.logger.info(f"Finished evaluating {batch_count} batches")
+            batch_count += 1
         predictions_df = pd.concat(predictions_df_list)
```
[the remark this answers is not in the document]
balikasg: yes, this is a bottleneck. But it is exactly this bottleneck that allows the pandas group_by that you commented above. I can not think of a simpler solution without sacrificing readability and requiring state-keeping of 100s of thousands of metrics per batch

python/ml4ir/applications/classification/model/classification_model.py
```diff
@@ -66,27 +67,16 @@ def evaluate(
...
         global_metrics = []  # group_name, metric, value
         grouped_metrics = []
+        batch_size = test_dataset._input_dataset._batch_size.numpy()  # Hacky way to get batch_size
         for metric in self.model.metrics:
-            metric.reset_states()
-            metric.update_state(predictions[label_name].values.tolist(), predictions[output_name].values.tolist())
-            global_metrics.append({"metric": metric.name,
-                                   "value": metric.result().numpy()})
+            global_metrics.append(self.calculate_metric_on_batch(metric, predictions, batch_size))
             for group_ in group_metrics_keys:
```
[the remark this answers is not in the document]
balikasg: these are good ideas. I will implement the change and the one that you cite subtest tomorrow.

python/ml4ir/applications/classification/model/classification_model.py
balikasg: `tolist` is more readable. But the comprehension maintains the inner arrays as np.array that saves memory.

python/ml4ir/applications/classification/model/classification_model.py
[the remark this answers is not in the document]
balikasg: I tried this. Changing metric calls that frequently results in tracing warning and a potential slowdown. For this reason I am leaving it as is.

python/ml4ir/applications/classification/tests/test_classification_model.py
balikasg: thanks for the hint, I was not aware of it though. It will have to be part of another PR though, I have alredy spent much time on this one.

python/ml4ir/applications/classification/model/classification_model.py
arvindsrikantan: You're right. I forgot that these are 2d lists

arvindsrikantan:
Removed the conflict after merging my branch. There is still a bug in argparser from the legacy code. I'll fix it tomorrow

arvindsrikantan:
> Removed the conflict after merging my branch. There is still a bug in argparser from the legacy code. I'll fix it tomorrow

...in a different PR

arvindsrikantan (approved):
I'm not sure if I have the "power" to approve anymore since part of my changes are in this PR. But I'm approving all changes introduced by Georgios

(Considering Georgios approved my changes, we should ideally be good to merge)

PR #94, opened 2020-11-16
Disabling group-wise metric computation to unblock e2e training

arvindsrikantan:


python/ml4ir/applications/classification/model/classification_model.py
```diff
@@ -62,52 +63,69 @@ def evaluate(
...
-                                                                              batch_size,
-                                                                              group_['name'], name))
-        global_metrics = pd.DataFrame(global_metrics)
-        grouped_metrics = pd.DataFrame(grouped_metrics).sort_values(by='size')
-        if logs_dir:
-            self.file_io.write_df(
-                grouped_metrics,
-                outfile=os.path.join(logs_dir, RelevanceModelConstants.GROUP_METRICS_CSV_FILE),
-                index=False
+        if compute_group_metrics_in_memory:
```
arvindsrikantan, line 75: Temporary patch till we decide on a plan for next quarter

python/ml4ir/applications/classification/tests/test_base.py
```diff
@@ -30,21 +30,22 @@ class ClassificationTestBase(unittest.TestCase):
     Setting default arguments and context for tests .../classification/tests folder.
     """
 
-    def setUp(
-        self,
+    @classmethod
+    def setUpClass(
```
arvindsrikantan, line 34: This should further speed up our tests. We won't need to split them up to run faster.

python/ml4ir/applications/classification/tests/test_classification_model.py
```diff
@@ -32,7 +33,8 @@ def test_csv_loss_metrics(self):
                         msg=f"Top5 Categorical_accuracy not in expected range."
                             f" Expected: {expected_acc} ± {tol}, Found: {self.metrics_dict[_metric]}")
 
-    def test_group_metrics_df(self):
+    @pytest.mark.skip(reason="Pending based on spike for EP evaluation")
```
arvindsrikantan, line 36: Pending planning for next quarter

python/ml4ir/applications/classification/tests/test_classification_model.py
```diff
@@ -32,7 +33,8 @@ def test_csv_loss_metrics(self):
...
+    def test_group_metrics_df(self):  # FIXME: Pending based on spike for EP evaluation
```
Ullimague, line 37: Note sure if we want to have sfdc notions like "spike for EP evaluation".

arvindsrikantan: Corrected it

python/ml4ir/applications/classification/model/classification_model.py
```diff
@@ -62,52 +63,69 @@ def evaluate(
...
         You can directly do a `model.evaluate()` only if the keras model is compiled
         """
+        # FIXME: We need to come to a consensus for computing in pandas or spark and how we plan to
+        #  do that. In the mean time we will stick with just the model evaluation metrics
+        compute_group_metrics_in_memory = False
+
         if not self.is_compiled:
             return NotImplementedError
         group_metrics_keys = self.feature_config.get_group_metrics_keys()
         metrics_dict = self.model.evaluate(test_dataset)
```
arvindsrikantan, line 73: Quick callout: This works fine for 7 days of data without issues (on a 40 gig container). It's just extremely slow as one would expect. (A potential solution if this is causing issues later is to increase the batch size for test significantly as inference is less memory intensive than training)

python/ml4ir/applications/classification/model/classification_model.py
balikasg: can we make this an argparser argument with default to `True` (for the huge majority of datasets I know, it would pass easily)?
Also, I would suggest adding just below: 
```
self.logger.info('Computing metric scores for the grouped metrics in the feature_config. Warning: currently, this collects the test data and predictions in memory in order to perform the groupBy operations. With very big data, it can lead in OOM issues.')
```
What do you think?

python/ml4ir/applications/classification/tests/test_classification_model.py
```diff
@@ -32,7 +33,8 @@ def test_csv_loss_metrics(self):
                         msg=f"Top5 Categorical_accuracy not in expected range."
                             f" Expected: {expected_acc} ± {tol}, Found: {self.metrics_dict[_metric]}")
 
-    def test_group_metrics_df(self):
+    @pytest.mark.skip(reason="Pending on planning for classification model evaluation")
```
balikasg, line 36: if we set the argparser arg to True, or when we have a way to control it, can this test pass?

python/ml4ir/applications/classification/model/classification_model.py
balikasg: yes. Also we are talking about 40M data points here.. I am not too in favour of increasing the batch size under the hood.
how much time does it take?

python/ml4ir/applications/classification/tests/test_classification_model.py
arvindsrikantan: Yes

python/ml4ir/applications/classification/model/classification_model.py
arvindsrikantan: I didn't mean to say we should change it under the hood. Should be changed in the arguments. It takes well over 4 hours

python/ml4ir/applications/classification/model/classification_model.py
arvindsrikantan: Makes sense. Made the changes 👍

python/ml4ir/applications/classification/model/classification_model.py
```diff
@@ -163,12 +187,12 @@ def calculate_metric_on_batch(self, metric, predictions, batch_size, group_name=
                 "size": predictions.shape[0]}
 
     def predict(
-        self,
-        test_dataset: data.TFRecordDataset,
-        inference_signature: str = "serving_default",
-        additional_features: dict = {},
-        logs_dir: Optional[str] = None,
-        logging_frequency: int = 25,
+            self,
```
arvindsrikantan, line 190: Pycharm is auto formatting this for me. Do you guys have different configuration by any chance for auto-formatting?

arvindsrikantan:
@balikasg , I made the suggested updates. Do take a look

balikasg (approved):
LGTM!

PR #95, opened 2020-11-18
Adding global 1D pooling @rev mohazahran@

lastmansleeping:
This PR adds a global 1D pooling/aggregation function to convert a list of values such as `[1., 2., 3.]` -> `[6., 2., 3., 1., 3.]`, which are the sum, mean, min, max respectively.


python/ml4ir/applications/ranking/tests/test_feature_layer.py
```diff
@@ -555,13 +599,98 @@ def test_categorical_indicator_with_vocabulary_file_without_ids(self):
...
+            feature_info={
+                "name": "f",
+                "feature_layer_info": {
+                    "args": {
+                        "fns": [
+                            "sum",
+                            "mean",
+                            "max",
+                            "min",
+                            "count_nonzero"
```
mohazahran, line 676: Just another possible test case maybe. If non of the pooling function is passed or only non-supported pooling functions are passed then the call to the pooling function would crash, no?

lastmansleeping: Yeah, what do you think we can do if no pooling function is passed? Ideally, that cannot be the case. But I can throw an appropriate error if that helps.

mohazahran (approved):

PR #98, opened 2020-11-27
Fix mismatch between name node_name in FeatureConfig mapping.

mbrette:
This mismatch led to different prediction between java/scala and python

### DIFF
--- a/jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/data/ModelFeaturesConfig.scala
+++ b/jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/data/ModelFeaturesConfig.scala
@@ -35,13 +35,14 @@ case class ModelFeaturesConfig(@JsonProperty("rank") initialRank: FeatureConfig,
 
 /**
   * At inference-time, ml4ir's FeatureConfig needs to know only the following properties per feature:
-  * @param nodeName the Tensorflow input graph node name to wire this feature into
+  * @param name the protobuf key name representing the input. Tensorflow will wire the value into the corresponding node
+  *             name
   * @param dTypeString string | int64 | float
   * @param servingConfig {@see ServingConfig} below
   * @param tfRecordType context | sequence (latter only for {@see SequenceExample} features)
   */
 @JsonIgnoreProperties(ignoreUnknown = true)
-case class FeatureConfig(@JsonProperty("node_name") nodeName: String,
+case class FeatureConfig(@JsonProperty("name") name: String,
                          @JsonProperty("dtype") dTypeString: String,
                          @JsonProperty("serving_info") servingConfig: ServingConfig,
                          @JsonProperty("tfrecord_type") tfRecordType: String) {
--- a/jvm/ml4ir-inference/src/test/resources/classification/feature_config_with_same_name.yaml
+++ b/jvm/ml4ir-inference/src/test/resources/classification/feature_config_with_same_name.yaml
@@ -10,7 +10,7 @@ query_key:
   serving_info:
     name: queryId
     required: false
-  default_value: 0
+    default_value: 0
   tfrecord_type: context
 label:
   name: entity_id
@@ -31,7 +31,7 @@ label:
   serving_info:
     name: entity_id
     required: false
-  default_value: ""
+    default_value: ""
   tfrecord_type: context
 features:
   - name: query_text
@@ -56,7 +56,7 @@ features:
     serving_info:
       name: query_text
       required: true
-    default_value: ""
+      default_value: ""
     tfrecord_type: context
   - name: query_text
     node_name: query_words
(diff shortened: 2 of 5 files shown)


jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/data/ModelFeaturesConfig.scala
```diff
@@ -35,13 +35,14 @@ case class ModelFeaturesConfig(@JsonProperty("rank") initialRank: FeatureConfig,
...
-  * @param nodeName the Tensorflow input graph node name to wire this feature into
+  * @param name the protobuf key name representing the input. Tensorflow will wire the value into the corresponding node
+  *             name
   * @param dTypeString string | int64 | float
   * @param servingConfig {@see ServingConfig} below
   * @param tfRecordType context | sequence (latter only for {@see SequenceExample} features)
   */
 @JsonIgnoreProperties(ignoreUnknown = true)
-case class FeatureConfig(@JsonProperty("node_name") nodeName: String,
+case class FeatureConfig(@JsonProperty("name") name: String,
```
balikasg, line 45: agreed. This raises the question if we need node_name and why. Same question for serving_info.name. 
Right now we have 3 names, I am wondering if we can live with 2.

balikasg (dismissed):
great!!

jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/data/ModelFeaturesConfig.scala
mbrette: I think we need node_name (it has to be unique, and it is visible in the graph, so useful for debug to have meaningful name).
But maybe we could get rid of serving_info.name.

mbrette: BTW, I double-checked that 'name' is not used as a key to a map of feature info. In the python code, the features are always manipulated as a list. (so 'name' does not have to be unique)

jvm/ml4ir-inference/src/test/resources/classification/feature_config_with_same_name.yaml
```diff
@@ -0,0 +1,155 @@
...
+        max_length: 20
+    preprocessing_info:
+      - fn: preprocess_text
+        args:
+          remove_punctuation: true
+          to_lower: true
+    serving_info:
+      name: query_text
+      required: true
+    default_value: ""
```
Ullimague, line 59: minor: default_value should be under serving_info.

jvm/ml4ir-inference/src/test/resources/classification/feature_config_with_same_name.yaml
```diff
@@ -0,0 +1,155 @@
...
+      fn: categorical_embedding_with_vocabulary_file
+      args:
+        vocabulary_file: ml4ir/applications/classification/tests/data/configs/vocabulary/domain_id.csv
+        embedding_size: 64
+        default_value: -1
+        num_oov_buckets: 1
+    serving_info:
+      name: domain_id
+      required: true
+    default_value: ""
```
Ullimague, line 119: minor: default_value should be under serving_info.

jvm/ml4ir-inference/src/test/resources/classification/feature_config_with_same_name.yaml
```diff
@@ -0,0 +1,155 @@
...
+      args:
+        vocabulary_file: ml4ir/applications/classification/tests/data/configs/vocabulary/entity_id.csv
+        embedding_size: 64
+        encoding_size: 64
+        num_oov_buckets: 1
+        max_length: 20
+    serving_info:
+      name: user_context
+      required: true
+    default_value: ""
```
Ullimague, line 154: minor: default_value should be under serving_info.

jvm/ml4ir-inference/src/test/resources/classification/feature_config_with_same_name.yaml
```diff
@@ -0,0 +1,155 @@
...
+    - fn: one_hot_vectorize_label
+  feature_layer_info:
+    type: numeric
+    fn: categorical_indicator_with_vocabulary_file
+    args:
+      vocabulary_file: ml4ir/applications/classification/tests/data/configs/vocabulary/entity_id.csv
+  serving_info:
+    name: entity_id
+    required: false
+  default_value: ""
```
Ullimague, line 34: minor: default_value should be under serving_info.

jakemannix: addressed in my update to this branch, @Ullimague !

Ullimague (approved):

PR #99, opened 2020-12-28
Division by zero fix for global_1d_pooling

lastmansleeping:


python/ml4ir/base/features/feature_fns/sequence.py
```diff
@@ -167,13 +163,14 @@ def global_1d_pooling(feature_tensor, feature_info, file_io: FileIO):
...
             if "padded_val" in args:
+                # NOTE: To avoid division by zero, we set 0 to a small value of 1e-10
                 pooled_tensors.append(
                     tf.math.divide(
                         tf.math.reduce_sum(masked_feature_tensor, axis=-1),
                         tf.math.reduce_sum(
                             tf.where(
                                 tf.equal(feature_tensor, padded_val_tensor),
-                                tf.constant(0, dtype=feature_tensor.dtype),
+                                tf.constant(1e-10, dtype=feature_tensor.dtype),
```
balikasg, line 173: Did this issue occur in a real scenario? Can you share a bit more? This will result from an output value of the layer that is not defined (inputs were all zero to reach this?) to something that will be perhaps quite big (division with small value). Perhaps this is what is done in general, I do not know. 
Also, do you think we would benefit from using the keras layers here (e.g., [this](https://github.com/tensorflow/tensorflow/blob/v2.4.0/tensorflow/python/keras/layers/pooling.py#L750) ) instead of writing our own ones (I understand this is not the scope of this PR)?

python/ml4ir/base/model/relevance_model.py
```diff
@@ -635,11 +635,14 @@ def save(
         # Save individual layer weights
         self.file_io.make_directory(os.path.join(model_file, "layers"), clear_dir=True)
         for layer in self.model.layers:
-            self.file_io.save_numpy_array(
-                np_array=layer.get_weights(),
-                file_path=os.path.join(model_file, "layers", "{}.npz".format(layer.name)),
-                zip=True,
-            )
+            try:
```
balikasg, line 638: This would fail in case the `file_path` does not exist I assume. Any other cases?

balikasg: OK just saw the ErrorType (`FileNotFoundError`)..

python/ml4ir/base/model/relevance_model.py
```diff
@@ -635,11 +635,14 @@ def save(
...
+                self.file_io.save_numpy_array(
+                    np_array=layer.get_weights(),
+                    file_path=os.path.join(model_file, "layers", "{}.npz".format(layer.name)),
+                    zip=True,
+                )
+            except FileNotFoundError:
+                self.logger.warning("Error saving layer: {}. Skipping...".format(layer.name))
```
balikasg, line 645: Maybe `Error saving layer: {} at {} due to FileNotFoundError`. Skipping... 
The user will be able to fix this easier with such a message

python/ml4ir/base/features/feature_fns/sequence.py
lastmansleeping: I initially was planning to use `Global1DPooling`. If you see, the name of my function was "inspired" from keras itself. I don't think the keras function supports padding/masking. So I had to do something custom. Let me know if there are other alternatives.

And yes, it was blowing up for padded records where all values are zero. I'm happy to hear better ways to handle this too.

python/ml4ir/base/model/relevance_model.py
lastmansleeping: I like it. Will update.

python/ml4ir/base/features/feature_fns/sequence.py
balikasg: I am not aware of other alternatives, mostly because I have not deep-dived in tf code. Perhaps the user should have checked in their data, but we ought to be able to handle the case I guess. I should check a ranking dataset to better underrstand the usage. Let's move on with this. Can you: 
- Update the docstring with a sentence saying we are doing some operations with masks
- ensure that `file_io` is used in this function (from a cursory look we don't use it)

lastmansleeping: Not sure I understand why we need `file_io` here. What do we need to read? The data is already tensors.

balikasg: yes, I agree. This function (`global_1d_pooling`) in its arguments (above, see the function definition) gets `file_io`. I think it is an argument that we do not use and we can remove. I can not comment in the `def` because the changes are far from there.

lastmansleeping: Added docstring.

balikasg (approved):

PR #101, opened 2021-01-11
numpy save of prediction probabilities instead of plain csv

balikasg:
Saving as a separate numpy file instead of csv the predictions. 
Does ~4-5 times better in terms of space

### DIFF
--- a/python/ml4ir/applications/classification/model/classification_model.py
+++ b/python/ml4ir/applications/classification/model/classification_model.py
@@ -221,20 +221,23 @@ class ClassificationModel(RelevanceModel):
             made with the `RelevanceModel`
         """
         if logs_dir:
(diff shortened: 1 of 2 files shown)


jakemannix:
-1 on this - yes, it's smaller, but it is impossible to read .np files from the JVM, from what I can tell.  

If you want a compact format, create a protobuf schema of some kind, so we can read / write in both java/scala and python

balikasg:
@jakemannix thanks for the comment. 
You are right, although I am not sure there is a need to have this readable from scala. 
We can keep the csv if we don't want the npz. No need to go to protobuf which is also complex

jakemannix:
> You are right, although I am not sure there is a need to have this readable from scala.

@ducouloa is working on doing integration tests where the scala side will want to check that the results of training / python prediction, so that's one case.  Also: just human eyes reading it.

PR #102, opened 2021-01-13
Integration test

ducouloa:


python/generate_classification.sh
```diff
@@ -0,0 +1,17 @@
...
+
+cd ${DIR}
+
+CLASSIFICATION_MODEL=ml4ir/applications/classification
+
+PYTHONPATH=. python ${CLASSIFICATION_MODEL}/pipeline.py \
+    --data_dir ${CLASSIFICATION_MODEL}/tests/data/csv/  \
+    --feature_config ${CLASSIFICATION_MODEL}/tests/data/configs/feature_config.yaml \
+    --model_config ${CLASSIFICATION_MODEL}/tests/data/configs/model_config.yaml \
+    --run_id end_to_end_classif \
```
ducouloa, line 12: This name needs to be factorized I guess.

python/generate_classification.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
ducouloa, line 1: Not sure where this script should belong

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -0,0 +1,109 @@
...
+
+    val bundleExecutor = new TFRecordExecutor(
+      bundlePath,
+      ModelExecutorConfig(
+        queryNodeName = "serving_tfrecord_protos",
+        scoresNodeName = "StatefulPartitionedCall_3"
+      )
+    )
+
+    // TODO: using the model yaml don't work.
```
ducouloa, line 91: I think this could be a child work item

jvm/ml4ir-inference/pom.xml
```diff
@@ -10,6 +10,44 @@
   <artifactId>ml4ir-inference</artifactId>
   <packaging>jar</packaging>
 
+  <profiles>
+    <profile>
+      <id>integration_tests</id>
```
ducouloa, line 15: so the command to run the integration_test is 
`mvn verify -Pintegration_tests`

I want to exclude unit test from this target, but no success at the moment.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
balikasg: Can you elaborate what you mean here?

python/generate_classification.sh
balikasg: create an `integration_test` dir under classification tests? Here is not a good place ;-)

python/generate_classification.sh
balikasg: `e2e`? I don't think this will block us

python/generate_classification.sh
balikasg: can we rename the whole file to something more descriptive? Perhaps `run_classifier.py` or `train_and_predict.py` or something else? the `generate` usually refers to generating classification data which is not what we are doing here.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -0,0 +1,109 @@
...
+        vector += line
+      }
+    }
+    vectorList.toList
+  }
+
+
+  @Test
+  def testClassificationGeneratedModelBundle(): Unit = {
+    // TODO: This need to be read from the mvn config
```
balikasg, line 77: the hard-coded paths you mean?

jvm/ml4ir-inference/pom.xml
balikasg: Do we run this today?

python/generate_classification.sh
ducouloa: I meant that `end_to_end_classif` will create a directory that is then used in the jvm tests. This should be a param from the build system, passed to this script.

jvm/ml4ir-inference/pom.xml
ducouloa: Not in circleCI.

I think it worths a (quick) child story

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
ducouloa: I'm using the `classification/feature_config_with_same_name.yaml` that works for me but that is not the one that is used to generate de model... 

I think we need to investigate the issue.

python/generate_classification.sh
ducouloa: Oops, I missed your naming recommendation... I went for "generate_model.sh"... I could do train_model.sh if this suits you best.

jvm/ml4ir-inference/scripts/generate_model.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Script to train a model on a classification dataset in the python project
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+cd ${DIR}/../../../python
+
+CLASSIFICATION_MODEL=ml4ir/applications/classification
+
+PYTHONPATH=. python ${CLASSIFICATION_MODEL}/pipeline.py \
```
jakemannix, line 11: PYTHONPATH=. ?

Will this run with a set-up virtualenv?  Or what is it using as its python environment?

jvm/ml4ir-inference/pom.xml
```diff
@@ -10,6 +10,73 @@
...
+                  <goal>test</goal>
+                </goals>
+                <configuration>
+                  <systemPropertyVariables>
+                    <bundleLocation>${bundleLocation}</bundleLocation>
+                    <runName>${runName}</runName>
+                  </systemPropertyVariables>
+
+                  <excludes>
+                    <exclude>TensorFlowInferenceTest</exclude>
```
jakemannix, line 65: why is this "excluded"?

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -0,0 +1,109 @@
...
+  val classLoader = getClass.getClassLoader
+
+  def resourceFor(path: String) = classLoader.getResource(path).getPath
+
+  /**
+   * Class holding a Example and prediction from Python
+   * @param sequence the Example context
+   * @param scores the prediction scores
+   */
+  class PredictionVector(var sequence: Map[String, String], var scores: Array[Float])
```
jakemannix, line 23: if you can avoid ever using "var", do - use "val" (the scala equivalent of a final variable).  But in this case, even better: use case class:

`case class PredictionVector(sequence: Map[String, String], scores: Array[Float])`

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Ugly parsing of the Python CSV predicition.
+   *
+   * Mainly removing the "b'" formatting and slicing into List and Map.
```
jakemannix, line 28: so do we know that the serialization that python uses allows us to just pull those b"blah" stuff out directly?  I think they're perhaps "UTF-8" byte-strings?

jvm/ml4ir-inference/scripts/generate_model.sh
ducouloa: Yes, this is tailored for virtualenv. 
This PYTHONPATH is supposed to be set in the user environment, but we're not sure it is.

jvm/ml4ir-inference/pom.xml
ducouloa: good catch, this is a leftover.
But I think we would like to clean up those test in the near future.

jvm/ml4ir-inference/scripts/generate_model.sh
jakemannix: How does it know to use the virtualenv?

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
ducouloa: Here's a snippet:

> query_key,query_text,query_words,domain_id,user_context,code,entity_id,category_label
> b'query_id_0',b'bernardo and unfold',"[b'bernardo' b'and' b'unfold' b'' b'' b'' b'' b'' b'' b'' b'' b'' b'' b''
>  b'' b'' b'' b'' b'' b'']",b'S',"[b'DDD' b'' b'' b'' b'' b'' b'' b'' b'' b'' b'' b'' b'' b'' b'' b'' b''
>  b'' b'' b'']",b'canary',[0. 0. 1. 0. 0. 0. 0. 0. 0.],"[2.9891068e-01 2.4615353e-01 2.2974105e-01 8.6946599e-02 1.3798673e-01
>  5.1893934e-05 6.1736333e-05 1.2362415e-04 2.4173807e-05]"
> b'query_id_1',b'stand elsinore prince am yourself make twelve',"[b'stand' b'elsinore' b'prince' b'am' b'yourself' b'make' b'twelve' b''
>  b'' b'' b'' b'' b'' b'' b'' b'' b'' b'' b'' b'']",b'L',"[b'EEE' b'EEE' b'BBB' b'CCC' b'BBB' b'BBB' b'AAA' b'EEE' b'GGG' b'FFF'
>  b'CCC' b'GGG' b'GGG' b'' b'' b'' b'' b'' b'' b'']",b'canary',[0. 0. 0. 0. 1. 0. 0. 0. 0.],"[3.0492017e-01 1.9346356e-01 2.2024015e-01 1.2455909e-01 1.5630856e-01
>  9.8194840e-05 1.1984430e-04 2.1972452e-04 7.0748545e-05]"

It's numpy.to_csv I think. This adds the extra typing stuff around the CSV value (like the bracket to indicate a list etc...).
I've look to change this output, I'm not sure there is an easy way.

jakemannix: we should make a note (followup story) to pick a better serialization format for that CSV.  Maybe store it in JSON, at the very minimum, and without python-isms like the b'foo' stuff.

jvm/ml4ir-inference/scripts/generate_model.sh
ducouloa: It this current form, the script is "just" calling `python` as you can see.
If you env is correctly setup (bin/activate stuff) this would work.

I expect that this will be the case in the circleCI docker image that would run the test.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
ducouloa: When doing so, the compiler complains line 103 where I'm getting the value of the sequence and scores directly:
 
`queryContext.sequence.asJava`

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
ducouloa: It may be addressable in this review.

Initially I wrote a standalone script that was calling the `predict` method with a param to log prediction in an external log files. For what I can remember the output was cleaner than that. Be we lose the benefit of not having to rely on an other script to generate a model.

Or I can try to fix the output.csv in Python.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -0,0 +1,109 @@
...
+      new PredictionVector(sequence, scores)
+    } catch {
+      case _: NumberFormatException => null
+    }
+
+  }
+
+  /**
+   * Read a prediction file from the python project.
+   * The prediction csv is not 1 row per line, so we need to reconcile the file.
```
jakemannix, line 51: what do you mean by "not 1 row per line"?

ducouloa: the 8 columns values of this CSV will be split in a variable number of lines... 

This is the first prediction_vector for example (4 lines in the csv output from python out of the box)

```
b'query_id_0',b'bernardo and unfold',"[b'bernardo' b'and' b'unfold' b'' b'' b'' b'' b'' b'' b'' b'' b'' b'' b''
b'' b'' b'' b'' b'' b'']",b'S',"[b'DDD' b'' b'' b'' b'' b'' b'' b'' b'' b'' b'' b'' b'' b'' b'' b'' b''
b'' b'' b'']",b'canary',[0. 0. 1. 0. 0. 0. 0. 0. 0.],"[2.9891068e-01 2.4615353e-01 2.2974105e-01 8.6946599e-02 1.3798673e-01
```

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Ugly parsing of the Python CSV predicition.
+   *
+   * Mainly removing the "b'" formatting and slicing into List and Map.
+   *
+   * @param line
+   * @return
+   */
+  def extractColumnValues(line: String): PredictionVector = {
```
jakemannix, line 33: change the method signature to be `Option[PredictionVector]`, then when you get number parsing exceptions, you can return None, otherwise return `Some(PredictionVector(sequence, scores))`

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -0,0 +1,109 @@
...
+   *
+   * @param csvFile path to the CSV
+   * @return List of PredictionVector
+   */
+  def readPredictionCSV(csvFile: String): List[PredictionVector] = {
+    val lines = Source.fromFile(csvFile).getLines.toList
+
+    var vectorList = new ListBuffer[PredictionVector]
+    var vector = ""
+    for (line <- lines.drop(1)) {
```
jakemannix, line 61: the predictions span multiple lines?

jvm/ml4ir-inference/scripts/generate_model.sh
jakemannix: Not sure, we'll have to check.  We may want to make sure it's been called.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
jakemannix: that's not a useful format.  Ugh.  We need to fix that.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
jakemannix: what does it say?  Did you make it a *case* class, not just a "class"?

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
jakemannix: @balikasg / @lastmansleeping do we programmatically (in python) read the output.csv anywhere?  Or is it just for debugging / humans?

lastmansleeping: Not for ranking at least. It is just for debugging and error analysis (which we haven't really done much of). I remember the QU team considering reading the output file to compute further metrics in a spark job. Not sure if that fruitioned though.

jakemannix: if they want to read it in a spark job, definitely want it in a better format than this.

Alain - looks like maybe changing the format from the python side, might be best?

jakemannix (changes requested):
As for next steps: I need to try out this branch and see it purring along, and I'm ok with adding automated running of this test in circleci as followup, sure (I think maybe the only real "gotcha" is finding a good docker base image that has *both* the jvm *and* the required python stuff, and then creating an "almost union" of the .circleci/config.yaml file, running both stages).  We also need to make sure and fix up that .csv writing - that parsing code is... a bit scary because of the formatting of the output.csv file.  

Then we just need to see why you're getting a scoring discrepancy.  Having it evaluate the model at all is a good (necessary) step.  Next up: actually have the test pass :)

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
balikasg: this is how bytes are saved in a CSV
For python, simply calling `ast.literal_eval` on the column fixes the issue. We can fix this before saving the csv if it is a problem, I see in scala it needs a hack..

balikasg: and yes, we are using this output exactly as @lastmansleeping said!

balikasg: WE should make the change in Python side, yes. I will try to raise a PR today

ducouloa: @balikasg I can address this in this current PR I think.

balikasg:
I am taking over the ml4ir code changes to not save bytes to remove this ugly parsing. ETA: today on Monday the latest.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
balikasg: I think it would be best to open a separate PR for this

jvm/ml4ir-inference/pom.xml
```diff
@@ -10,6 +10,71 @@
...
+                    <argument>${runName}</argument>
+                  </arguments>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-failsafe-plugin</artifactId>
```
ducouloa, line 51: Note: as stated in the commit message, I had to move to failsafe for integration-test.

The tweak with the surefire was not working fine apparently, as it seems the test aren't run at the integration-test phase.

I hope it won't make a difference during the 3PP approval process?

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
ducouloa: Yes, first line is the header ( I skip it) then we ran over ~100 predictions

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
ducouloa: 😓 case class... that was it...
Many thanks for this. Whoa.. looks like Scala has some goodies embedded!

balikasg (dismissed):

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+
+  /**
+   * Read a prediction file from the python project.
+   *
+   * @param csvFile path to the CSV
+   * @return List of PredictionVector
+   */
+  def readPredictionCSV(csvFile: String): List[PredictionVector] = {
```
jakemannix, line 52: so now that the CSV are not full of line-breaks, you can do this method without any mutable state, in one line of scala:

`Source.fromFile(csvFile).getLines.toList.tail.map(extractColumnValues)`

python/ml4ir/applications/classification/model/classification_model.py
```diff
@@ -232,7 +232,11 @@ def predict(
         # than a list on numpy arrays
         predictions_df[self.output_name] = [x for x in predictions_]
         if logs_dir:
-            np.set_printoptions(threshold=sys.maxsize)  # write the full vector in the csv not ...
+            np.set_printoptions(formatter={'all':lambda x: str(x.decode('utf-8')) if isinstance(x, bytes) else str(x)},
```
jakemannix, line 235: did we check that this doesn't break @balikasg and @Ullimague 's evaluation spark job for entity prediction, when this CSV changes it's shape a little?

jakemannix (commented):
Looking pretty good now that the CSV isn't so messed up.

jvm/ml4ir-inference/pom.xml
jakemannix: 3PP should be fine

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
jakemannix: although since extractColumnValues returns Option[T] now, maybe that last method would actually be `.flatMap(extractColumnValues)`.  That works, I think.

jakemannix (approved):
lgtm

python/ml4ir/applications/classification/model/classification_model.py
balikasg: I am aware of the change, I will adapt the spark job by simplifying it.

balikasg (approved):

PR #103, opened 2021-01-20
Updating and cleaning python libraries

Ullimague:
Keeping python library to minimum requirements.


jakemannix (commented):
and all tests pass, yay!

If QA for the story can make sure to actually train both kinds of models with this branch, and verify notebooks still work, this looks like a way trimmed down set of dependencies.

How does the code parse the YAML if we remove PyYAML, now?

lastmansleeping (approved):
LGTM. Was able to build the image locally, install dependencies, run all tests and train both ranking and EP models successfully. Thanks for the work @Ullimague

PR #105, opened 2021-01-21
Setting `drop-remainder=False` to guarantee we predict on all instances

balikasg:


python/ml4ir/applications/classification/tests/test_classification_model.py
```diff
@@ -14,25 +14,29 @@ def test_csv_loss_metrics(self):
...
+        self.assertTrue(np.isclose(self.metrics_dict["loss"], expected_loss, atol=tol),
                         msg=f"Loss not in expected range."
                             f" Expected: {expected_loss} ± {tol}, Found: {self.metrics_dict['loss']}")
 
-        expected_acc = 0.203
-        tol = 0.01
-        self.assertTrue(np.isclose(self.metrics_dict["categorical_accuracy"], expected_acc, rtol=tol),
+        expected_acc = 0.2
+        tol = 0.05
+        self.assertTrue(np.isclose(self.metrics_dict["categorical_accuracy"], expected_acc, atol=tol),
```
balikasg, line 25: absolute points work better imho here as the numbers as small. Also, from a coders perspective is is easier to understand

python/ml4ir/applications/classification/tests/test_classification_model.py
```diff
@@ -14,25 +14,29 @@ def test_csv_loss_metrics(self):
...
         tol = 0.01
         _metric = "top_5_categorical_accuracy"
-        self.assertTrue(np.isclose(self.metrics_dict[_metric], expected_acc, rtol=tol),
+        self.assertTrue(np.isclose(self.metrics_dict[_metric], expected_acc, atol=tol),
                         msg=f"Top5 Categorical_accuracy not in expected range."
                             f" Expected: {expected_acc} ± {tol}, Found: {self.metrics_dict[_metric]}")
 
+        # Assert we predict for all the items
+        expected_size_predictions = 200  # the same with data in test
+        self.assertTrue(self.predictions.shape[0] == expected_size_predictions)
```
balikasg, line 38: add a classification test to ensure we predict on all instances.

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -212,10 +212,9 @@ def test_cyclic_lr_in_training_pipeline(self):
                    0.010900003,0.001,0.0059499955,0.010900003,0.015849996,0.020800006,0.02575,0.020800006,0.015849996,
                    0.010900003,0.0059499955,0.001,0.0034749978,0.0059500015,0.008424998,0.010900003,0.013375,
                    0.010900003,0.008424998,0.0059500015,0.0034749978,0.001,0.0022374988,0.0034749978,0.0047125025,
-                   0.0059500015,0.0071875,0.0059500015,0.0047125025]
+                   0.0059500015,0.0071875,0.0059500015,0.0047125025,0.0034749978, 0.0022374988]
 
-        for i in range(len(lr_list)):
-            assert np.isclose(lr_gold[i], lr_list[i])
+        assert np.all(np.isclose(lr_gold, lr_list))
```
balikasg, line 217: I decided to vectorize this.

python/ml4ir/base/data/tfrecord_reader.py
```diff
@@ -761,7 +761,7 @@ def read(
 
     # Create BatchedDataSet
     if batch_size:
-        dataset = dataset.batch(batch_size, drop_remainder=True)
+        dataset = dataset.batch(batch_size, drop_remainder=False)
```
balikasg, line 764: the gist of this PR.

lastmansleeping: @balikasg I thought that for training, we would still drop the remainder?

balikasg: this is what was hinted by @jakemannix in the slack discussion. But: 
- I could not think of an advantage doing so. 
- Doing so would require an extra argument to be exposed to allow tweaking with this value depending on `fit`/`predict`/`evaluate`. I think this would lead to unnecessary code complexity. 
So i decided to go with the simpler solution. Is there any particular reason you would prefer to have `drop_remainder=True` during training @lastmansleeping ?

Ullimague (approved):
This LGTM, I'd be more comfortable for @lastmansleeping to approval as well, as having more context on it.

balikasg:
thanks @Ullimague, noted, I will wait a bit more and merge tomorrow!

python/ml4ir/base/data/tfrecord_reader.py
lastmansleeping: Not really. I just worried ranking model would not work with `drop_remainder=False` at training time. I've seen this in the past. But happy to see all tests pass.

lastmansleeping (approved):
@balikasg LGTM. I had reviewed this last week. I forgot to check back for your reply!

I was just worried setting `drop_remainder=False` would break training. Looks like it doesn't. So no worries from my end.

PR #106, opened 2021-01-25
Adding LinearRankingModel @rev asaintrequier@

lastmansleeping:
This PR adds the changes needed to train (no changes necessary) and save a simple one layer linear model for ranking. The PR also contains accompanying config files and unit tests.


python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -272,3 +281,86 @@ def mask_padded_records(predictions, features_dict):
...
+            ),
+            name="weight")
+        self.logger.info("Linear Model Coefficients:\n{}".format(
+            linear_model_coefficients.to_csv()))
+        self.file_io.write_df(
+            linear_model_coefficients,
+            outfile=os.path.join(models_dir, "coefficients.csv"),
+            index=True
+        )
+
```
lastmansleeping, line 358: @Ullimague This is the main change.

lastmansleeping: @asaintrequier FYI

python/ml4ir/applications/ranking/config/parse_args.py
```diff
@@ -25,6 +25,13 @@ def define_args(self):
             "applications/ranking/config/keys.py",
         )
 
+        self.add_argument(
+            "--use_linear_model",
```
Ullimague, line 29: I was wondering if we should use an enum such as "model_type". This can live like this for now, but we will have to have other bools and deal with conflicts when we come-up with mode then generic vs linear model.

One other way would be to have a param/field "type: RankingModel", "LinearRankingModel" in the feature config yaml.

Any thoughts @lastmansleeping @jakemannix?

python/ml4ir/applications/ranking/tests/test_ranking_model.py
```diff
@@ -93,3 +94,35 @@ def test_csv_and_tfrecord(self):
         # Compare CSV and TFRecord loss and accuracies
         assert np.isclose(tfrecord_loss, csv_loss, rtol=0.01)
         assert np.isclose(tfrecord_mrr, csv_mrr, rtol=0.01)
+
+    def test_linear_ranking_model_save(self):
```
Ullimague, line 98: Do we also plan to follow-up / have test to load the model @lastmansleeping (in python and jvm side)?

python/ml4ir/applications/ranking/config/parse_args.py
lastmansleeping: Yeah, I like the former for now. The problem (maybe, I'm not sure yet) with the latter approach is that a linear model might have multiple different meanings. As in, we can choose to add dropout, other post score layers, maybe even feature function layers, etc. I did try to do the latter but then dropped it for these reasons. I can do the former if you think that's a good idea.

python/ml4ir/applications/ranking/tests/test_ranking_model.py
lastmansleeping: Replying on slack due to confidentiality.

python/ml4ir/applications/ranking/config/parse_args.py
balikasg: This is not correct use of arguments in Python and argparse. As a matter of fact, I believe this will not work, passing something as `bool` does not do True/False as one would expect. Please see here: https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/base/config/parse_args.py#L256

Also, the most pythonic way to do this to pass with an action store true/false.

@Ullimague argparser support choices out of the box, see here: https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/base/config/parse_args.py#L62

balikasg: `loss_type` and `scoring_type` can be improved a lot too using these two pointers (all necessary checks will be done by argparser)

python/ml4ir/applications/ranking/tests/data/configs/linear_model/model_config.yaml
```diff
@@ -0,0 +1,39 @@
...
+    activation: null
+optimizer:
+  key: adam
+  gradient_clip_value: 5.0
+lr_schedule:
+  key: constant
+  learning_rate: 0.01
+
+
+#Below is an example of defining other learning rate schedules.
```
balikasg, line 18: Better remove all these comments ;-)

balikasg (changes requested):
My main points are easy to fix: 
- Update argparser to be used expected
- Remove commented code unless needed

python/ml4ir/applications/ranking/tests/data/configs/linear_model/model_config.yaml
lastmansleeping: Will do. I had left it as a reference to use other `lr_schedule` in case people were interested.

lastmansleeping:
@Ullimague I've moved the linear model to the model_config YAML as we discussed. You can set the config to a linear model using `architecture_key: linear` now.

@asaintrequier To test the change, you can run:
```
--feature_config ml4ir/applications/ranking/tests/data/configs/linear_model/feature_config.yaml \
--model_config ml4ir/applications/ranking/tests/data/configs/linear_model/model_config.yaml \
--run_id test \
--data_format tfrecord \
--loss_key rank_one_listnet \
--num_epochs 1
```

python/ml4ir/applications/ranking/config/parse_args.py
lastmansleeping: @balikasg Thanks, I will update to this.

lastmansleeping:
@asaintrequier I've addressed the comment we discussed earlier - add column name as `feature` and `weight`.

asaintrequier (approved):

jakemannix:
Question: while internally, we implement linear ranking models in Solr, in this open source setup, we need to verify that JVM inference works for this as well.

PR #107, opened 2021-01-26
Run integration test with CircleCI

ducouloa:


balikasg:
Great progress @ducouloa ! Excited to see this!

.circleci/config.yml
```diff
@@ -2,32 +2,38 @@
...
+            . venv/bin/activate
             cd jvm
-            mvn test
-            mvn surefire-report:report-only
-      - store_artifacts:
-          path: jvm/ml4ir-inference/target/site/surefire-report.html
+            mvn verify -Pintegration_tests
       - store_test_results:
-          path: jvm/ml4ir-inference/target/site/surefire-report.html
+          path: jvm/ml4ir-inference/target
```
jakemannix, line 36: we keep the entire target dir now?  why?

.circleci/config.yml
```diff
@@ -53,12 +59,10 @@ jobs:
...
             . venv/bin/activate
             cd python
-            python3 -m pytest ml4ir/applications/classification/tests --html=pytest_report.html
+            python3 -m pytest ml4ir/applications/classification/tests --junitxml=test_results/results.xml
           no_output_timeout: 45m  # Time for running the tests before we timeout
-      - store_artifacts:
-          path: python/classification/pytest_report.html
       - store_test_results:
-          path: python/classification/pytest_report.html
+          path: python/test_results
```
jakemannix, line 65: do we need to make sure this is in .gitignore?

jakemannix (dismissed):
can you make sure to take care of the CLA issue, @ducouloa ?

Otherwise, looks great!

.circleci/config.yml
ducouloa: 1. The path to the report was not good, it doesn't exist (if you look at previous jobs, you would see no test results aggregate)
2. Putting the target directory allows to aggregate the results of surefire (utest) and failsafe (itest) at once.
3. It's CircleCi compatible to point to a "root" result directory, and circleCi infers the test suite name from the generic result directory names.

.circleci/config.yml
ducouloa: That makes sense. This will only leave inside the Docker container and not clutter user's workspace though.

balikasg (approved):
LGTM

PR #109, opened 2021-02-01
Using python 3.7, and same base image as ci

Ullimague:


python/ml4ir/build/Dockerfile
```diff
@@ -1,18 +1,12 @@
 # Using python based centos7 docker base image:
-FROM centos/python-36-centos7:latest
+FROM ducouloa/ml4ir:0.1
```
Ullimague, line 2: Since is mage is used as part of out ci: https://github.com/salesforce/ml4ir/blob/master/.circleci/config.yml#L7
Built from this Dockerfile https://github.com/salesforce/ml4ir/blob/master/.circleci/images/Dockerfile

lastmansleeping: @Ullimague Should we create an account for salesforce search, perhaps?

jakemannix: we could, sure, but not sure if it's needed for now.

jakemannix (approved):
go for it!

PR #112, opened 2021-02-17
Upgrading ml4ir version to 0.0.4 

lastmansleeping:
I've bumped the ml4ir python module version to 0.0.4 and published the whl to PyPi -> https://pypi.org/project/ml4ir/

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -16,7 +16,7 @@ def getReadMe():
 setup(
     name="ml4ir",
     packages=find_namespace_packages(include=["ml4ir.*"]),
-    version="0.0.3",
+    version="0.0.4",
     description="Machine Learning libraries for Information Retrieval",
     long_description=getReadMe(),
     long_description_content_type="text/markdown",


python/setup.py
```diff
@@ -16,7 +16,7 @@ def getReadMe():
 setup(
     name="ml4ir",
     packages=find_namespace_packages(include=["ml4ir.*"]),
-    version="0.0.3",
+    version="0.0.4",
```
balikasg, line 19: thanks!

balikasg (approved):

python/setup.py
ducouloa: @balikasg , @lastmansleeping can we tag the repository when we are changing the version?
Could we also add some basic release notes in a dedicated file or in the tag?

I have a hard time figuring out why we increment the version number without context here.

lastmansleeping: Sure. Can you share any examples for that @ducouloa ?

PR #113, opened 2021-02-17
Cleaning up base requirements.txt and adding changelog

lastmansleeping:
- Removing unnecessary dependencies from base `requirements.txt`
- Adding a changelog for version management

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -7,6 +7,7 @@ pandas==1.2.1
 pytest==6.2.1
 pytest-cov==2.11.0
 pytest-html==3.1.1
+PyYAML==5.4.1
 
 # Pyspark for spark_io
 pyspark==3.0.1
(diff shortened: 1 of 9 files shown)


python/requirements.txt
```diff
@@ -17,6 +7,7 @@ pandas==1.2.1
 pytest==6.2.1
 pytest-cov==2.11.0
 pytest-html==3.1.1
+PyYAML==5.4.1
```
jakemannix, line 10: are we adding this back in on purpose?

lastmansleeping, line 10: @Ullimague @jakemannix I had to add this because the library that was installing PyYAML transitively must have been moved to `build-requirements.txt` now.

jakemannix (approved):
As long as we know PyYaml is ok, or remove it, LGTM

python/requirements.txt
lastmansleeping: I had to add this because the library that was installing PyYAML transitively must have been moved to build-requirements.txt now.

jakemannix: hmmm... but do we have approval for it?

PR #115, opened 2021-02-22
temperature_scaling

marjanhs:
implementing temperature scaling for (classification) models.


python/ml4ir/base/config/default_model_config.yaml
```diff
@@ -24,6 +24,9 @@ optimizer:
 lr_schedule:
   key: constant
   learning_rate: 0.01
+calibration:
```
lastmansleeping, line 27: @marjanhs Perfect.

python/ml4ir/base/pipeline.py
```diff
@@ -390,6 +390,26 @@ def run(self):
...
+            if self.args.execution_mode in {
+                ExecutionModeKey.TRAIN_INFERENCE_EVALUATE,
+                ExecutionModeKey.TRAIN_EVALUATE,
+                ExecutionModeKey.TRAIN_INFERENCE,
+                ExecutionModeKey.TRAIN_ONLY,
+                ExecutionModeKey.INFERENCE_EVALUATE_RESAVE,
+                ExecutionModeKey.EVALUATE_RESAVE,
+                ExecutionModeKey.INFERENCE_RESAVE,
+            }:
+                if 'calibration' in self.model_config:
```
lastmansleeping, line 403: [Minor] @marjanhs You should probably add "calibration" as a string under `ml4ir/base/config/keys` probably ModelConfigKeys? Should be fine for now though.

python/requirements.txt
```diff
@@ -11,3 +11,6 @@ PyYAML==5.4.1
 
 # Pyspark for spark_io
 pyspark==3.0.1
+
+#calibration
+tensorflow_probability==0.8.0
```
lastmansleeping, line 16: @marjanhs Can you add this to `build-requirements.txt` as well?

marjanhs: added

lastmansleeping (changes requested):
@marjanhs I know you are still working on it, but I am just adding for visibility -> Try to add a couple of unit tests for the functions under `temperature_scaling`.

python/ml4ir/base/pipeline.py
```diff
@@ -390,6 +390,26 @@ def run(self):
...
+                    if self.model_config['calibration']['key'] == 'temperature_scaling':
+                        temperature = self.model_config['calibration']['temperature'] if \
+                            'temperature' in self.model_config['calibration'] else 1.5
+                        temperature_scale(relevance_model=relevance_model,
+                                          dataset=relevance_dataset,
+                                          logger=self.logger,
+                                          logs_dir_local=self.logs_dir_local,
+                                          temperature_init=temperature)
+
```
lastmansleeping, line 412: @marjanhs I recommend moving this call inside of `RelevanceModel.calibrate()`. That will allow you to keep the logic of mapping model config to the calibration function contained to RelevanceModel. Keep in mind that customers might use RelevanceModel as is by importing without the pipeline. The pipeline is just an example of one such use.

python/ml4ir/base/pipeline.py
```diff
@@ -445,6 +465,11 @@ def finish(self, job_status, job_info):
 
         return self
 
+    @staticmethod
+    def execute_temperature_scaling(relevance_model, dataset, logger, logs_dir_local):
```
lastmansleeping, line 469: @marjanhs Where is this function used?

python/ml4ir/base/pipeline.py
```diff
@@ -390,6 +390,26 @@ def run(self):
                 experiment_tracking_dict, orient="index", columns=["value"]
             ).to_csv()
 
+            # temperature scaling
+            if self.args.execution_mode in {
```
lastmansleeping, line 394: @marjanhs Thinking out loud - should we do this right after training, before we evaluate? Does that make sense? Will the model saving work that way?

marjanhs: It needs to be applied on a trained model. It does not matter its place after or before evaluation. Why do u thin it won't work if it moves to a place before evaluation ?

python/ml4ir/base/pipeline.py
marjanhs: removing it ...

python/ml4ir/base/pipeline.py
marjanhs: ok, changed accordingly.

python/ml4ir/base/pipeline.py
marjanhs: changed accordingly.

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -364,3 +364,6 @@ def save(
             required_fields_only=required_fields_only,
             pad_sequence=pad_sequence,
         )
+
+    def calibration(self, **kwargs):
```
lastmansleeping, line 368: [Minor] @marjanhs Typically, it is best practice to name functions with verbs. So maybe `calibrate()`?

python/ml4ir/base/config/keys.py
```diff
@@ -131,3 +131,9 @@ class FileHandlerKey(Key):
 
     LOCAL = "local"
     SPARK = "spark"
+
+
+class CalibrationKey(Key):
```
lastmansleeping, line 136: Nice!

python/ml4ir/base/model/relevance_model.py
```diff
@@ -778,3 +779,11 @@ def _build_callback_hooks(
         # Add more here
 
         return callbacks_list
+
+    def calibration(self, relevance_dataset, logger, logs_dir_local, temperature_init):
+        temperature_scale(model=self.model,
```
lastmansleeping, line 784: @marjanhs You can maybe add an info log here saying that you are calibrating the model.

python/ml4ir/base/pipeline.py
```diff
@@ -296,7 +297,7 @@ def run(self):
             relevance_model = self.get_relevance_model()
             self.logger.info("Relevance Model created")
 
-            if self.args.execution_mode in {
+            '''if self.args.execution_mode in {
```
lastmansleeping, line 300: @marjanhs Did you mean to comment this out? For testing, you could probably run in save only mode to make things faster, I guess.

python/ml4ir/base/pipeline.py
lastmansleeping: For temperature scaling, I don't think it matters. It might for other techniques. Not sure at the moment. Let's leave it here for now. Looks good.

python/ml4ir/base/model/relevance_model.py
marjanhs: added

python/ml4ir/base/pipeline.py
marjanhs: oh yes, sorry . fixed it

python/ml4ir/applications/ranking/model/ranking_model.py
marjanhs: changed

python/ml4ir/base/config/default_model_config.yaml
```diff
@@ -24,6 +24,9 @@ optimizer:
...
+  key: temperature_scaling
```
balikasg, line 28: should we calk it `method` instead of `key`? it is fair to point that the same is happening to the `optimizer` and `lr_schedule` sections though.

python/ml4ir/base/model/calibrations/temperature_scaling.py
```diff
@@ -0,0 +1,247 @@
...
+from ml4ir.base.data.relevance_dataset import RelevanceDataset
+from ml4ir.base.model.scoring.scoring_model import ScorerBase
+
+import shutil
+
+TEMPERATURE_SCALE = 'temp_scaling_scores'
+
+
+def make_val_and_grad_fn(fn):
+    """ Returns a function that computes y = fn(*args, **kwargs) and gradients of y wrt
```
balikasg, line 25: thank you for adding docstrings @marjanhs ! It helps! 
Please ensure consistency ie., start with capital, no space between quotes and first letter.

python/ml4ir/base/model/calibrations/temperature_scaling.py
```diff
@@ -0,0 +1,247 @@
...
+
+def run(optimizer, logger):
+    """Run an optimizer and measure it's evaluation time."""
+    optimizer()  # Warmup.
+    with timed_execution(logger):
+        result = optimizer()
+    return np_value(result)
+
+
+def accuracy(scores, labels):
```
balikasg, line 69: FYI We do have dedicated metrics files where we define some metrics (categorical accuracy among them). 

Perhaps this function is better off living there although I am not 100% on this because here is looks like you want to do an one-off calculation.

python/ml4ir/base/model/calibrations/temperature_scaling.py
```diff
@@ -0,0 +1,247 @@
...
+
+def dict_to_zipped_csv(data_dic, root_dir):
+    """
+    saves input dictionary to a zipped csv file
+    Parameters:
+        data_dic: input dict to be converted to  zipped csv file
+        path: path to save the output file
+    """
+    # creating zip dir
+    zip_dir_path = os.path.join(root_dir, TEMPERATURE_SCALE)
```
balikasg, line 90: FYI there is previous work on io, e.g, [local io](https://github.com/salesforce/ml4ir/blob/23482afcaff23beaead79fb092c297ea9df0cf6e/python/ml4ir/base/io/local_io.py) that offers some of these for convenience. Please take a look if you have not already.

python/ml4ir/base/model/calibrations/temperature_scaling.py
```diff
@@ -0,0 +1,247 @@
...
+                                                         logits_nps_test,
+                                           labels_nps_test, temperature=temper)
+
+    # write the full vector in the csv not ...
+    np.set_printoptions(
+        formatter={'all': lambda x: str(x.decode('utf-8')) if isinstance(x, bytes) else str(x)},
+        linewidth=sys.maxsize,
+        threshold=sys.maxsize)
+    # avoiding .tolist() that is not memory efficient
+    data_dic = {'org_scores': [x for x in org_softmaxes.numpy()],
```
balikasg, line 235: why do we call them 'org_scores' ? Same, below for 'ts_scores'.

python/ml4ir/base/model/calibrations/temperature_scaling.py
```diff
@@ -0,0 +1,247 @@
+"""
+Implementing temperature scaling proposed in https://github.com/gpleiss/temperature_scaling
+The tensorflow implementation and functions are inspired by
+https://www.tensorflow.org/probability/examples/Optimizers_in_TensorFlow_Probability
+"""
+import sys
```
balikasg, line 6: let's call the dir `calibration` instead of `calibrations`
PITA, but I strongly believe it makes more sense.

python/ml4ir/base/model/relevance_model.py
balikasg: thanks!

python/ml4ir/base/pipeline.py
balikasg: @lastmansleeping @marjanhs we should discuss a bit more about this choice (perhaps in a follow up PR with an approved design before). My concerns: 
-  I think it is better to have a dedicated ExecutionModeKey for calibration. It can be misleading to have it as part of the model training for example. This because it is a process done after training completes, possibly requiring its very own data. Which leads me believe that a typical train/val/test split may not be enough and we may overfit in what we are reporting here.  
- What if we support 2 types of calibration and ones wants to compare for example? Would they need 2 model configs, each one repeating the other? Which suggests that the calibration requires a config, but preferably imho its own (clear concern separation) as it seems to be an independent method from the model itself. 
- Right now, it seems we are not saving the model with calibration just some predictions. This can also be misleading for the user and a bit concerning. These predictions are not copied in HDFS (or perhaps they are as part of the finish method of classification). 
- Worth discussing a bit how one would serve this in the scala part too.

python/ml4ir/base/model/relevance_model.py
```diff
@@ -778,3 +779,12 @@ def _build_callback_hooks(
         # Add more here
 
         return callbacks_list
+
+    def calibrate(self, relevance_dataset, logger, logs_dir_local, temperature_init):
```
balikasg, line 783: I am a bit confused now. I see the `NotImplemented` above in ranking model. Should it come here, so that the models inherit it (so we do not touch the ranking file) and define this `calibrate` on the `ml4ir/applications/classification/model/classification_model.py`? Is this implementation specific to classification approaches or classification data?

python/ml4ir/base/pipeline.py
balikasg: I also believe that due to these inconsistencies, we are better using it under an `experimental` name of something. But we may be far from this maturity

balikasg (changes requested):
Most suggestions are good. Please take a look on the comments and let me know what you believe.

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: `org_score`: original score (without temperature scaling)
`ts_score`: scores with temperature scoring

python/ml4ir/base/model/relevance_model.py
marjanhs: It is expected to have calibration for ranking in future. @lastmansleeping  also recommended to keep `calibration` here.

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: thank u. will take a look

python/ml4ir/base/model/calibrations/temperature_scaling.py
```diff
@@ -0,0 +1,247 @@
...
+    learns a temperature parameter using Temperature Scaling (TS) technique on the validation set.
+    It, then, computes the probability scores of the test set with and without TS and writes them
+    in a .zip file.
+    Parameters:
+        model : tf.keras.Model object to be used for temperature scaling
+        scorer:  ScorerBase object of the RelevanceModel
+        dataset: RelevanceDataset to be used for temperature scaling.
+        logger : Logger
+        logs_dir_local: path to save the TS scores
+        temperature_init: temperature initial value
```
lastmansleeping, line 168: @marjanhs Also add what you are returning. Check the format [here](https://numpydoc.readthedocs.io/en/latest/format.html)

python/ml4ir/base/model/calibrations/temperature_scaling.py
```diff
@@ -0,0 +1,247 @@
...
+    """
+    # get an intermediate model with logits as output --> MUST NOT BE AN ACTIVATION (e.g.
+    # SOFTMAX) LAYER
+    intermediate_model = get_intermediate_model(model, scorer)
+
+    start = tf.Variable(initial_value=[temperature_init], shape=(1), dtype=tf.float32, name="temp",
+                        trainable=True)
+
+    def nll_loss_fn(temperature):
```
lastmansleeping, line 177: @marjanhs Why not make this a `tf.function` too?

python/ml4ir/base/model/calibrations/temperature_scaling.py
```diff
@@ -0,0 +1,247 @@
...
+    logits (inputs of softmax).
+    Parameters:
+        model:
+        scorer:
+    """
+    # get  last layer's output  --> MUST **NOT** BE AN ACTIVATION (e.g. SOFTMAX) LAYER
+    final_layer_name = scorer.model_config['layers'][-1]['name']
+    layer_output = model.get_layer(name=final_layer_name).output
+
+    return tf.keras.models.Model(inputs=model.input, outputs=layer_output)
```
lastmansleeping, line 115: @marjanhs One thing I'm concerned about and should be covered in the unittests is you are not freezing any of the weights before you calibrate the model. Every weight will be relearned/adjusted in this case.

python/ml4ir/base/model/calibrations/temperature_scaling.py
```diff
@@ -0,0 +1,247 @@
...
+    Parameters:
+        scorer: Scorer of the RelevanceModel
+        logits: numpy array, logits (input of softmax)
+        labels: numpy array, class labels
+        temperature: TF.Tensor, a parameter of size=(1)
+    Returns:
+        accuracy and NLL loss
+    """
+    # convert to TF.Tensor
+    logits_tensor = tf.constant(logits, name='logits_test', dtype=tf.float32)
```
lastmansleeping, line 130: @marjanhs Why are we converting them to `tf.constant` here? Aren't they already tensors? This might make things slow. (Not 100% sure)

lastmansleeping: If your goal is to cast you can do that with `tf.cast(..., name="...")`

python/ml4ir/base/model/calibrations/temperature_scaling.py
```diff
@@ -0,0 +1,247 @@
...
+            tf.nn.sparse_softmax_cross_entropy_with_logits(labels=labels_tensor,
+                                                           logits=logits_w_temp))
+
+    @tf.function
+    def nll_with_lbfgs():
+        return tfp.optimizer.lbfgs_minimize(make_val_and_grad_fn(nll_loss_fn),
+                                            initial_position=start)
+
+    # evaluation on validation set before temperature scaling
+    logits_nps, labels_nps = get_logits_labels(intermediate_model, dataset.validation)
```
lastmansleeping, line 197: @marjanhs Just for my knowledge - is there anything wrong with using the training set here as well?

python/ml4ir/base/model/calibrations/temperature_scaling.py
```diff
@@ -0,0 +1,247 @@
...
+        labels_tensor = tf.constant(labels_nps, name='labels_tensor', dtype=tf.int32)
+        logits_w_temp = tf.divide(logits_tensor, temperature)
+
+        return tf.reduce_mean(
+            tf.nn.sparse_softmax_cross_entropy_with_logits(labels=labels_tensor,
+                                                           logits=logits_w_temp))
+
+    @tf.function
+    def nll_with_lbfgs():
+        return tfp.optimizer.lbfgs_minimize(make_val_and_grad_fn(nll_loss_fn),
```
lastmansleeping, line 193: @marjanhs Can you add description for why we are using this here instead of say doing SGD? A reference should work too. More as documentation for future.

python/ml4ir/base/model/calibrations/temperature_scaling.py
```diff
@@ -0,0 +1,247 @@
...
+        tensor: TF.Tensor
+                a tensor object
+    """
+    if isinstance(tensor, tuple):
+        return type(tensor)(*(np_value(t) for t in tensor))
+    else:
+        return tensor.numpy()
+
+
+def run(optimizer, logger):
```
lastmansleeping, line 61: [Minor] @marjanhs I would probably call this `tune_probability` or something. `run` was a little confusing.

python/ml4ir/base/model/calibrations/temperature_scaling.py
lastmansleeping: Yes, ideally you should use `local_io` everywhere because we can then swap it out directly for something else later. Even better is pass the `file_io` object from `RelevanceModel` to this module.

python/ml4ir/base/model/calibrations/temperature_scaling.py
```diff
@@ -0,0 +1,247 @@
...
+
+    # write the full vector in the csv not ...
+    np.set_printoptions(
+        formatter={'all': lambda x: str(x.decode('utf-8')) if isinstance(x, bytes) else str(x)},
+        linewidth=sys.maxsize,
+        threshold=sys.maxsize)
+    # avoiding .tolist() that is not memory efficient
+    data_dic = {'org_scores': [x for x in org_softmaxes.numpy()],
+                'ts_scores': [x for x in ts_softmaxes.numpy()],
+                'org_predicted_label': tf.argmax(org_softmaxes, axis=-1).numpy(),
```
lastmansleeping, line 237: @marjanhs Worth noting that original prediction and temperature scaled prediction are the same. It is obvious to us, but might be easier for newer folks to onboard.

python/ml4ir/base/model/calibrations/temperature_scaling.py
```diff
@@ -0,0 +1,247 @@
...
+                'true_label': labels_nps_test,
+                }
+
+    logger.info(f'original test accuracy: {acc_test_org}, \ntemperature scaling '
+                     f'test accuracy: {acc_test_ts} \n')
+
+    # the file is big and should be zipped
+    zip_dir_path = dict_to_zipped_csv(data_dic, logs_dir_local)
```
lastmansleeping, line 245: @marjanhs I don't understand why we need to save this file here. Can you link a reference for me? Sorry :)

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: yes , but let's keep it here as it is not intended for RelevanceModel with `feature_config `and `metadata_features`.

python/ml4ir/base/config/default_model_config.yaml
marjanhs: not sure about this one and how exactly it differs for `calibration`. @lastmansleeping  could u comment on that ?

python/ml4ir/base/model/calibrations/temperature_scaling.py
```diff
@@ -0,0 +1,247 @@
...
+        return tfp.optimizer.lbfgs_minimize(make_val_and_grad_fn(nll_loss_fn),
+                                            initial_position=start)
+
+    # evaluation on validation set before temperature scaling
+    logits_nps, labels_nps = get_logits_labels(intermediate_model, dataset.validation)
+    org_acc_op, org_nll_loss_op, _ = eval_relevance_model(scorer, logits_nps,
+                                                          labels_nps)
+
+    # perform temperature scaling
+    results = run(nll_with_lbfgs, logger)
```
lastmansleeping, line 202: @marjanhs Any reason we can not use the training set here as well?

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: @lastmansleeping  how about `run_optimizer` ?

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: I don't think that can cause a problem as we are creating `tf.constant` from logits  in `nll_loss_fn` https://github.com/salesforce/ml4ir/pull/115/files#diff-5e55239ed1f122b6b40beccad8e8301108acd2e382474342814c595590f776dcR177

Doesn't that freeze the weights ?

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: this function returns nothing

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: I have added the reference on top of the page:
 https://github.com/salesforce/ml4ir/pull/115/files#diff-5e55239ed1f122b6b40beccad8e8301108acd2e382474342814c595590f776dcR177
(the implementations are inspired by the reference)

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: yes, the calibration must be trained on the validation set (unseen data during training)  then  be evaluated on the test data.

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: ok, will add it as a comment

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: @lastmansleeping  we would like to know if the output probabilities of the test set  are well calibrated compared to the original probabilities (before calibration). So, we are saving them to plot reliability diagrams and compute average errors  (ECE) and maximum Error (MCE) later.

python/ml4ir/base/pipeline.py
marjanhs: Thank you @balikasg for the comments. 

- regarding supporting  different methods of calibration, they are a bit more complicated than temperature scaling. For the, we do need to discuss how to engage them in ml4ir. Currently it is part of [model_config](https://github.com/salesforce/ml4ir/pull/115/files#diff-1e0df01beb104b660c3c2805412d4527f3b35fe61e76ecce8d6410101e96fee9R27)

- I personally am not in favor  of adding a separate ExecutionModeKey as calibration can be a post processing step after training or on a trained model for evaluation step. 

- Saving the model with calibration was also my concern. I see two ways to do that : 1- only save temperature (scalar value) and then have a specific load function that joins the saved trained model with temperature 2- save the intermediate model by adding temperature . there might be other ways that worth to discuss (but won't add  for this PR). 

- about saving the scores in HDFS, should I added to the finish?

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: @lastmansleeping  I am concern about any possible side effect when it converts to a graph? is there any? (look at this example : https://www.tensorflow.org/probability/examples/Optimizers_in_TensorFlow_Probability#problem_definition)

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: ok, using it ...

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: no they are not tensors. They (logits) are `numpy` arrays

python/ml4ir/base/model/calibrations/temperature_scaling.py
GhislainBrun:
```suggestion
    data_dic = {'original_scores': [x for x in org_softmaxes.numpy()],
```

GhislainBrun: As a general rule, I would use full names everywhere unless I have a good reason not to.

python/ml4ir/base/model/calibrations/temperature_scaling.py
```diff
@@ -0,0 +1,247 @@
...
+                                           labels_nps_test, temperature=temper)
+
+    # write the full vector in the csv not ...
+    np.set_printoptions(
+        formatter={'all': lambda x: str(x.decode('utf-8')) if isinstance(x, bytes) else str(x)},
+        linewidth=sys.maxsize,
+        threshold=sys.maxsize)
+    # avoiding .tolist() that is not memory efficient
+    data_dic = {'org_scores': [x for x in org_softmaxes.numpy()],
+                'ts_scores': [x for x in ts_softmaxes.numpy()],
```
GhislainBrun, line 236:
```suggestion
                'temperature_scaling_scores': [x for x in temperature_scaling_softmaxes.numpy()],
```

python/ml4ir/base/pipeline.py
marjanhs: (btw I will raise another PR to add it to the finish)

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: but added here , too.

python/ml4ir/base/pipeline.py
lastmansleeping: I'm most interested in the model saving part. I am not comfortable with how we are doing it right now. But let's look at it in a follow up PR.

python/ml4ir/base/config/default_model_config.yaml
lastmansleeping: Should be fine for now. We might want to rename these things like @balikasg mentions later.

python/ml4ir/base/model/calibrations/temperature_scaling.py
lastmansleeping: Haha, yes. I was going to say this as well, but avoided because everything else had an `s` at the end. I like `calibration` better too.

python/ml4ir/base/model/relevance_model.py
lastmansleeping: This is mostly in hopes that we can have something generic that we can ship with `RelevanceModel`. But if we realize that it doesn't work, we can move it. I like the idea of having a common base implementation and overrides if necessary.

python/ml4ir/base/model/calibrations/temperature_scaling.py
lastmansleeping: Yes, I noticed that after I left this comments. Forgot to update this. I think your current implementation works because you use `tf.constant` on the logits. But I fear it is not going to be very optimal on a large dataset. I recommend leaving it as is for now and we should sync up about effort needed to use `RelevanceModel` directly with freezing all the learned weights and then modifying the last layer with temperature. This has a lot of advantages like directly saving the TFRecord signature with the temperature scaling embedded into it. We can come back to this though.

python/ml4ir/base/model/calibrations/temperature_scaling.py
lastmansleeping: Definitely worth documenting the data type in the docstrings.

python/ml4ir/base/model/calibrations/temperature_scaling.py
lastmansleeping: Works for me.

python/ml4ir/base/model/calibrations/temperature_scaling.py
lastmansleeping: Okay. That makes sense. Thanks!

Do you plan to add the plots and metrics to ml4ir as well? I think it might be useful? Otherwise, I'm assuming the current workflow involves training a model and then running a post-eval script? That should be fine too, as long as you check it in somewhere. ;)

lastmansleeping: I recommend doing this in a follow up PR. Can be in one of our internal repos as well if you think it contains salesforce-specific info.

lastmansleeping: @balikasg thoughts?

python/ml4ir/base/model/calibrations/temperature_scaling.py
lastmansleeping: Sounds good. Thanks!

python/ml4ir/base/model/calibration/temperature_scaling.py
```diff
@@ -0,0 +1,328 @@
...
+
+    # perform temperature scaling
+    results = run_optimizer(nll_with_lbfgs, logger)
+
+    # evaluation on validation set with temperature scaling
+    temper = tf.constant(results.position)
+    acc_op, _, _ = eval_relevance_model(scorer, logits_nps, labels_nps, temperature=temper)
+
+    logger.info("=" * 50)
+    logger.info(f'temperature value : {results.position}')
```
lastmansleeping, line 290: @marjanhs Can you add the temperature value to the `experiment_tracking_dict` https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/base/pipeline.py#L390 ?

python/ml4ir/applications/classification/tests/test_calibration.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+
+
+
+
+
+
+
+
```
lastmansleeping, line 91: [Minor] @marjanhs Can you remove these trailing new lines?

python/ml4ir/applications/classification/tests/test_calibration.py
```diff
@@ -0,0 +1,91 @@
...
+        labels_tensor = tf.constant(labels, name='logits_test', dtype=tf.float32)
+
+        acc = accuracy(scores_tensor, labels_tensor)
+        self.assertEqual(acc, 0.5)
+
+    def test_temperature_scaling(self):
+        """Tests temperature scaling """
+
+        # computing logits before temperature scaling of the validation set
+        logits_nps, labels_nps = get_logits_labels(self.classification_model.model,
```
lastmansleeping, line 47: @marjanhs What is `_nps`?

python/ml4ir/applications/classification/tests/test_calibration.py
```diff
@@ -0,0 +1,91 @@
...
+        # Tests the learned temperature parameter
+        expected_value = 1.55
+        atol = 0.1
+        self.assertTrue(np.isclose(results.position[0], expected_value, atol=atol))
+
+        # computing logits after temperature scaling of the validation set
+        logits_nps_w_ts, labels_nps_w_ts = get_logits_labels(self.classification_model.model,
+                                                   self.relevance_dataset.validation)
+
+        # Tests if the model weights were frozen and not affected by temperature scaling
```
lastmansleeping, line 67: @marjanhs Perfect. Thank you!

lastmansleeping (commented):
The PR looks good and ready to be merged after a few minor comments are addressed. I have one main concern though. How do we plan to use this? The results out of the `RelevanceModel` will be the activated scores and not the logits in the SavedModel. We can use the CSV file with pre post scores to measure calibration error and confidence, but we can't really ship this model yet, right? In the follow up PRs I propose we tackle the following:
* Saving the model using `RelevanceModel.save()` - this would require us to rewire the temperature term into the model's last layer. It should be doable IMO. We can discuss details offline if needed. (Maybe we can get rid of the Keras wrapper model as well, not sure yet).
* Add script/functions used to measure calibration metrics and plot the charts.

@marjanhs

python/ml4ir/applications/classification/tests/test_calibration.py
marjanhs: numpys

marjanhs: using the full names ...

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: yes, have added : https://github.com/salesforce/ml4ir/pull/115/files#diff-bf0198f6577050795b504acc0a05fccc56e8942b07df2a0497c4cc467bfefd39R172-R174

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: @lastmansleeping  we can add them to ml4ir, but I don't think it is a priority as the code(in pytorch) is ready and can be used independently of `Tensorflow`. It only needs the `.csv` file we are generating here. I should add the code to an internal repo for now (and if gets priority convert it to Tensorflow  and add it here).

python/ml4ir/base/model/calibration/temperature_scaling.py
marjanhs: added

python/ml4ir/base/pipeline.py
```diff
@@ -407,10 +403,17 @@ def run(self):
...
-                        relevance_model.calibrate(relevance_dataset=relevance_dataset,
-                                                  logger=self.logger,
-                                                  logs_dir_local=self.logs_dir_local,
-                                                  temperature_init=temperature)
+                        results = relevance_model.calibrate(relevance_dataset=relevance_dataset,
+                                                            logger=self.logger,
+                                                            logs_dir_local=self.logs_dir_local,
+                                                            temperature_init=temperature)
+
+                        experiment_tracking_dict.update({CalibrationKey.TEMPERATURE:
```
marjanhs, line 411: adding temperature value

python/ml4ir/base/pipeline.py
```diff
@@ -407,10 +403,17 @@ def run(self):
...
+                                                             results.position[0]})
+
```
marjanhs, line 413: adding temperature value above

python/ml4ir/base/model/relevance_model.py
```diff
@@ -780,12 +781,32 @@ def _build_callback_hooks(
...
+            probabilities, calibrated probabilities, ...)
+        temperature_init: float
+            temperature initial value
+        Returns
+        -------
+        `Union[np.ndarray, Tuple[np.ndarray, ...]]`
+        optimizer output containing temperature value learned during temperature scaling
+
+        """
+        logger.info("=" * 50)
```
marjanhs, line 804: adding missing docstring

python/ml4ir/base/model/relevance_model.py
```diff
@@ -18,7 +20,6 @@
 from ml4ir.base.model.callbacks.debugging import DebuggingCallback
 from ml4ir.base.model.calibration.temperature_scaling import temperature_scale
 
-from typing import Dict, Optional, List, Union, Type
```
marjanhs, line 20: change the order to comply with Pylint

python/ml4ir/base/pipeline.py
jakemannix: I'm also ok with a followup item tracking the scaling as part of the model, but it does need to be done to make this fully functional.  So we should annotate this clearly in the docs (docstring and README), and then file a followup work item to track saving + serving this temperature scaling wrapper.

python/ml4ir/base/model/calibrations/temperature_scaling.py
jakemannix: post-eval in pytorch? What does this post-eval script do?

python/ml4ir/base/model/calibrations/temperature_scaling.py
jakemannix: I thought we have to separate out a _totally_ unseen dataset for calibration?   Meaning: we train the main model on a `train` set, with per-epoch `validation`, then finally express the offline quality on a `test` dataset.  *Then* we need a totally independent `calibration` set, right?

lastmansleeping (approved):
LGTM. Thanks for all the work @marjanhs 
Let's work on the two items mentioned in my previous comment in follow up PRs.

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: @jakemannix  it computes the  inputs of reliability diagram, ECE and MCE .

python/ml4ir/base/model/calibrations/temperature_scaling.py
marjanhs: @jakemannix  here is the pipeline:
we train  a (classification) model (let's call it model A) using `training set` and validate it on the `validation set`. Actually,
the `validation set` is being used for **training** the temperature parameter during temperature scaling. Here is the reference:[https://github.com/gpleiss/temperature_scaling/blob/master/train.py#L174]( https://github.com/gpleiss/temperature_scaling/blob/master/train.py#L174)
Then, we examine the calibrated model (model B = trained model A + temperature applied on logits) on the `test set` . We save the probability scores of model A and model B of the `test set `to  perform  analysis.

python/ml4ir/base/model/calibration/temperature_scaling.py
```diff
@@ -0,0 +1,329 @@
...
+        formatter={'all': lambda x: str(x.decode('utf-8')) if isinstance(x, bytes) else str(x)},
+        linewidth=sys.maxsize,
+        threshold=sys.maxsize)
+    # avoiding .tolist() that is not memory efficient
+    # note: temperature scaling does not change the accuracy as it does not change the maximum. So,
+    # the temperature scaling predicted labels must be the same as  original
+    # predicted labels: `original_predicted_label`
+    data_dic = {'original_scores': [x for x in original_softmaxes.numpy()],
+                'temperature_scaling_scores': [x for x in temperature_scaling_softmaxes.numpy()],
+                'original_predicted_label': tf.argmax(original_softmaxes, axis=-1).numpy(),
```
balikasg, line 319: Perhaps calling `scores` -> `probabilities` for future? Thanks for the changes though!

python/ml4ir/base/model/relevance_model.py
```diff
@@ -778,3 +780,33 @@ def _build_callback_hooks(
         # Add more here
 
         return callbacks_list
+
+    def calibrate(self, relevance_dataset, logger, logs_dir_local, temperature_init)\
```
balikasg, line 784: To me it is still striking that we are defining with to the base relevance model and we explicitly define a `NotImplemented` in the ranking side to bypass this. It suggests that this function is not generic enough for the base model and should live in the classification model @marjanhs @lastmansleeping .

python/ml4ir/base/pipeline.py
balikasg: @marjanhs we are saying the same thing I believe. All the options you lay down simply suggest that the user should explicitly say 'do-calibration', and the model-config is not ideal because it describes the model architecture. I see there is a consensus for reworking  the saving part, we can discuss there with any experience and feedback we gather from executing the code at scale.

balikasg (approved):
LGTM! A nice feature for a great project for an awesome new contributor!

PR #116, opened 2021-02-26
Reverting all changes from SequenceExample restructuring

lastmansleeping:
Reverting all the changes made in https://github.com/salesforce/ml4ir/pull/91

Most of the revert were straightforward. I got stuck with inconsistencies and debugging weird shape errors. But at the end it is a clean swap.

We should really *really* re-revert this once we add the end to end tests because this is the intended use of SequenceExamples and also gives us more freedom to play with the tensor shapes.


lastmansleeping:
@ducouloa @jakemannix I should probably bump the version to `0.1.0`. Thoughts?

lastmansleeping:
@marjanhs I've removed calibration from the default model config.

python/ml4ir/applications/classification/tests/data/configs/model_config.yaml
```diff
@@ -16,4 +16,7 @@ layers:
     rate: 0.0
   - type: dense
     name: final_dense
-    activation: null
\ No newline at end of file
+    activation: null
+calibration:
```
lastmansleeping, line 20: @marjanhs I've removed it from the default config and added it here. You can test it using ->
```
python3 ml4ir/applications/classification/pipeline.py \
--batch_size 32 \
--run_id test \
--data_format tfrecord \
--execution_mode train_inference_evaluate
```

jakemannix:
> @ducouloa @jakemannix I should probably bump the version to `0.1.0`. Thoughts?

I think that could make sense, yes.  And then when we re-integrate this data shape, we bump to 0.2?

lastmansleeping:
> > @ducouloa @jakemannix I should probably bump the version to `0.1.0`. Thoughts?
> 
> I think that could make sense, yes. And then when we re-integrate this data shape, we bump to 0.2?

Yep, that's what I was thinking.

ducouloa (approved):
From what I can tell, this looks good to me.

I've only the following interrogations:
- I'm wondering what those `file_0.tfrecord` empty files are.
- If I had to implement the serving side of the model, how should I proceed to format my `tf_record` correctly? Where is the specification of the interface?

lastmansleeping:
@ducouloa Thanks for the approval.

> If I had to implement the serving side of the model, how should I proceed to format my tf_record correctly? Where is the specification of the interface?

You should be able to use the current implementation of TFRecord on the inference side. That was the goal of the revert, right?

@jakemannix should be able to help you better with the scala implementation. But you can find the python side implementation in `tfrecord_helper` which converts a dataframe into a TFRecord proto.

lastmansleeping:
@ducouloa 

> I'm wondering what those file_0.tfrecord empty files are.

They are not empty. They are binary files using the reverted SequenceExample format. You can either regenerate them by running ml4ir-ranking with CSV input (which will create intermediate TFRecord files). For this PR, to be super careful, I copied over files from the commit before my original "breaking" PR itself.

jakemannix:
> @ducouloa Thanks for the approval.
> 
> > If I had to implement the serving side of the model, how should I proceed to format my tf_record correctly? Where is the specification of the interface?
> 
> You should be able to use the current implementation of TFRecord on the inference side. That was the goal of the revert, right?
> 
> @jakemannix should be able to help you better with the scala implementation. But you can find the python side implementation in `tfrecord_helper` which converts a dataframe into a TFRecord proto.
> 
> To look at the current TFRecord format, I recommend looking at an example. You can load the test data as below and compare with similar log from the JVM code as well.
> 
> ```
> import tensorflow as tf
> proto = next(iter(tf.data.TFRecordDataset("/path/to/ml4ir/python/ml4ir/applications/ranking/tests/data/tfrecord/train/file_0.tfrecord")))
> se = tf.train.SequenceExample()
> se.ParseFromString(proto.numpy())
> se
> ```

This python example code sadly doesn't help a JVM programmer, @lastmansleeping  :(   Loading the test data in python, shows stuff we don't see in the JVM (python serialization).

But @ducouloa - @lastmansleeping is right: reverting this should allow us to make *no* changes to the JVM side, to set up an integration test: do the same thing we did on the classification code for an integration test.

ducouloa:
Yes, I get the point about getting back to a JVM/Python parity.

My question was more about the documentation/specification process we can have in the case an external person want to create his own serving (let's say in GO, any other language for example), and how we can specify the interface when we will re-introduce this change and adapt the JVM part.

lastmansleeping:
> Yes, I get the point about getting back to a JVM/Python parity.
> 
> My question was more about the documentation/specification process we can have in the case an external person want to create his own serving (let's say in GO, any other language for example), and how we can specify the interface when we will re-introduce this change and adapt the JVM part.

Ideally, there shouldn't be a case where a user would want to redefine the JVM TFRecord logic once we have end-to-end tests and re-reverted. That's the standard TFRecord implementation. What we have currently is not too far off. It is just limiting in some ways. I have been planning to post a blog for a while, maybe I should do it. There is very little documentation available around this stuff.

jakemannix:
`There is very little documentation available around this stuff.` - all the more reason why we need to document how this looks, at a structural level, and then how that logical structure translates into the python and java code.

As for someone trying to implement inference in Go, or some other language, it's going to be a lot more work than just how they build the TFRecord protobufs, as we don't currently expose a gRPC service for this library, so they'd have to make sure whatever Tensorflow inference wrapper library they use knows how to basically do what we're doing in https://github.com/salesforce/ml4ir/blob/master/jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/TFRecordExecutor.scala , otherwise the TFRecord object isn't going to help them.

PR #119, opened 2021-03-19
adding temperature scaling layer and saving the model with TS

marjanhs:
This PR is a **draft**  to add the learned temperature parameter to the model and save the model with temperature.

### DIFF
--- a/python/ml4ir/applications/classification/tests/test_calibration.py
+++ b/python/ml4ir/applications/classification/tests/test_calibration.py
@@ -86,6 +86,25 @@ class TestCalibration(ClassificationTestBase):
     def test_add_temperature_layer(self):
         """Tests whether adding temperature scaling layer scales the logits as expected """
 
+        # Tests whether temperature = 1.0 has no effect on outputs
+        output_original = self.classification_model.model.predict(
+            self.relevance_dataset.validation).squeeze()
+
+        temperature = 1.0
+        self.classification_model.add_temperature_layer(temperature=temperature,
+                                                        layer_name=TEMPERATURE_LAYER_NAME)
+
+        output_calibration = self.classification_model.model.predict(
+            self.relevance_dataset.validation).squeeze()
+
+        rtol = 1e-6
+        np.testing.assert_allclose(output_original, output_calibration, rtol=rtol,
+                                   err_msg="outputs differs between models with and "
+                                           "without temperature = 1.0 ")
+
+        # refreshing relevance model
+        self.classification_model = self.classification_pipeline.get_relevance_model()
+
         # getting logits of val. set from original model (without temperature scaling)
         model_wo_ts = get_intermediate_model(self.classification_model.model,
                                                   self.classification_model.scorer)
@@ -132,7 +151,7 @@ class TestCalibration(ClassificationTestBase):
                          msg="difference between number of layers of models with and without "
                              "temperature should be one")
 
-        logits_before_saving = self.classification_model.model.predict(
+        outputs_before_saving = self.classification_model.model.predict(
             self.relevance_dataset.validation).squeeze()
 
         # saving and loading the model with TS layer
@@ -159,13 +178,13 @@ class TestCalibration(ClassificationTestBase):
                          msg="number of layers for model with temperature before and after "
                              "saving differs!")
 
-        logits_loaded_model = self.classification_model.model.predict(
+        ouputs_loaded_model = self.classification_model.model.predict(
             self.relevance_dataset.validation).squeeze()
 
         # tests whether the saved model and loaded model predicts same results
         rtol = 1e-6
-        np.testing.assert_allclose(logits_before_saving, logits_loaded_model, rtol=rtol,
-                                   err_msg="logits differs between models with temperature before"
+        np.testing.assert_allclose(outputs_before_saving, ouputs_loaded_model, rtol=rtol,
+                                   err_msg="Outputs differ between models with temperature before"
                                            " and after saving!")
 
 
--- a/python/ml4ir/applications/classification/tests/test_calibration.py
+++ b/python/ml4ir/applications/classification/tests/test_calibration.py
@@ -6,20 +6,22 @@ import pandas as pd
 import tensorflow as tf
 import numpy as np
 
-from ml4ir.base.model.calibration.temperature_scaling import dict_to_zipped_csv, \
-    TEMPERATURE_SCALE, accuracy, get_logits_labels, temperature_scale
+from ml4ir.base.model.calibration.temperature_scaling import dict_to_csv, \
+    TEMPERATURE_SCALE, accuracy, get_logits_labels, temperature_scale, get_intermediate_model
 from ml4ir.applications.classification.tests.test_base import ClassificationTestBase
+
+TEMPERATURE_LAYER_NAME = 'temperature_layer'
 
 
 class TestCalibration(ClassificationTestBase):
     """Class to test temperature scaling from `ml4ir.base.model.calibration.temperature_scaling` """
 
-    def test_dict_to_zipped_csv(self):
+    def test_dict_to_csv(self):
         """Tests if the .zip file has been created and contains .csv file.
          It also tests if .csv file contains true values"""
 
         data_dict = {'feature': [1, 2.0, 3], 'labels': [0, 0, 1]}
-        dict_to_zipped_csv(data_dict, self.output_dir, self.file_io)
+        dict_to_csv(data_dict, self.output_dir, self.file_io, zip_output=True)
 
         filename_zip = os.path.join(self.output_dir, TEMPERATURE_SCALE)
         filename_csv = os.path.join(self.output_dir, TEMPERATURE_SCALE, f'{TEMPERATURE_SCALE}.csv')
@@ -38,15 +40,15 @@ class TestCalibration(ClassificationTestBase):
         labels_tensor = tf.constant(labels, name='logits_test', dtype=tf.float32)
 
         acc = accuracy(scores_tensor, labels_tensor)
-        self.assertEqual(acc, 0.5)
+        self.assertEqual(acc, 0.5, msg="accuracy function does not work as expected")
 
     def test_temperature_scaling(self):
         """Tests temperature scaling """
 
-        # computing logits before temperature scaling of the validation set
+        # Computing logits before temperature scaling of the validation set
         logits_numpys, labels_numpys = get_logits_labels(self.classification_model.model,
                                                          self.relevance_dataset.validation)
-        # sanity check the shape of logits and labels
+        # Sanity check the shape of logits and labels
         self.assertTrue(len(logits_numpys) == len(labels_numpys))
 
         results = temperature_scale(self.classification_model.model,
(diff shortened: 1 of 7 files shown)


python/ml4ir/base/model/relevance_model.py
```diff
@@ -602,14 +603,16 @@ def save(
...
+            sub directory name to save the model into
 
         Notes
         -----
         All the functions passed under `preprocessing_keys_to_fns` here must be
         serializable tensor graph operations
         """
 
-        model_file = os.path.join(models_dir, "final")
+        model_file = os.path.join(models_dir, sub_dir)
```
marjanhs, line 615: add the flexibility to save any model (original model, calibrated model, ...)

python/ml4ir/base/pipeline.py
```diff
@@ -410,6 +410,27 @@ def run(self):
...
+                        # the following raises error , both for original model (without TS) and
+                        # calibrated model
+                        # ValueError: You tried to call `count_params` on query,
+                        # but the layer isn't built.
+                        # You can build it manually via: `query.build(batch_input_shape)
+
+                        #path = os.path.join(self.models_dir_local, 'final_calibrated')
+                        #path = os.path.join(path, 'default')
+                        #m = relevance_model.load(model_file=path)
+                        #print(m.summary())
```
marjanhs, line 433: Apparently there is a problem with model loading even for the original model (a known issue)

python/ml4ir/base/model/relevance_model.py
```diff
@@ -810,3 +813,44 @@ def calibrate(self, relevance_dataset, logger, logs_dir_local, temperature_init)
...
+        """
+
+        # get  last layer's output  --> MUST **NOT** BE AN ACTIVATION (e.g. SOFTMAX) LAYER
+        final_layer_name = self.scorer.model_config['layers'][-1]['name']
+
+        final_layer = self.model.get_layer(name=final_layer_name).output
+        temperature_layer = TemperatureScalingLayer(name='temperature_layer',
+                                                    temperature=temperature)(final_layer)
+
+        # using the `last Activation layer` as final activation function before computing loss
```
marjanhs, line 839: we expect the last layer to be an activation layer. Otherwise, do nothing!

python/ml4ir/base/model/calibration/temperature_scaling.py
```diff
@@ -324,6 +326,47 @@ def nll_with_lbfgs():
                      f'test accuracy: {acc_test_temperature_scaling} \n')
 
     # the file is big and should be zipped
-    zip_dir_path = dict_to_zipped_csv(data_dic, logs_dir_local, file_io)
-    logger.info(f"Read and zipped  {zip_dir_path}")
+    #zip_dir_path = dict_to_zipped_csv(data_dic, logs_dir_local, file_io)
+    #logger.info(f"Read and zipped  {zip_dir_path}")
```
marjanhs, line 330: takes too long to zip.. commenting for now

balikasg: Fair. But you should be trying with smaller data instead! 

Tip: you should start with a test for this functionality (you know we will ask for a test ;-)). This test will take you some time to write (but you will commit it too) and it will check that the predictions are the same before saving and after saving. Since you are training every time set the seeds and use one epoch. Also, this way you will be able to develop locally (I hope that you are prototyping this locally and don't use splunk to debug as it is a huge pain). There already temp data for the test..

python/ml4ir/base/model/calibration/temperature_scaling.py
```diff
@@ -324,6 +326,47 @@ def nll_with_lbfgs():
...
+
+    def get_config(self):
+        """
+        Returns a Python dict containing the configuration of the layer
+        """
+        config = super(TemperatureScalingLayer, self).get_config()
+        config.update({"temperature": self.temperature})
+        return config
+
+    def call(self, input):
```
balikasg, line 360: Elegant!

python/ml4ir/base/model/calibration/temperature_scaling.py
```diff
@@ -324,6 +326,47 @@ def nll_with_lbfgs():
...
+        """
+        Scales the input and computes input/temperature.
+        Parameters
+        ---------
+        input: `Tensor` object
+            Tensors (should be logits) as nominator in scaling operation.
+
+        Returns
+        ------
+        scaled input w.r.t temperature parameter (input/temperature).
```
balikasg, line 370: I love this docstring!

python/ml4ir/base/model/calibration/temperature_scaling.py
```diff
@@ -324,6 +326,47 @@ def nll_with_lbfgs():
...
 
     # the file is big and should be zipped
-    zip_dir_path = dict_to_zipped_csv(data_dic, logs_dir_local, file_io)
-    logger.info(f"Read and zipped  {zip_dir_path}")
+    #zip_dir_path = dict_to_zipped_csv(data_dic, logs_dir_local, file_io)
+    #logger.info(f"Read and zipped  {zip_dir_path}")
     return results
+
+
+class TemperatureScalingLayer(tf.keras.layers.Layer):
```
balikasg, line 334: This is a nice, descriptive name!

python/ml4ir/base/model/relevance_model.py
balikasg: Does this break any of the pipelines we already have? Since you are adding this to the `base` class it will propagate to all models.

python/ml4ir/base/model/relevance_model.py
```diff
@@ -810,3 +813,44 @@ def calibrate(self, relevance_dataset, logger, logs_dir_local, temperature_init)
                                  logs_dir_local=logs_dir_local,
                                  temperature_init=temperature_init,
                                  file_io=self.file_io)
+
+    def add_temperature_layer(self, temperature: float):
```
balikasg, line 817: I am happy you are adding all these. I will repeat something I have insisted in other reviews though. The `base` pipelines and the `base models` are base for a reason: they contain things shared between classification and ranking models. This way we achieve DRY. Functionality that respects this is welcome. If not, it is debt in a sense. 

Now, temperature scaling is not something shared. It only works with classification and [we have a classification model already](https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/applications/classification/model/classification_model.py). This functionality should live there! Is there any reason why you believe this is a better place? 

BTW, see how you are notifying the user with inline comments to be aware for this.

python/ml4ir/base/model/relevance_model.py
```diff
@@ -810,3 +813,44 @@ def calibrate(self, relevance_dataset, logger, logs_dir_local, temperature_init)
...
+
+        # get  last layer's output  --> MUST **NOT** BE AN ACTIVATION (e.g. SOFTMAX) LAYER
+        final_layer_name = self.scorer.model_config['layers'][-1]['name']
+
+        final_layer = self.model.get_layer(name=final_layer_name).output
+        temperature_layer = TemperatureScalingLayer(name='temperature_layer',
+                                                    temperature=temperature)(final_layer)
+
+        # using the `last Activation layer` as final activation function before computing loss
+        i = 1
```
balikasg, line 840: you should use a more descriptive variable name here, what you are going to do in the next lines is not trivial.

balikasg (changes requested):
thanks! I have left a few comments, please take a look! 
We need a test for this too! We actually need several tests (one python test and one integration test to validate predictions are equivalent between java too), but let's start with one.

python/ml4ir/base/model/relevance_model.py
marjanhs: It should not break as I have set the default value to `'final'`. See line 575:
`sub_dir: str = 'final'`

python/ml4ir/base/model/calibration/temperature_scaling.py
marjanhs: yes It works nice in my local box and with small data. I had added the test for zipping part before . See [here.](https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/applications/classification/tests/test_calibration.py#L17)
Apparently, it gets stuck with real (big) data!

python/ml4ir/base/model/relevance_model.py
marjanhs: yes I agree with you, we do not have calibration with TS for classification models currently. The only reason for  this to be in base class was to have the placeholder for ranking models, too. (we desire to have calibrator for rankers and may have them in future , of course not now)

python/ml4ir/base/model/calibration/temperature_scaling.py
```diff
@@ -112,8 +112,8 @@ def accuracy(scores, labels) -> np.ndarray:
     return acc.result().numpy()
 
 
-def dict_to_zipped_csv(data_dic: dict, root_dir: str, file_io: FileIO) -> str:
-    """Saves input dictionary to a zipped csv file
+def dict_to_csv(data_dic: dict, root_dir: str, file_io: FileIO, zip_output: bool = True) -> str:
+    """Saves input dictionary to a csv file and zips if requested
```
marjanhs, line 116: adding the flexibility of whether zip the output or not

python/ml4ir/base/model/calibration/temperature_scaling.py
```diff
@@ -344,11 +352,25 @@ def __init__(self, temperature, name='temperature', **kwargs):
...
+        ----------
+        input_shape: tuple
+            a tuple showing layer input shape
+        """
+        self.temperature_var = self.add_weight("temperature",
+                                               shape=[1],
+                                               dtype=tf.float32,
+                                               initializer=tf.keras.initializers.
+                                               Constant(value=self.temperature))
+
```
marjanhs, line 373: defining temperature as a model parameter so that its value be saved/loaded using  `save()` and `load_model() ` tf functions

python/ml4ir/base/pipeline.py
```diff
@@ -399,19 +399,20 @@ def run(self):
...
                     if self.model_config[CalibrationKey.CALIBRATION]['key'] == \
                             CalibrationKey.TEMPERATURE_SCALING:
-                        temperature = self.model_config[CalibrationKey.CALIBRATION][
-                            CalibrationKey.TEMPERATURE] if \
-                            CalibrationKey.TEMPERATURE in\
-                            self.model_config[CalibrationKey.CALIBRATION] else 1.5
+                        kwargs = self.model_config[CalibrationKey.CALIBRATION][
+                            CalibrationKey.ARGS] if CalibrationKey.ARGS in self.model_config[
+                            CalibrationKey.CALIBRATION] else {}
+
```
marjanhs, line 405: getting and passing all arguments related to temperature scaling

python/ml4ir/applications/classification/tests/test_calibration.py
```diff
@@ -6,20 +6,22 @@
 import tensorflow as tf
```
marjanhs, line 6: two new tests for model saving/loading with TS added to this file and failure messages added to the previous tests

marjanhs:
@balikasg  @Ullimague 
- review feedback applied
- python unit tests added
next step is adding integration test. Could you review the new stuff meanwhile ?
Thanks!

python/ml4ir/base/model/calibration/temperature_scaling.py
```diff
@@ -19,8 +19,10 @@
 import tensorflow_probability as tfp
 
 from ml4ir.base.data.relevance_dataset import RelevanceDataset
+#from ml4ir.base.model.relevance_model import RelevanceModel
```
Ullimague, line 22: Don't forget to remove it if needed.

python/ml4ir/base/model/relevance_model.py
```diff
@@ -808,5 +809,47 @@ def calibrate(self, relevance_dataset, logger, logs_dir_local, temperature_init)
...
+                     a scalar value to scale the last activation layer inputs
+        layer_name: str
+            name of the temperature scaling layer
+        Returns
+        -------
+        `RelevanceModel`
+        updated RelevanceModel object with temperature
+        """
+
+        # get  last layer's output  --> MUST **NOT** BE AN ACTIVATION (e.g. SOFTMAX) LAYER
```
Ullimague, line 832: But isn't typically our last layer an activation layer such as a softmax?
What would happen if that's the case?

python/ml4ir/base/model/relevance_model.py
```diff
@@ -808,5 +809,47 @@ def calibrate(self, relevance_dataset, logger, logs_dir_local, temperature_init)
                                  dataset=relevance_dataset,
                                  logger=logger,
                                  logs_dir_local=logs_dir_local,
-                                 temperature_init=temperature_init,
-                                 file_io=self.file_io)
+                                 file_io=self.file_io,
+                                 **kwargs)
+
+    def add_temperature_layer(self, temperature: float = 1.0, layer_name: str = 'temperature_layer'):
```
Ullimague, line 815: Just to double check, the default value temperature: float = 1.0 doesn't have any effect, right?

balikasg: A reasonable test ;-)

python/ml4ir/base/pipeline.py
```diff
@@ -399,17 +399,29 @@ def run(self):
...
                         results = relevance_model.calibrate(relevance_dataset=relevance_dataset,
                                                             logger=self.logger,
                                                             logs_dir_local=self.logs_dir_local,
-                                                            temperature_init=temperature)
+                                                            **kwargs)
```
balikasg, line 409: we went from something clear and explicit to `**kwargs` here. What is the benefit?

balikasg: `zip_output`? explicit is better than implicit. Let's change this back.

python/ml4ir/base/pipeline.py
```diff
@@ -399,17 +399,29 @@ def run(self):
...
+                        # temperature scaling layer
+                        relevance_model.add_temperature_layer(results.position[0])
+                        # saving calibrated (with temperature scaling layer) model
+                        relevance_model.save(
+                            models_dir=self.models_dir_local,
+                            preprocessing_keys_to_fns={},
+                            postprocessing_fn=None,
+                            required_fields_only=not self.args.use_all_fields_at_inference,
+                            pad_sequence=self.args.pad_sequence_at_inference,
+                            sub_dir='final_calibrated'
```
balikasg, line 423: this has an impact on the downstrean jobs that use the 'final' signature. Will this code produce two sub-dirs?

balikasg (changes requested):
thanks for the tests, huge progress. Some comments left. Do we need to update the Readme anywhere?

python/ml4ir/base/model/relevance_model.py
balikasg: But yes, jokes apart, it should have no effect. Although, thinking twice it is not a very silly test.

python/ml4ir/base/model/relevance_model.py
balikasg: @marjanhs I am happy you agree! Still, it remained in `base`! This function is not a placeholder though. A lot of the assumptions the code does apply only to classification models. Therefore imho it is not generic to be in base. To give you another example, in your imports 
```python
from ml4ir.base.model.calibration.temperature_scaling import temperature_scale,\
    TemperatureScalingLayer
```
It would be saner to have this as: 
```python
from ml4ir.applications.classification.calibration.temperature_scaling import temperature_scale,\
    TemperatureScalingLayer
```

balikasg: that said, it is not the most important issue here perhaps, a lot of nice things happened in the revision below.

python/ml4ir/base/model/relevance_model.py
marjanhs: in our implementation we add the final activation (softmax) as a separate layer in the last layer. 
NN: Input, Dense1, Dense2, Activation ----- adding temperature scaling layer-----> NN: Dense1, Dense2, TS layer, Activation
I create a totally new functional API Model from the existing Model to add a new layer in a right place. I do not want to loose any layer and in the above example I do want Dense2 (without softmax) be right before TS layer because I want to scale the output of Dense2.

python/ml4ir/base/model/relevance_model.py
marjanhs: @Ullimague  yes, T=1.0 has no effect because of  X/1.0==X

python/ml4ir/base/pipeline.py
marjanhs: @balikasg  I should add more args to the temperature scaling layer in future, specifically for its optimizer. if I do not use **kwargs at the top level ,`calibrate()` in `pipeline.py `and `temperature_scale()` in `relevance_model.py`, I have to pass tons of args to these function explicitly with redundant docstrings. However,  all the temperature scaling args have been explicitly defined in its right place  [here ](https://github.com/salesforce/ml4ir/blob/e8aabc6eb57a4fd3807933ebe1f1fb05d2a9e776/python/ml4ir/base/model/calibration/temperature_scaling.py#L222) and we have all of them there.

python/ml4ir/base/pipeline.py
marjanhs: @balikasg  yes, pipeline.py (for example for training)  generates two sub-directories: 1- the default `final` directory to save the original model without temperature scaling ( I have changed the hardcoded `final` in the `save` function to sub_dir=`final` as a input argument. 2- the `final_calibrated` model that is a model with TS  layer

python/ml4ir/base/model/relevance_model.py
marjanhs: yes, I am agree with u that calibration works for classification.another reason is that because the ranking model is kind of a classification, we can easily try calibration with a little tweaks with current implementation or implement a totally new calibration for ranking.  That's why I have still kept it in the base.

marjanhs:
Thank you @Ullimague @balikasg for your reviews.  Please see my answers to your comments.  I also updated the readme file (`model_config.md`) and cleaned up the code.  This PR is no longer a draft and ready for final review. For integration test I should create a follow-up WI and PR.

python/ml4ir/base/model/relevance_model.py
Ullimague: Ok that makes sense, just to double check, it is currently working under the assumption we only have a softmax as our last activation function, right? Would we have to revisit it if we consider something else then https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/applications/classification/config/keys.py#L7 (CATEGORICAL_CROSS_ENTROPY) ?

python/ml4ir/base/model/relevance_model.py
Ullimague: It would be actually great to have such a test. And pretty trivial to add?

Ullimague: I'm actually curious to know, if we train 2 same models (and with the same random seeds), one without temperature, and one with temperature T=1. Are the predictions the same?

marjanhs: the prediction must be the same ; I added a test for temperature = 1.0 and compared the outputs of tho models (one with temperature =1.0 and the other without temperature) . Please see the test [here](https://github.com/salesforce/ml4ir/pull/119/commits/7fa0c7fcb7cb16b1c57c5a99144ef9e2a8d13903#diff-c604f786e9419883b3fdc3e0d13d6fa60b85ec7f2c8cb129dcd213abe5d7bac9R89)

python/ml4ir/base/model/relevance_model.py
marjanhs: sigmoid/softmax/or any Activation for the last output should be fine. The thing is that the activation should produce a probability vector (sum to 1) otherwise adding calibration does not make sense in theory because TS  is supposed to calibrate the classification probabilities.

Ullimague: We should probably add that as comment/doc somewhere :))

marjanhs: @Ullimague  yes, definitely. added here :
https://github.com/salesforce/ml4ir/pull/119/commits/afa20e84d414f72e33cc4330be6e9c5df8584932#diff-76516993b9207e862e3d127124d5549728d65b4ff22fc829dc4770226eae0cb4R29

Ullimague (approved):

balikasg (approved):
Thank you for the effort and for taking care of the comments! Great stuff!

PR #121, opened 2021-04-29
Another test harness, raised from the dead

jakemannix:
TL;DR: took an old branch with some test code, updated it with master, trained a new dummy ranking model in python, wired that into the jvm test harness, and it is not crashing, and in fact getting scores within 1% of each other from python / scala (usually much closer than that).

@lastmansleeping pointed me to an old branch we'd been using in preparation for the Activate talk, in which we really wanted to know for sure that scala/python were behaving the same.  I had written some code then to read the model_predictions.csv file and compare, but then we never cleaned it up for review.

What happens on this branch: if you take the feature_config in jvm/ml4ir-inference/src/test/resources/ranking_happy_path and put it in python/data, and train a simple ranking model with it: 

`docker-compose run ml4ir python3 ml4ir/applications/ranking/pipeline.py --data_dir ml4ir/applications/ranking/tests/data/tfrecord --feature_config ml4ir/applications/ranking/tests/data/configs/feature_config.yaml --run_id test_with_list_loss --data_format tfrecord --execution_mode train_inference_evaluate --loss_key rank_one_listnet`

this will emit a model into `python/models/test_with_list_loss` and predictions into `python/logs/test_with_list_loss`.  These have been copied into (and checked in) in `jvm/ml4ir-inference/src/test/resources/ranking_happy_path` where they are used in `TensorFlowInferenceTest.testSavedModelBundleWithCSVData`.  This test runs all 1500 examples from the training data csv file though scala inference, and checks to see that the scores are within 0.01 of the same.  Most are within 1e-6, but a few are less precise (not sure why yet).


lastmansleeping:
> Most are within 1e-6, but a few are less precise (not sure why yet).
@jakemannix The beauty of this setup is that I can swap the test model with a far more complex (more features) L3 model like the one we have for Account, say, and verify if the parity gets worse. I will do this and update here. At this point I think it could very well be due to the java-c++-python floating point conversions.

jvm/ml4ir-inference/src/main/scala/ml4ir/inference/tensorflow/data/StringMapHelpers.scala
```diff
@@ -0,0 +1,48 @@
...
+                                        predictedScores: Array[Float])
+
+object StringMapCSVLoader {
+
+  def loadDataFromCSV(dataPath: String, featureConfig: ModelFeaturesConfig): Iterable[StringMapQueryAndPredictions] = {
+    val lines = Source.fromFile(dataPath).getLines().toList
+    val (header, dataLines) = (lines.head, lines.tail)
+    val colNames = header.split(",").map(_.replaceAll("\"", ""))
+    val lineMapper: String => Map[String, String] = (line: String) =>
+      colNames.zip(line.split(",")
```
ducouloa, line 23: I know where you are here! 🙂
We had this formatting issue on the previous integration test. It looks to me that using this string helper is finally the more flexible solution, and I'm all in to have it merge. 

Do you agree? Or would you prefer fixing the Python formatting again?

jakemannix: having CSV files with `b'foo bar'` in them are... ugly.  It feels like that should change.  Is that easy, @lastmansleeping ?

ducouloa: I did this in the `classification_model.py` ... We end up with [this](https://github.com/salesforce/ml4ir/pull/102/files#r561292870) ... Not a huge fan

jakemannix: and re: this being an integration test, @ducouloa - this code assumes the data is already checked in to git, and it probably shouldn't be - it needs your code where we first run the python side (from the exact same git state as the jvm side that's being tested!) to train the model and produce outputs, then compare.  We don't want it to check against a static set.

Also: we want an easily parametrized test harness: someone says "I give python code all of: (test.csv, feature_config.yaml, model_config.yaml, other command-line args), it gives me (model_bundle, model_predictions.csv), then we hand that to JVM, which verifies that the JVM can take all of those and compare model_predictions.csv from the two sides for that config + model + data".  This should be runnable for any set of configs, both at PR-merge time, as well as on production models on real data (i.e. done by / after the production model training completes).

jakemannix: yeah, I hear ya - that change looks ugly, I agree.  It's because we're not using cross-platform standard serialization.  I know everyone is complaining about protobuf, but something like a standard avro, protobuf, yaml, json format would be nice, rather than pandas_df.to_csv()...

PR #122, opened 2021-05-04
Add end_to_end ranking test

ducouloa:
Add a full end_to_end test for ranking pipeline. Train a model in
Python, run some predictions and verify that the jvm is returning the
same values

### DIFF
--- a/jvm/ml4ir-inference/scripts/generate_model.sh
+++ b/jvm/ml4ir-inference/scripts/generate_model.sh
@@ -20,9 +20,19 @@ $EXECUTOR ${CLASSIFICATION_MODEL}/pipeline.py \
     --data_dir ${CLASSIFICATION_MODEL}/tests/data/csv/  \
     --feature_config ${CLASSIFICATION_MODEL}/tests/data/configs/feature_config.yaml \
     --model_config ${CLASSIFICATION_MODEL}/tests/data/configs/model_config.yaml \
-    --run_id ${1} \
+    --run_id ${1}_classification \
     --data_format csv \
     --execution_mode train_inference_evaluate \
     --batch_size 32
 
+
+RANKING_MODEL=ml4ir/applications/ranking
+
+$EXECUTOR ${RANKING_MODEL}/pipeline.py \
+    --data_dir ${RANKING_MODEL}/tests/data/csv \
+    --feature_config ${RANKING_MODEL}/tests/data/configs/feature_config.yaml \
+    --run_id ${1}_ranking \
+    --data_format csv \
+    --execution_mode train_inference_evaluate \
+    --batch_size 32
 cd -
--- a/jvm/ml4ir-inference/scripts/generate_model.sh
+++ b/jvm/ml4ir-inference/scripts/generate_model.sh
@@ -34,5 +34,6 @@ $EXECUTOR ${RANKING_MODEL}/pipeline.py \
     --run_id ${1}_ranking \
     --data_format csv \
     --execution_mode train_inference_evaluate \
+    --loss_key rank_one_listnet \
     --batch_size 32
 cd -
--- a/jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
+++ b/jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
@@ -1,11 +1,12 @@
 package ml4ir.inference.tensorflow
 
 import scala.collection.JavaConverters._
-import ml4ir.inference.tensorflow.data.{ModelFeaturesConfig, StringMapExampleBuilder, StringMapSequenceExampleBuilder, TestData}
+import ml4ir.inference.tensorflow.data.{FeatureConfig, ModelFeaturesConfig, ServingConfig, StringMapExampleBuilder, StringMapSequenceExampleBuilder, TestData}
 import org.junit.Test
 import org.junit.Assert._
 import org.tensorflow.example._
 
+import java.util.stream.Collectors
 import scala.io.Source
 
 @Test
@@ -27,18 +28,8 @@ class TensorFlowInferenceIT extends TestData {
   object StringMapCSVLoader {
 
     def loadDataFromCSV(dataPath: String, featureConfig: ModelFeaturesConfig): Iterable[StringMapQueryAndPredictions] = {
-      val servingNameTr = Map("query_id" -> "queryId",
-        "clicked" -> "clicked",
-        "text_match_score" -> "textMatchScore",
-        "page_views_score" ->                        "pageViewsScore",
-        "quality_score" ->                               "qualityScore",
-        "name_match" ->                              "nameMatch",
-        "query_text" ->       "q",
-        "domain_id" -> "domainID",
-        "domain_name" -> "domainName",
-        "rank" -> "rank",
-        "ranking_score" -> "rankingScore",
-        "new_rank" -> "newRank");
+      val servingNameTrList = featureConfig.features.map { case FeatureConfig(train, _, ServingConfig(inference, _), _) => train -> inference}
+      val servingNameTr = servingNameTrList.toMap;
 
       val lines = Source.fromFile(dataPath).getLines().toList
       val (header, dataLines) = (lines.head, lines.tail)
--- a/jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
+++ b/jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
@@ -6,7 +6,6 @@ import org.junit.Test
 import org.junit.Assert._
 import org.tensorflow.example._
 
-import java.util.stream.Collectors
 import scala.io.Source
 
 @Test
@@ -22,18 +21,20 @@ class TensorFlowInferenceIT extends TestData {
    */
   case class PredictionVector(sequence: Map[String, String], scores: Array[Float])
 
+  /**
+   * Class holding a Ranking sequence of documents from Python prediction.
+   */
   case class StringMapQueryAndPredictions(queryContext: Map[String, String],
                                           docs: List[Map[String, String]],
                                           predictedScores: Array[Float])
   object StringMapCSVLoader {
 
     def loadDataFromCSV(dataPath: String, featureConfig: ModelFeaturesConfig): Iterable[StringMapQueryAndPredictions] = {
-      val servingNameTrList = featureConfig.features.map { case FeatureConfig(train, _, ServingConfig(inference, _), _) => train -> inference}
-      val servingNameTr = servingNameTrList.toMap;
+      val servingNameTr = featureConfig.features.map { case FeatureConfig(train, _, ServingConfig(inference, _), _) => train -> inference }.toMap
 
       val lines = Source.fromFile(dataPath).getLines().toList
       val (header, dataLines) = (lines.head, lines.tail)
-      val colNames = header.split(",").map( n => servingNameTr.getOrElse(n, "null"))
+      val colNames = header.split(",").map( n => servingNameTr.getOrElse(n, n))
       val lineMapper: String => Map[String, String] = (line: String) => colNames.zip(line.split(",")).toMap
       val data: List[Map[String, String]] = dataLines.map(lineMapper)
 
@@ -45,12 +46,12 @@ class TensorFlowInferenceIT extends TestData {
       val groupMapper = (group: List[Map[String, String]]) => {
         val context: Map[String, String] = group.head.filterKeys(contextFeatures.contains)
         val docs: List[Map[String, String]] = group.map(_.filterKeys(sequenceFeatures.contains))
-        val predictedScores: Array[Float] = group.map(_.apply("rankingScore").toFloat).toArray
+        val predictedScores: Array[Float] = group.map(_.apply("ranking_score").toFloat).toArray
         (context, docs, predictedScores)
       }
 
       val contextsAndDocs: Iterable[(Map[String, String], List[Map[String, String]], Array[Float])] =
-        data.groupBy(_("queryId")).values.map(groupMapper)
+        data.groupBy(_("query_id")).values.map(groupMapper)
 
       contextsAndDocs.map(pair => StringMapQueryAndPredictions(pair._1, pair._2, pair._3))
     }
@@ -58,7 +59,7 @@ class TensorFlowInferenceIT extends TestData {
   }
 
   /**
-   * Basic parsing of the Python CSV predicition.
+   * Basic parsing of the Python CSV prediction.
    *
    * @param line
    * @return
--- a/jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
+++ b/jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
@@ -27,6 +27,17 @@ class TensorFlowInferenceIT extends TestData {
   case class StringMapQueryAndPredictions(queryContext: Map[String, String],
                                           docs: List[Map[String, String]],
                                           predictedScores: Array[Float])
+
+  /**
+   *
+   * @param dataPath fully qualified filesystem path to the "model_predictions.csv" file as produced by
+   *                 the train_inference_evaluate mode of pipeline.py (see the README at
+   *                 https://github.com/salesforce/ml4ir/tree/master/python )
+   * @param featureConfig the in-memory representation of the "feature_config.yaml" that the python training
+   *                      process used for training
+   * @return an iterable collection over what the training code got for ranking inference results, to compare with
+   *         what the JVM inference sees
+   */
   object StringMapCSVLoader {
 
     def loadDataFromCSV(dataPath: String, featureConfig: ModelFeaturesConfig): Iterable[StringMapQueryAndPredictions] = {
--- a/jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
+++ b/jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
@@ -1,12 +1,11 @@
 package ml4ir.inference.tensorflow
 
 import scala.collection.JavaConverters._
-import ml4ir.inference.tensorflow.data.{ModelFeaturesConfig, StringMapExampleBuilder, TestData}
+import ml4ir.inference.tensorflow.data.{ModelFeaturesConfig, StringMapExampleBuilder, StringMapSequenceExampleBuilder, TestData}
 import org.junit.Test
 import org.junit.Assert._
 import org.tensorflow.example._
 
-import scala.collection.mutable.ListBuffer
 import scala.io.Source
 
 @Test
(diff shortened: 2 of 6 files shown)


jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -27,18 +28,8 @@ class TensorFlowInferenceIT extends TestData {
...
-        "page_views_score" ->                        "pageViewsScore",
-        "quality_score" ->                               "qualityScore",
-        "name_match" ->                              "nameMatch",
-        "query_text" ->       "q",
-        "domain_id" -> "domainID",
-        "domain_name" -> "domainName",
-        "rank" -> "rank",
-        "ranking_score" -> "rankingScore",
-        "new_rank" -> "newRank");
+      val servingNameTrList = featureConfig.features.map { case FeatureConfig(train, _, ServingConfig(inference, _), _) => train -> inference}
```
ducouloa, line 31: @jakemannix I can't get this code to make a simple Map[String, String]

ducouloa:
@FranciscoBorges giving some context about this review: 
The integration test are run on the JVM side of the project as part of a `mvn verify` target. (check the pom.xml there).

There's a pre-integration step that will run a script to train the model in Python, and then the Integration test will execute prediction against the model prediction in the dedicated test directories.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -53,11 +89,109 @@ class TensorFlowInferenceIT extends TestData {
     return Source.fromFile(csvFile).getLines.toList.tail.flatMap(extractColumnValues)
```
jakemannix, line 89: no need for return keyword in scala (last value in an expression / method is the "return value" of it)

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -53,11 +89,109 @@ class TensorFlowInferenceIT extends TestData {
...
+      maskedScore <- maskedScores
+      docScore <- docScores
+    } {
+      assertTrue(
+        s"docScore ($docScore) should be > masked score ($maskedScore)",
+        docScore > maskedScore
+      )
+    }
+    if (query.predictedScores != null) {
+      assertArrayEquals("scores aren't close enough: ", docScores, query.predictedScores, 1e-2f)
```
jakemannix, line 111: it looks like this actually passes with 1e-4 (but not 1e-6) which is better than what I was seeing before.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -53,11 +89,109 @@ class TensorFlowInferenceIT extends TestData {
...
+    }
+
+  }
+
+  def validateScores(scores: Array[Float], numDocs: Int) = {
```
jakemannix, line 116: we don't need two totally almost identical "validateScores" methods, do we?

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -53,11 +89,109 @@ class TensorFlowInferenceIT extends TestData {
...
+      assertTrue(
+        s"docScore ($docScore) should be > masked score ($maskedScore)",
+        docScore > maskedScore
+      )
+    }
+    assertTrue(
+      "second doc should score better than first",
+      scores(1) > scores(0)
+    )
+    println(scores.mkString(", "))
```
jakemannix, line 135: we should probably remove these superfluous println()'s

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -53,11 +89,109 @@ class TensorFlowInferenceIT extends TestData {
...
+  }
+
+  @Test
+  def testRankingSavedModelBundleWithCSVData(): Unit = {
+    val generatedBundleLocation = System.getProperty("bundleLocation")
+    def modelName = System.getProperty("runName") + "_ranking"
```
jakemannix, line 141: we should have a non-@Test annotated method which takes in bundleLocation and runName as method parameters and does what this method does after these lines.  Then the @Test method just grabs the System properties and executes this method with them.

That way, we can easily change it to run not using system properties, but instead commandline args, or having the training script write pairs of paths to a file, or whatever.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -53,11 +89,109 @@ class TensorFlowInferenceIT extends TestData {
...
+  }
 
   @Test
   def testClassificationGeneratedModelBundle(): Unit = {
     val generatedBundleLocation = System.getProperty("bundleLocation")
-    def modelName = System.getProperty("runName")
+    def modelName = System.getProperty("runName") + "_classification"
     val bundlePath = generatedBundleLocation + "models/" + modelName + "/final/tfrecord"
     val predictionPath = generatedBundleLocation + "logs/" + modelName + "/model_predictions.csv"
     //val featureConfigPath = generatedBundleLocation + "ml4ir/applications/classification/tests/data/configs/feature_config.yaml"
```
jakemannix, line 197: have we figured out yet why the classification feature_config.yaml doesn't work?  Is it because of the docker image issue again?

jakemannix (approved):
other than minor suggestions, LGTM!

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
ducouloa: It doesn't seems to be a docker issue for this one. Maybe a parsing issue with the extra-argument.
I'll add a follow-up work to figure out the problem.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
ducouloa: I added this. But I didn't add (yet) a bootstrap class to have this run stand-alone.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
ducouloa: 1e-6 passes locally, let's see what is the CI saying.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
ducouloa: 👍

Ullimague (approved):

jvm/ml4ir-inference/scripts/generate_model.sh
```diff
@@ -20,9 +20,19 @@ $EXECUTOR ${CLASSIFICATION_MODEL}/pipeline.py \
     --data_dir ${CLASSIFICATION_MODEL}/tests/data/csv/  \
     --feature_config ${CLASSIFICATION_MODEL}/tests/data/configs/feature_config.yaml \
     --model_config ${CLASSIFICATION_MODEL}/tests/data/configs/model_config.yaml \
-    --run_id ${1} \
+    --run_id ${1}_classification \
```
lastmansleeping, line 23: @ducouloa Is this meant to be `_ranking`?

jvm/ml4ir-inference/scripts/generate_model.sh
```diff
@@ -20,9 +20,19 @@ $EXECUTOR ${CLASSIFICATION_MODEL}/pipeline.py \
...
 
+
+RANKING_MODEL=ml4ir/applications/ranking
+
+$EXECUTOR ${RANKING_MODEL}/pipeline.py \
+    --data_dir ${RANKING_MODEL}/tests/data/csv \
+    --feature_config ${RANKING_MODEL}/tests/data/configs/feature_config.yaml \
+    --run_id ${1}_ranking \
+    --data_format csv \
+    --execution_mode train_inference_evaluate \
```
lastmansleeping, line 36: I recommend adding `--loss_key rank_one_listnet` here as well. This is the most commonly used loss internally.

python/ml4ir/base/io/local_io.py
```diff
@@ -147,6 +147,8 @@ def write_df(self, df, outfile: str = None, sep: str = ",", index: bool = True)
             dataframe in csv form if outfile is None
         """
         self.log("Writing dataframe to : {}".format(outfile))
+        np.set_printoptions(formatter={'all': lambda x: str(x.decode('utf-8')) if isinstance(x, bytes) else str(x)},
```
lastmansleeping, line 150: Good one. Worth leaving a note why you are doing this. I mean, I know, but good for documentation.
@ducouloa

python/ml4ir/applications/ranking/tests/data/configs/feature_config.yaml
```diff
@@ -45,7 +45,7 @@ features:
     node_name: text_match_score
     trainable: true
     dtype: float
-    log_at_inference: false
+    log_at_inference: true
```
lastmansleeping, line 48: @ducouloa I mentioned this to @jakemannix as well. This is fine (and even good) for now. But we should also think about decoupling the end to end testing from reading the features from the predictions file.

There are two reasons:
- At training time, this data is being read from tfrecords and then is getting converted into numpy/pandas objects. There is a lot of type conversion happening. More importantly, the data here already has all the feature processing applied. This is a good thing, but might not be later on.
- Less important, but still relevant - if we want to use this for training time validation (which we definitely should), we will have to write an entire CSV file full of duplicate feature values (feature values that are already available in the TFRecord). If we are training 50 models (for hyper param optimization), this means that we will be writing back 50x the input data size onto deepsea (or at the least the docker image). We should avoid this if possible. Unfortunately, there is no easy solution to this until we can read TFRecord files directly in java. I don't recommend delving into that now.

@ducouloa Do NOT implement any of the changes recommended here for now. We need to think more about this. As I said, this is more than good for now.

python/ml4ir/base/model/relevance_model.py
```diff
@@ -490,6 +491,13 @@ def predict(
         for predictions_dict in test_dataset.map(_predict_fn).take(-1):
             predictions_df = pd.DataFrame(predictions_dict)
             if logs_dir:
+                np.set_printoptions(
+                    formatter={'all': lambda x: str(x.decode('utf-8')) if isinstance(x, bytes) else str(x)},
+                    linewidth=sys.maxsize, threshold=sys.maxsize)  # write the full line in the csv not the truncated version.
+                for col in predictions_df.columns:
+                    if isinstance(predictions_df[col].values[0], bytes):
+                        predictions_df[col] = predictions_df[col].str.decode('utf8')
+
```
lastmansleeping, line 500: 👍🏼

jvm/ml4ir-inference/scripts/generate_model.sh
```diff
@@ -20,9 +20,19 @@ $EXECUTOR ${CLASSIFICATION_MODEL}/pipeline.py \
...
     --feature_config ${CLASSIFICATION_MODEL}/tests/data/configs/feature_config.yaml \
     --model_config ${CLASSIFICATION_MODEL}/tests/data/configs/model_config.yaml \
-    --run_id ${1} \
+    --run_id ${1}_classification \
     --data_format csv \
     --execution_mode train_inference_evaluate \
     --batch_size 32
 
+
+RANKING_MODEL=ml4ir/applications/ranking
```
lastmansleeping, line 29: @ducouloa Can you add a section in the readme to explain how this script is being used? I assumed the Java test would call this bash script, but doesn't look like that is happening. Is it the other way around?

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -50,18 +86,105 @@ class TensorFlowInferenceIT extends TestData {
...
+      )
+    }
+    if (query.predictedScores != null) {
+      assertArrayEquals("scores aren't close enough: ", docScores, query.predictedScores, 1e-6f)
+    }
+
+  }
+
+  @Test
+  def testRankingSavedModelBundleWithCSVData(): Unit = {
```
lastmansleeping, line 117: @ducouloa Is there a way to run this function by passing say command line arguments to the model and predictions CSV? I ask because I am interested in using this when we train every single model - like a seal of approval and the corresponding jar version associated with it.
CC: @Ullimague

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -50,18 +86,105 @@ class TensorFlowInferenceIT extends TestData {
...
+    evaluateRankingInferenceAccuracy(bundlePath, predictionPath, featureConfigPath)
+  }
+
+  def evaluateRankingInferenceAccuracy(bundlePath: String, predictionPath: String, featureConfigPath: String) = {
+    val allScores: Iterable[(StringMapQueryAndPredictions, SequenceExample, Array[Float], Array[Float])] = runQueriesAgainstDocs(
+      predictionPath,
+      bundlePath,
+      featureConfigPath,
+      "serving_tfrecord_protos",
+      "StatefulPartitionedCall_1"
```
lastmansleeping, line 133: We should probably not hardcode this. @ducouloa @jakemannix ?

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -23,7 +22,44 @@ class TensorFlowInferenceIT extends TestData {
...
   /**
-   * Basic parsing of the Python CSV predicition.
+   * Class holding a Ranking sequence of documents from Python prediction.
+   */
+  case class StringMapQueryAndPredictions(queryContext: Map[String, String],
+                                          docs: List[Map[String, String]],
+                                          predictedScores: Array[Float])
+  object StringMapCSVLoader {
+
+    def loadDataFromCSV(dataPath: String, featureConfig: ModelFeaturesConfig): Iterable[StringMapQueryAndPredictions] = {
```
lastmansleeping, line 32: @ducouloa Can you add docstrings for the arguments here?

lastmansleeping (commented):
@ducouloa I've left a few comments. But should be good as approved. Take a look when free.

Thanks for this work :)

python/ml4ir/applications/ranking/tests/data/configs/feature_config.yaml
jakemannix: @lastmansleeping can we open a work item to track the issues around log_at_inference usage in tests/prod configs?

jvm/ml4ir-inference/scripts/generate_model.sh
jakemannix: @lastmansleeping it's used by the `mvn verify -Pintegration_tests` test step - does the README not explain this well enough?  https://github.com/salesforce/ml4ir#continuous-integration

python/ml4ir/applications/ranking/tests/data/configs/feature_config.yaml
lastmansleeping: Yep, let me do that.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
jakemannix: that's my bad - I didn't give @ducouloa my scaladocs for this:

`    /**
     *
     * @param dataPath fully qualified filesystem path to the "model_predictions.csv" file as produced by
     *                 the train_inference_evaluate mode of pipeline.py (see the README at
     *                 https://github.com/salesforce/ml4ir/tree/master/python )
     * @param featureConfig the in-memory representation of the "feature_config.yaml" that the python training 
     *                      process used for training
     * @return an iterable collection over what the training code got for ranking inference results, to compare with
     *         what the JVM inference sees
     */`

jvm/ml4ir-inference/scripts/generate_model.sh
lastmansleeping: No, that works. I think a line or two about what the end to end test is doing specifically would be useful. My intention is to say some DS with minimal JVM knowledge trained a model with ml4ir and wanted to verify if they could deploy it, how would they verify it. We have the code right now, would be nice to have the steps to do that too. If this requires additional work to generalize, we can create a follow up WI.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
jakemannix: One fairly simple thing we could really use, @ducouloa , is the ability to do e.g.:

```
mvn test -Dtest=TensorFlowInferenceIT#testRankingSavedModelBundleWithCSVData -DbundleLocation=/path/to/my/trained/model -DrunName=myRunName
```

This would remove the need for a "standalone bootstrap", right?

jakemannix: actually @ducouloa , @lastmansleeping - what I wrote above literally just works.  You just need to have the paths pointing to the right place (i.e. where a) the model bundle is, b) where the python logs are, and c) where the feature_config.yaml is.  If all those are right, the above `mvn test` commandline will work on any ranking model, as long as you ran training in train_inference_evaluate with a feature_config with log_at_inference=true

lastmansleeping: That's great if we can call it from command line. I can hook it up at the end of the Flowsnake model training.

jvm/ml4ir-inference/scripts/generate_model.sh
ducouloa: Ok, I'll add this.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
ducouloa: This is currently hardcoded everywhere (even in the Core-app  code - I think W-8375415 is to tackle the work) - I can create an equivalent for ml4ir project

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
ducouloa: Thanks a lot for this Jake.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
FranciscoBorges: @ducouloa the bootstrap still getting added? If so, please add a quick paragraph about it at whichever would be the appropriated place at the docs.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
FranciscoBorges: @ducouloa if we are not adding the bootstrap, I think we should document the CLI command that @jakemannix shared.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
ducouloa: @FranciscoBorges There is W-9280938 to follow-up.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
ducouloa: i added a note in the main README.md for this.

Ullimague (approved):

lastmansleeping (approved):
LGTM

PR #123, opened 2021-05-10
Adding L1 sanity tests

mohazahran:
- Adding L1 sanity tests
- Exposing the bias value (if there is one) at the end of the training.


python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,258 @@
...
+    - seed: Random seed
+    Returns: the best seprabale data set by a Perceptorn.
+    """
+    separable = False
+    seed = args.seed
+    rng = check_random_state(args.seed)
+    best_accuracy = 0
+    best_accuracy_data = None
+    trials = 1000
+    while not separable and trials > 0:
```
balikasg, line 81: This is a strange loop in the sense that `seperable` is handcoded to `False`.. We could simplify I guess..

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,258 @@
...
+    """
+    X = [df['f' + str(i)] for i in range(featureCount)]
+    Y = df['clicked']
+    model.fit(np.array(X).T, Y)
+    for i in range(featureCount):
+        model.coef_[0][i] = 1.0
+    if model.fit_intercept:
+        model.intercept_ = 1.0
+    else:
+        model.intercept_ = 0
```
balikasg, line 152: You could also try something in the lines of: 
```
model.intercept = float(model.fit_intercept)
```

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -346,6 +346,10 @@ def save(
                 tf.squeeze(dense_layer.get_weights()[0]).numpy())
             ),
             columns=["feature", "weight"])
+        # Adding log for bias value
+        if len(dense_layer.get_weights())>1:
+            bias_val = dense_layer.get_weights()[1][0]
+            linear_model_coefficients.loc[len(linear_model_coefficients.index)] = ['bias', bias_val]
```
balikasg, line 352: It is strange that we are hard-coding a name like `bias` in this low-level code. What is the effect of these lines of code before/after their addition to the dataframe `linear_model_coefficients`? I have not used this part of the library yet so this may be a silly comment.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,258 @@
...
+        samples = datasets.make_classification(n_classes=args.n_classes, n_samples=args.n_samples, n_features=args.n_features,
+                                               n_redundant=args.n_redundant, n_informative=args.n_informative,
+                                               n_clusters_per_class=args.n_clusters_per_class, flip_y=args.flip_y,
+                                               class_sep=args.class_sep, random_state=rng)
+        p = Perceptron(fit_intercept=True, max_iter=1000, verbose=0, random_state=seed, validation_fraction=0.0001)
+        p.fit(samples[0], samples[1])
+        acc = p.score(samples[0], samples[1])
+        if acc > best_accuracy:
+            best_accuracy = acc
+            best_accuracy_data = samples
```
balikasg, line 91: it is 1K values, we could easily have a list and get max that would probably increase readability

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,258 @@
+import unittest
+import warnings
+import pandas as pd
+import numpy as np
+from sklearn import datasets
+from sklearn.linear_model import Perceptron, LogisticRegression
+from sklearn.utils import check_random_state
```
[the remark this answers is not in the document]
mohazahran: One main point in doing these sanity tests is to be independent of ml4ir in the sense that I cannot use ml4ir to test ml4ir.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,258 @@
...
+            best_accuracy = acc
+            best_accuracy_data = samples
+        acc = p.score(samples[0], samples[1])
+        if acc == 1.0:
+            break
+        trials -= 1
+    return best_accuracy_data[0], best_accuracy_data[1]
+
+
+def calculate_MRR(prediction, y):
```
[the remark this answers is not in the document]
mohazahran: yes, but again, trying to be totally independent is one aspect of sanity checks.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
mohazahran: You're right. I made a change and the break statement needs to be replaced by "separable = True"

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
mohazahran: I'm trying here to set an initial value for the intercept to be 1.0 if the model is configured to use an intercept.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,258 @@
...
+    - class_sep: Larger values spread out the clusters/classes and make the classification task easier.
+    - seed: Random seed
+    Returns: the best seprabale data set by a Perceptorn.
+    """
+    separable = False
+    seed = args.seed
+    rng = check_random_state(args.seed)
+    best_accuracy = 0
+    best_accuracy_data = None
+    trials = 1000
```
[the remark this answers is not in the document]
mohazahran: Sanity tests trying to assess ml4ir ability to learn L1 model that should be as least as good two simple models ( as a Perceptron and a logistic regression) in MRR with these models trained for accuracy and ml4ir is selecting models based on MRR.
As for running time, it's very cheap to train a Perceptron. All the 4 tests takes about 4 mins to run, if we see that's need to be decreased we can reduce the number of trails.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,258 @@
...
+    while not separable and trials > 0:
+        samples = datasets.make_classification(n_classes=args.n_classes, n_samples=args.n_samples, n_features=args.n_features,
+                                               n_redundant=args.n_redundant, n_informative=args.n_informative,
+                                               n_clusters_per_class=args.n_clusters_per_class, flip_y=args.flip_y,
+                                               class_sep=args.class_sep, random_state=rng)
+        p = Perceptron(fit_intercept=True, max_iter=1000, verbose=0, random_state=seed, validation_fraction=0.0001)
+        p.fit(samples[0], samples[1])
+        acc = p.score(samples[0], samples[1])
```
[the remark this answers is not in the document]
mohazahran: I checked and it's only classification accuracy.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
balikasg: I meant you could replace the whole if/else with a single statement 
```
model.intercept_ = float(model.fit_intercept)
```
Not too important though

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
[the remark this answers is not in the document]
balikasg: I assume you did (and I also know it because I like sklearn a lot). What I am trying to say is that for someone who does not know this price of information using 
```
sklearn.metrics.accurcy_score(.., ..)
``` 
is self-explanatory.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
balikasg: Thanks. It makes more sense now. For 2-dim we can do it easily, for more I can not think of a way.  
I'd like to understand : 
- why are you trying to generate linearly separable data?
- why is this comparison (scores on the training data actually) meaningful?

mohazahran: The idea of having a linearly separable data is to ensure that a perfect solution exists and ml4ir must reach it. However, test cases are not all linearly separable and the definition of passing a test case is when ml4ir's MRR is at least as good as Perceptron's and logistic regression. (I added you to a quip doc that perform analysis to ml4ir's sanity tests).

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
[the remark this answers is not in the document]
jakemannix: I don't think rewriting MRR in a unit test is the right way to check if the calculation is right.  Is this not just assuming that the code you wrote here is "the gold standard" of computing MRR, and asking "is ml4ir core computing what my code is doing"?

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,258 @@
...
+        model.intercept_ = 0
+    model.fit(np.array(X).T, Y)
+    return model
+
+
+def generate(args):
+    """
+    create an ml4ir consumable dataframe after generating dataset.
+    """
+    X, Y = generate_data(args)
```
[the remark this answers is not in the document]
jakemannix: yes, agreed, let's not just pass around opaque dictionaries of arguments.

python/requirements.txt
```diff
@@ -8,6 +8,7 @@ pytest==6.2.1
 pytest-cov==2.11.0
 pytest-html==3.1.1
 PyYAML==5.4.1
+scikit-learn==0.21.3
```
jakemannix, line 11: as mentioned, let's not add a huge new dependency just for a test.

jakemannix (changes requested):
Two bigger requests: 

1) for sanity checking, I am generally not a fan of just seeing if we compute the same thing as another package, as it just requires that we're "bug-compatible" with them.  I guess it's ok, but I'd rather say do this, for sanity checking a regression: 

* set a training schema of having N numeric features, and have the target variable be generated from the features as logit(f_1(x_1) + f_2(x_2) + ... + f_n(x_n) + f_interactions(x_1, ..., x_n)) + gaussian noise
* (where f_1(x) = x, f_2(x) = 1/((x^2)+1), f_3(x) = exp(-x), f_interactions(x_1, x_2...) = product((x_i)^2) or something like that.
* because you have an explicit form for how the dependent variable is generated, you can check: how close is ml4ir getting, and is it on average, within "noise" of the real value?  If f_n(x) are always linear, a logistic regression model should be very close.  For highly nonlinear f_n, it won't, so maybe "sanity" is just checking case when each of the f_i are linear.
* but if we do go with a simpler approach of just comparing to sklearn, we have to not depend on it in our package.  That's too heavyweight.

2) please use type-hints everywhere.

jakemannix:
looks like unit tests aren't passing CI, currently, as well

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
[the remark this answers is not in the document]
mohazahran: I can do that but that would make the tests essentially frozen to a specific number of tests with specific datasets. The way I had in mind to make it configurable so that we can add easily more tests and control the difficulties.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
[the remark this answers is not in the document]
mohazahran: That's true, but again doing so leaves no stones unturned and it was reassuring to find MRR calculated by both have same numbers.

python/build-requirements.txt
```diff
@@ -17,6 +17,7 @@ pandas==1.2.1
 pytest==6.2.1
 pytest-cov==2.11.0
 pytest-html==3.1.1
+scikit-learn==0.21.3
```
lastmansleeping, line 20: If we are using a third party library (which we probably shouldn't for this case), let's make sure it is a pre-approved version.

python/ml4ir/applications/ranking/model/ranking_model.py
lastmansleeping: Makes sense to me. We do not use bias in our training because of how the model is deployed, but this looks like worth having.

python/ml4ir/applications/ranking/tests/data/configs/model_config_sanity_tests.yaml
```diff
@@ -0,0 +1,13 @@
+architecture_key: linear
+layers:
+  - type: dense
+    name: linear_layer
+    units: 1
+    kernel_initializer: Ones
```
lastmansleeping, line 6: Commented on the document as well. But I don't think just doing this makes the two models the same.
- the optimizer state could be different.
- the order of data mini batches will for sure be different because of the `tf.data` optimization leading to possibly different convergence.
@mohazahran

python/requirements.txt
lastmansleeping: Yes, we can just add it to the `build-requirements.txt` I think.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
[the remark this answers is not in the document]
lastmansleeping: I kinda agree with @mohazahran here. If we don't have `sklearn`, and want to compare with another model trained on `ml4ir` we might not be seeing the full suite of issues. But I think what @jakemannix proposes is a good middle ground for this.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,258 @@
...
+import numpy as np
+from sklearn import datasets
+from sklearn.linear_model import Perceptron, LogisticRegression
+from sklearn.utils import check_random_state
+from ml4ir.applications.ranking.pipeline import RankingPipeline
+from ml4ir.applications.ranking.config.parse_args import get_args
+import shutil
+import pathlib
+import random
+from argparse import Namespace
```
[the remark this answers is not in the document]
lastmansleeping: +1
I typically follow
```
core packages

ml4ir

constants
```

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
+import unittest
```
lastmansleeping, line 1: @mohazahran Please add [numpy style docstrings](https://numpydoc.readthedocs.io/en/latest/format.html) for all the methods and. as @jakemannix mentioned, [type hints](https://docs.python.org/3/library/typing.html).

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+    rows = []
+    query_count = 0
+    document_count = 0
+    while consumed_clicked_docs < len(clicked_docs) and consumed_nonclicked_docs < len(nonclicked_docs):
+        qid = 'query'+str(query_count)
+        query_count += 1
+        r = {'query_id':qid}
+        docs_count = random.randint(2, 25)
+        r['record_id'] = 'record'+str(document_count)
+        r['fr'] =docs_count
```
lastmansleeping, line 38: We really shouldn't be hardcoding `query_id`, `fr` (especially this). Can we not use the `feature_config.query_key` and `feature_config.rank` here? @mohazahran

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+import random
+from argparse import Namespace
+warnings.filterwarnings("ignore")
+
+
+def create_ml4ir_dataframe(X, Y):
+    """
+    Takes the generated data (X: data features, Y: clicked) and formulate a dataframe consumable by ml4ir
+    """
+    def add_features_to_record(X, xid, r, label):
```
lastmansleeping, line 20: @mohazahran What are the arguments here?

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+        for i in range(len(X[xid])):
+            r['f' + str(i)] = X[xid][i]
+        r['clicked'] = label
+
+    clicked_docs = X[Y == 1]
+    nonclicked_docs = X[Y == 0]
+    consumed_clicked_docs = 0
```
lastmansleeping, line 27: @mohazahran What are `consumed_clicked_docs`?

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+    consumed_nonclicked_docs = 0
+    rows = []
+    query_count = 0
+    document_count = 0
+    while consumed_clicked_docs < len(clicked_docs) and consumed_nonclicked_docs < len(nonclicked_docs):
+        qid = 'query'+str(query_count)
+        query_count += 1
+        r = {'query_id':qid}
+        docs_count = random.randint(2, 25)
+        r['record_id'] = 'record'+str(document_count)
```
lastmansleeping, line 37: @mohazahran You probably don't need `record_id` if you have `rank`, no?

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+                add_features_to_record(nonclicked_docs, consumed_nonclicked_docs, r, 0)
+                consumed_nonclicked_docs += 1
+                document_count += 1
+                rows.append(r)
+            else:
+                break
+    return pd.DataFrame(rows)
+
+
+def generate_data(args):
```
lastmansleeping, line 57: I get the temptation to do this and I've done it in the past as well, but I think listing all arguments individually here is a more robust coding patten. @mohazahran

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+    - n_redundant: The number of redundant features
+    - n_informative: The number of informative (Non-redundant) features
+    - n_clusters_per_class: The number of clusters per class.
+    - flip_y: The fraction of samples whose class is assigned randomly. Larger
+        values introduce noise in the labels and make the classification
+        task harder.
+    - class_sep: Larger values spread out the clusters/classes and make the classification task easier.
+    - seed: Random seed
+    - n_trials: Number of trials of data generation to obtain a linearly separable data
+    Returns: the best seprabale data set by a Perceptorn.
```
lastmansleeping, line 72:
```suggestion
    Returns: the best seprabale data set by a perceptron.
```

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+def generate_data(args):
+    """
+    Generate data with the specified parameters
+    - n_classes: Number of classes in the datasets. Should be 2 (clicked, not clicked)
+    - n_samples: Number of samples to generate
+    - n_features: Number of features per sample
+    - n_redundant: The number of redundant features
+    - n_informative: The number of informative (Non-redundant) features
```
lastmansleeping, line 64: @mohazahran This is very interesting.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
lastmansleeping: You should also definitely add default values for all these arguments, maybe the same as the ones in sklearn if that's the best for the test.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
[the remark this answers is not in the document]
lastmansleeping: I did not remember this because incidentally I have not used sklearn in a while.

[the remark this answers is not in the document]
lastmansleeping: I think what also is a useful thing to incorporate (I include myself as well) is to write code like
```
...score(y_true=samples[0], y_pred=samples[1])
```

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
lastmansleeping: @mohazahran This is worth documenting in the code as well. There is a lot of logic happening here, good stuff. But `generate_data` is such a catch all function. We need to update it's description with the high level steps you are performing.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
lastmansleeping: Okay. I now understand after spending some time here. Worth adding a comment for the loop header. Something like `New dataset is generated until it is linearly separable or max num trials are reached`

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+    X = [df['f' + str(i)] for i in range(len(W)-1)]
+    Y = df['clicked']
+    wrapper_model.fit(np.array(X).T, Y)
+    for i in range(len(W)-1):
+        wrapper_model.coef_[0][i] = W[i]
+    wrapper_model.intercept_ = W[-1]
+    return wrapper_model
+
+
+def test_ml4ir_sanity_pipeline(args, p, log_dir):
```
lastmansleeping, line 179: @mohazahran Not sure if this will be picked up by `pytest`.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+    """
+    separable = False
+    best_accuracy = 0
+    best_accuracy_data = None
+    random.seed(args.seed)
+    seeds = list(range(args.n_trials))
+    random.shuffle(seeds)
+    trials = len(seeds)-1
+    while not separable and trials > 0:
+        samples = datasets.make_classification(n_classes=args.n_classes, n_samples=args.n_samples, n_features=args.n_features,
```
lastmansleeping, line 82: @mohazahran Do we need to generate the data multiple times to get linearly separable? Does setting `class_sep` to a really high value not work out of the box? I'm just trying to understand the reasoning behind the 100 trials here.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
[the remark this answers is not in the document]
lastmansleeping: @mohazahran I think we should have separate unit tests for computing metrics, for sure. Ideally, we would only want to test the optimization/weights convergence in these tests. But I am not too opposed to what is done here for now (until we add extensive unit tests for MRR, ACR itself).

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+    - n_clusters_per_class: The number of clusters per class.
+    - flip_y: The fraction of samples whose class is assigned randomly. Larger
+        values introduce noise in the labels and make the classification
+        task harder.
+    - class_sep: Larger values spread out the clusters/classes and make the classification task easier.
+    - seed: Random seed
+    - n_trials: Number of trials of data generation to obtain a linearly separable data
+    Returns: the best seprabale data set by a Perceptorn.
+    """
+    separable = False
```
lastmansleeping, line 74: Should this be an argument or is this indirectly controlled already by `class_sep`? @mohazahran

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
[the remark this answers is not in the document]
lastmansleeping: FYI `relevance_pipeline.run()` already returns all the metrics you'd need I think. Check it out -> https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/base/pipeline.py#L273_L277

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+        clickedDocScore = pred[clickedDocIdx]
+        # ties are resolved with the worst case (i.e. the correct doc will have the lowest place among other docs with same score)
+        equalScoreCount = sortedPred.count(clickedDocScore)
+        predictedRank = sortedPred.index(clickedDocScore) + equalScoreCount - 1.0
+        mrr += 1.0 / (predictedRank + 1.0)
+    mrr = mrr / len(clickedDocIdxs)
+    return mrr
+
+
+def check_MRR(df, model, featureCount=2):
```
lastmansleeping, line 118: @mohazahran [Minor] Why are you switching to camel case for variable names here? Can we keep it consistent?

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+    queries = df.groupby('query_id')
+    for qid, group in queries:
+        X = [group['f' + str(i)] for i in range(featureCount)]
+        pred = model.decision_function(np.array(X).T)
+        mrr = calculate_MRR(pred, group['clicked'])
+        MRR += mrr
+    return MRR/len(queries)
+
+
+def check_accuracy(df, model, featureCount=2):
```
lastmansleeping, line 132: Can rename to `compute_model_accuracy` @mohazahran

lastmansleeping: You should add docstrings here so that people know what `model` really is.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
lastmansleeping: Why are you trying to set the intercept here?

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+    acc = model.score(np.array(X).T, Y)
+    return acc
+
+
+def train_sklearn_model(df, model, featureCount=2):
+    """
+    Trains a sklearn model (Specifically, logistic regression and Perceptron)
+    """
+    X = [df['f' + str(i)] for i in range(featureCount)]
+    Y = df['clicked']
```
lastmansleeping, line 144: @mohazahran Please use `feature_config.label` here.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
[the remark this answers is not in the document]
lastmansleeping: Haha, I mentioned the same somewhere above as well with personal notes like @balikasg .

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+        train_ml4ir(p.as_posix(), feature_config_file.as_posix(), model_config_file.as_posix(), log_dir.as_posix())
+        logisticReg = LogisticRegression(fit_intercept=False, random_state=args.seed, max_iter=100,
+                                warm_start=True)
+        perceptron = Perceptron(fit_intercept=False, max_iter=100, random_state=args.seed, validation_fraction=0.0001,
+                       warm_start=True)
+        ml4ir_weights = pd.read_csv("models/test_command_line/coefficients.csv")["weight"].tolist()
+        ml4ir_weights.append(0.0)
+        ml4ir = wrap_linear_model_in_sklearn_model(df, ml4ir_weights)
+        train_sklearn_model(df, logisticReg, featureCount=args.n_features)
+        train_sklearn_model(df, perceptron, featureCount=args.n_features)
```
lastmansleeping, line 197: @mohazahran We should definitely get rid of this logic here IMO. If I understand right, you are trying to load the trained model weights from ml4ir and then assign them + freeze a sklearn Perceptron. This is then used to compute MRR and accuracy. It is a lot of overkill I think. You can very well get all these metrics from `relevance_model.run_pipeline()` already.

CC: @jakemannix

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+def train_ml4ir(data_dir, feature_config, model_config, logs_dir):
+
+    argv = ['--data_dir', data_dir, '--feature_config', feature_config, '--loss_type', "pointwise", '--scoring_type', "pointwise",
+            '--run_id', 'test_command_line', '--data_format', 'csv', '--execution_mode', 'train_inference_evaluate', '--loss_key', 'sigmoid_cross_entropy',
+            '--num_epochs', '150', '--model_config', model_config, '--batch_size', '16', '--logs_dir', logs_dir,
+            '--max_sequence_size', '25', '--train_pcent_split', '1.0', '--val_pcent_split', '-1', '--test_pcent_split', '-1', '--early_stopping_patience', '1000']
+    args = get_args(argv)
+    rp = RankingPipeline(args=args)
```
lastmansleeping, line 212: @mohazahran Conversely, if you do not want to do this string list thing, you could directly call `RankingModel` as well, but you'd probably need some additional code. I don't feel too strongly about this as this is also one of the intended use cases of ml4ir.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+    rp.run()
+
+
+class TestML4IRSanity(unittest.TestCase):
+    def setUp(self):
+        self.dir = pathlib.Path(__file__).parent
+        self.p = pathlib.Path(self.dir) / "data" /"ml4ir_sanity_test_working_dir"
```
lastmansleeping, line 219: What is `p` here? I get it, but I am trying to show you how a new reader might not understand. Always useful to have descriptive variable names.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def tearDown(self):
+        shutil.rmtree(self.p)
+
+    def test_ml4ir_sanity_1(self):
+        args = Namespace(n_classes=2, n_samples=1000, n_features=2, n_redundant=0, n_informative=2,
+                         n_clusters_per_class=2, flip_y=-1, class_sep=1.0, seed=13, n_trials=100)
+        ml4ir_mrr, perceptron_mrr, log_regression_mrr= test_ml4ir_sanity_pipeline(args, self.p, self.log_dir)
+        assert ml4ir_mrr >= perceptron_mrr
+        assert ml4ir_mrr >= log_regression_mrr
```
lastmansleeping, line 238: I don't get why we expect it to be better actually. Shouldn't we check if it is `np.isclose(atol=0.1)`?

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+    argv = ['--data_dir', data_dir, '--feature_config', feature_config, '--loss_type', "pointwise", '--scoring_type', "pointwise",
+            '--run_id', 'test_command_line', '--data_format', 'csv', '--execution_mode', 'train_inference_evaluate', '--loss_key', 'sigmoid_cross_entropy',
+            '--num_epochs', '150', '--model_config', model_config, '--batch_size', '16', '--logs_dir', logs_dir,
+            '--max_sequence_size', '25', '--train_pcent_split', '1.0', '--val_pcent_split', '-1', '--test_pcent_split', '-1', '--early_stopping_patience', '1000']
+    args = get_args(argv)
+    rp = RankingPipeline(args=args)
+    rp.run()
+
+
+class TestML4IRSanity(unittest.TestCase):
```
lastmansleeping, line 216: @mohazahran Just calling out that this only tests pointwise loss and not the listwise loss. Let's discuss about this more on Monday. Would love to hear your thoughts on why the listwise loss is not performing as well on this dataset.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,266 @@
...
+        model.coef_[0][i] = 1.0
+    if model.fit_intercept:
+        model.intercept_ = 1.0
+    else:
+        model.intercept_ = 0
+    model.fit(np.array(X).T, Y)
+    return model
+
+
+def generate(args):
```
lastmansleeping, line 156: You should rename this in a more descriptive way. @mohazahran

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
lastmansleeping: Same for all tests.

lastmansleeping (commented):
Looking good so far @mohazahran. Definitely very interesting work here. I've left a bunch of comments on improving the code. Let me know what you think. And let us discuss on why listwise loss is not performing as expected when you are back.

I also think it is worth not merging this branch into `master` until we identify the reason for the "failures" and fix them. That was the goal of this work technically (not in this work-item). But make sure you identify potential improvements if you see them. I personally think the pointwise loss model is a success. But would love to hear your thoughts.

CC: @jakemannix

balikasg (commented):
Great improvements! 
Some comments: 
- Notice the code replication between `test_linear_ml4ir_sanity_1`, ..., `test_linear_ml4ir_sanity_N`.
- How are these datasets generated? Do we need 8? 
- How are these model coefficients generated? How can contributor Z modify them in the future if something changes? Are we hiding the `sklearn` dependency under the mat now ;-)  
- Do we really need Perceptron?

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
FranciscoBorges: `p` is indeed a very non descriptive field name. 
> Even for a test?
Even for a test.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,178 @@
...
+        self.log_dir.mkdir(parents=True, exist_ok=True)
+        train = self.p / 'train'
+        train.mkdir(parents=True, exist_ok=True)
+        test = self.p / 'test'
+        test.mkdir(parents=True, exist_ok=True)
+        validation = self.p / 'validation'
+        validation.mkdir(parents=True, exist_ok=True)
+
+    def tearDown(self):
+        shutil.rmtree(self.p)
```
FranciscoBorges, line 102: Honestly, I find this kind of call very scary. Test code should not have calls like this.

Please see

https://testfixtures.readthedocs.io/en/latest/files.html#the-decorator

https://testfixtures.readthedocs.io/en/latest/files.html#manual-usage

the last one mentions how to use the library in the context of a `setup` method.

lastmansleeping:
@mohazahran Make sure you re-request review for approval once you are done addressing all comments. I know there are a lot of comments, so feel free to ping me offline to discuss trade offs.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,175 @@
...
+    """
+    Train a pointwise ranker, listwise loss model using ml4ir
+    """
+    argv = ['--data_dir', data_dir, '--feature_config', feature_config, '--loss_type', "listwise", '--scoring_type',
+            "listwise",
+            '--run_id', 'test_command_line', '--data_format', 'csv', '--execution_mode', 'train_inference_evaluate',
+            '--loss_key', 'rank_one_listnet',
+            '--num_epochs', '150', '--model_config', model_config, '--batch_size', '1', '--logs_dir', logs_dir,
+            '--max_sequence_size', '25', '--train_pcent_split', '1.0', '--val_pcent_split', '-1', '--test_pcent_split',
+            '-1', '--early_stopping_patience', '1000']
```
lastmansleeping, line 87: @mohazahran You should indent these things. Otherwise it is a nightmare to read and to play with.
```suggestion
    argv = ["--data_dir", data_dir,
            "--feature_config", feature_config,
            "--loss_type", "listwise",
            "--scoring_type", "listwise",
            "--run_id", "test_command_line",
            "--data_format", "csv",
            "--execution_mode", "train_inference_evaluate",
            "--loss_key", "rank_one_listnet",
            "--num_epochs", "150",
            "--model_config", model_config,
            "--batch_size", "1",
            "--logs_dir", logs_dir,
            "--max_sequence_size", "25",
            "--train_pcent_split", "1.0",
            "--val_pcent_split", "-1",
            "--test_pcent_split", "-1",
            "--early_stopping_patience", "25",
            "--metrics_keys", "MRR", "categorical_accuracy",
            "--monitor_metric", "categorical_accuracy"]
```

You will notice that I have used one trick here - I have set the `monitor_metric` to `categorical_accuracy`. This takes the MRR and accuracy closer to logistic regression for listwise loss. 

And again, I think your success criteria is too strict, especially given the fact that there are only about 40 queries. I would use `np.isclose(..., atol=0.01)`

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,175 @@
...
+def run_sanity_test(n_features, fname, perceptron, log_reg, working_dir, log_dir):
+    """
+    Runs sanity test for linear models.
+    """
+    df = pd.read_csv(pathlib.Path(__file__).parent / "data" / "L1_sanity_tests" / fname)
+    ml4ir_mrr, perceptron_mrr, log_regression_mrr = ml4ir_sanity_pipeline(df, perceptron, log_reg,
+                                                                          working_dir, log_dir,
+                                                                          n_features)
+    assert ml4ir_mrr >= perceptron_mrr
+    assert ml4ir_mrr >= log_regression_mrr
```
lastmansleeping, line 102:
```suggestion
    assert np.isclose(ml4ir_mrr, perceptron_mrr, atol=0.01) or ml4ir_mrr >= perceptron_mrr
    assert np.isclose(ml4ir_mrr, log_regression_mrr, atol=0.01) or ml4ir_mrr >= log_regression_mrr
```

I think this is a fair modification.

lastmansleeping:
@mohazahran I think the reason the linear model save test is failing is because you are now also saving the bias coefficient which wasn't being saved before. Worth checking there.

mohazahran:
> Two bigger requests:
> 
> 1. for sanity checking, I am generally not a fan of just seeing if we compute the same thing as another package, as it just requires that we're "bug-compatible" with them.  I guess it's ok, but I'd rather say do this, for sanity checking a regression:
> 
> * set a training schema of having N numeric features, and have the target variable be generated from the features as logit(f_1(x_1) + f_2(x_2) + ... + f_n(x_n) + f_interactions(x_1, ..., x_n)) + gaussian noise
> * (where f_1(x) = x, f_2(x) = 1/((x^2)+1), f_3(x) = exp(-x), f_interactions(x_1, x_2...) = product((x_i)^2) or something like that.
> * because you have an explicit form for how the dependent variable is generated, you can check: how close is ml4ir getting, and is it on average, within "noise" of the real value?  If f_n(x) are always linear, a logistic regression model should be very close.  For highly nonlinear f_n, it won't, so maybe "sanity" is just checking case when each of the f_i are linear.
> * but if we do go with a simpler approach of just comparing to sklearn, we have to not depend on it in our package.  That's too heavyweight.
> 
> 1. please use type-hints everywhere.

lastmansleeping (commented):
@mohazahran Thanks for incorporating all the comments and feedback.

I still see a few important pending items
- **add docstrings** - missing argument descriptions (and type hints) (numpy docstring style guide)[https://numpydoc.readthedocs.io/en/latest/format.html]
- **remove custom MRR computation** and use value obtained from relevance_model.run(). This is the metric we use for comparing models and declaring victory and should be used here as well. You can remove the weights for log_regression and perceptron and just store their MRR instead. This would speed up the tests a little as well. I would prefer a separate JSON file or a high level dictionary constant.
- **add description for the datasets** under each test. You can even add the make_classification function with arguments in the comments. This will help us replicate results if needed.

Let me know once you finish adding them.

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+    df.to_csv(working_dir / 'train' / 'data.csv')
+    df.to_csv(working_dir / 'validation' / 'data.csv')
+    df.to_csv(working_dir / 'test' / 'data.csv')
+
+    fconfig_name = "feature_config_sanity_tests_" + str(n_features) + "_features.yaml"
+    feature_config_file = pathlib.Path(__file__).parent / "data" / "configs" / fconfig_name
+    model_config_file = pathlib.Path(__file__).parent / "data" / "configs" / "model_config_sanity_tests.yaml"
+    train_ml4ir(working_dir.as_posix(), feature_config_file.as_posix(), model_config_file.as_posix(), log_dir.as_posix())
+    ml4ir_results = pd.read_csv(log_dir / 'test_command_line'/ '_SUCCESS', header=None)
```
lastmansleeping, line 25: [Minor] @mohazahran This should be returned by `rp.run()` as well. But this works too.

lastmansleeping:
@mohazahran 
- There are still artifacts that say "L1". You should replace them with "linear model".
- Change the `readme` to a markdown file (`.md`) so that github can render it.

python/ml4ir/applications/ranking/tests/data/L1_sanity_tests/README
```diff
@@ -0,0 +1,8 @@
+- dataset1: 2 features, linearly separable, large margin class separation
```
balikasg, line 1: This is a README file and we can be a bit more verbose if we want to actually provide info. 
- Can be a markdown table. 
- With the cols Dataset path, Nb of points, Nb Features, is linearly separable, margin of separation,..
- It should also provide some information on how these data were generated (if one needs to generate more in the future).
- How is {small, large} margins of separations are defined?

python/ml4ir/applications/ranking/tests/test_ml4ir_sanity.py
```diff
@@ -0,0 +1,200 @@
...
+    def setUp(self):
+        self.dir = pathlib.Path(__file__).parent
+        self.working_dir = TempDirectory()
+        self.log_dir = self.working_dir.makedir('logs')
+        self.working_dir.makedir('train')
+        self.working_dir.makedir('test')
+        self.working_dir.makedir('validation')
+
+    def tearDown(self):
+        TempDirectory.cleanup_all()
```
balikasg, line 100: this is very elegant and nice.

python/ml4ir/applications/ranking/tests/data/L1_sanity_tests/README
lastmansleeping: +1

lastmansleeping: I recommended even just a snapshot of the `make_classification()` method used with the arguments.

balikasg (approved):
thanks for all the effort!

lastmansleeping (approved):
Looks great!
Thanks for all the work on this @mohazahran You really did a great deep dive. :)

PR #124, opened 2021-05-20
Adding tf native fns to the feature_layer

lastmansleeping:
Adding support for using native tensorflow feature fns such as `tf.math.log` from the `feature_config` YAML. This allows us to embed such ops within the tf graph and thereby the SavedModel without writing a separate wrapper function for each.

Example feature config:
```
    feature_layer_info:
      type: numeric
      fn: tf_native_op
      args:
        ops:
          - fn: tf.math.add
            kwargs:
              y: 1.
          - fn: tf.math.log
 ```
 The above function does a natural log of the input tensor after shifting it by 1.
 
 Other minor fixes:
 - Fixing percentage metric by doing `* 100.`
 - Fixing dynamic args with booleans

### DIFF
--- a/docs/source/misc/changelog.md
+++ b/docs/source/misc/changelog.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.1] - 2021-05-20
+
+### Added
+
+- Support for using native tf/keras feature functions from the feature config YAML
+
 ## [0.1.0] - 2021-03-01
 
 ### Changed
--- a/jvm/ml4ir-inference/scripts/generate_model.sh
+++ b/jvm/ml4ir-inference/scripts/generate_model.sh
@@ -30,7 +30,7 @@ RANKING_MODEL=ml4ir/applications/ranking
 
 $EXECUTOR ${RANKING_MODEL}/pipeline.py \
     --data_dir ${RANKING_MODEL}/tests/data/csv \
-    --feature_config ${RANKING_MODEL}/tests/data/configs/feature_config.yaml \
+    --feature_config ${RANKING_MODEL}/tests/data/configs/feature_config_integration_test.yaml \
     --run_id ${1}_ranking \
     --data_format csv \
     --execution_mode train_inference_evaluate \
--- a/jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
+++ b/jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
@@ -119,7 +119,7 @@ class TensorFlowInferenceIT extends TestData {
       )
     }
     if (query.predictedScores != null) {
-      assertArrayEquals("scores aren't close enough: ", docScores, query.predictedScores, 1e-6f)
+      assertArrayEquals("scores aren't close enough: ", docScores, query.predictedScores, 1e-4f)
     }
 
   }
--- a/jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
+++ b/jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
@@ -119,6 +119,7 @@ class TensorFlowInferenceIT extends TestData {
       )
     }
     if (query.predictedScores != null) {
+      // The success threshold was set to 1e-6f, but this was too strict. So we have updated to 1e-4f
       assertArrayEquals("scores aren't close enough: ", docScores, query.predictedScores, 1e-4f)
     }
 
--- a/python/ml4ir/applications/ranking/model/metrics/metrics_helper.py
+++ b/python/ml4ir/applications/ranking/model/metrics/metrics_helper.py
@@ -188,8 +188,8 @@ def get_grouped_stats(
             label_col=label_col,
             old_rank_col=old_rank_col,
             new_rank_col=new_rank_col,
-            group_keys=set(list(group_keys)),
-            secondary_labels=set(list(secondary_labels)),
+            group_keys=group_keys,
+            secondary_labels=secondary_labels,
         )
 
     # Select clicked records
@@ -277,7 +277,7 @@ def summarize_grouped_stats(df_grouped):
     for col in df_grouped_metrics.to_dict().keys():
         if "failure" in col:
             metric_name = col[len("mean_old_") :]
-            df_grouped_metrics["perc_improv_mean_{}".format(metric_name)] = 100. * (
+            df_grouped_metrics["perc_improv_mean_{}".format(metric_name)] = (
                 df_grouped_metrics["mean_old_{}".format(metric_name)]
                 - df_grouped_metrics["mean_new_{}".format(metric_name)]
             ) / df_grouped_metrics["mean_old_{}".format(metric_name)]
--- a/python/ml4ir/applications/ranking/model/metrics/metrics_helper.py
+++ b/python/ml4ir/applications/ranking/model/metrics/metrics_helper.py
@@ -277,7 +277,7 @@ def summarize_grouped_stats(df_grouped):
     for col in df_grouped_metrics.to_dict().keys():
         if "failure" in col:
             metric_name = col[len("mean_old_") :]
-            df_grouped_metrics["perc_improv_mean_{}".format(metric_name)] = (
+            df_grouped_metrics["perc_improv_mean_{}".format(metric_name)] = 100. * (
                 df_grouped_metrics["mean_old_{}".format(metric_name)]
                 - df_grouped_metrics["mean_new_{}".format(metric_name)]
             ) / df_grouped_metrics["mean_old_{}".format(metric_name)]
--- a/python/ml4ir/applications/ranking/model/metrics/metrics_helper.py
+++ b/python/ml4ir/applications/ranking/model/metrics/metrics_helper.py
@@ -188,8 +188,8 @@ def get_grouped_stats(
             label_col=label_col,
             old_rank_col=old_rank_col,
             new_rank_col=new_rank_col,
-            group_keys=group_keys,
-            secondary_labels=secondary_labels,
+            group_keys=set(list(group_keys)),
+            secondary_labels=set(list(secondary_labels)),
         )
 
     # Select clicked records
@@ -277,7 +277,7 @@ def summarize_grouped_stats(df_grouped):
     for col in df_grouped_metrics.to_dict().keys():
         if "failure" in col:
             metric_name = col[len("mean_old_") :]
-            df_grouped_metrics["perc_improv_mean_{}".format(metric_name)] = (
+            df_grouped_metrics["perc_improv_mean_{}".format(metric_name)] = 100. * (
                 df_grouped_metrics["mean_old_{}".format(metric_name)]
                 - df_grouped_metrics["mean_new_{}".format(metric_name)]
             ) / df_grouped_metrics["mean_old_{}".format(metric_name)]
--- a/python/ml4ir/applications/ranking/model/ranking_model.py
+++ b/python/ml4ir/applications/ranking/model/ranking_model.py
@@ -156,10 +156,10 @@ class RankingModel(RelevanceModel):
                 label_col=self.feature_config.get_label("node_name"),
                 old_rank_col=self.feature_config.get_rank("node_name"),
                 new_rank_col=RankingConstants.NEW_RANK,
-                group_keys=self.feature_config.get_group_metrics_keys(
-                    "node_name"),
-                secondary_labels=self.feature_config.get_secondary_labels(
-                    "node_name"),
+                group_keys=list(set(self.feature_config.get_group_metrics_keys(
+                    "node_name"))),
+                secondary_labels=list(set(self.feature_config.get_secondary_labels(
+                    "node_name"))),
             )
             if df_grouped_stats.empty:
                 df_grouped_stats = df_batch_grouped_stats
--- a/python/ml4ir/applications/ranking/tests/data/configs/feature_config.yaml
+++ b/python/ml4ir/applications/ranking/tests/data/configs/feature_config.yaml
@@ -66,13 +66,13 @@ features:
       args:
         ops:
           - fn: tf.math.add
-            kwargs:
+            args:
               y: 0.
           - fn: tf.math.subtract  # The goal here is to see the end-to-end functionality of tf_native_op without modifying the tests
-            kwargs:
+            args:
               y: 0.
           - fn: tf.clip_by_value
-            kwargs:
+            args:
               clip_value_min: 0.
               clip_value_max: 1000000.
     serving_info:
--- a/python/ml4ir/applications/ranking/tests/data/configs/feature_config.yaml
+++ b/python/ml4ir/applications/ranking/tests/data/configs/feature_config.yaml
@@ -64,12 +64,13 @@ features:
       shape: null
       fn: tf_native_op
       args:
-        - fn: tf.math.add
-          kwargs:
-            y: 1.
-        - fn: tf.math.subtract  # The goal here is to see the end-to-end functionality of tf_native_op without modifying the tests
-          kwargs:
-            y: 1.
+        ops:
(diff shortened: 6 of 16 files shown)

### CODE REVIEW

File: python/ml4ir/applications/ranking/tests/data/configs/feature_config.yaml [hunk 1]
```diff
@@ -62,6 +62,15 @@ features:
...
       shape: null
+      fn: tf_native_op
+      args:
+        ops:
+          - fn: tf.math.add
+            kwargs:
+              y: 1.
+          - fn: tf.math.subtract  # The goal here is to see the end-to-end functionality of tf_native_op without modifying the tests
+            kwargs:
+              y: 1.
```
lastmansleeping, line 73: This will run both the python and python-jvm end-to-end tests for verification with the tf_native_op.

lastmansleeping:
@jakemannix @ducouloa Ah come on! the end to end test is failing. This is a great thing, but...
```
scores aren't close enough: : arrays first differed at element [1]; expected:<0.48640153> but was:<0.48640254>
```

python/ml4ir/applications/ranking/tests/test_transfer_learning.py
```diff
@@ -100,7 +100,7 @@ def test_model_saving(self):
             for f in glob.glob(os.path.join(self.args.models_dir, "final", "layers", "*"))
         ]
         for layer in model.model.layers:
-            assert "{}.npz".format(layer.name) in saved_layers
+            assert "{}.npz".format(layer.name.split("/")[0]) in saved_layers
```
lastmansleeping, line 103: There are some nested layers. This is a more robust check.

python/ml4ir/base/config/dynamic_args.py
```diff
@@ -28,6 +28,11 @@ def cast_dynamic_val(val: str):
                 # int
                 return int(val)
         elif val.isalnum():
+            # bool
+            if val.lower() == "true":
+                return True
+            elif val.lower() == "false":
+                return False
```
lastmansleeping, line 35: This allows me to not have 10 different feature configs for doing feature ablation studies. I can just set `--feature_config.features.feature_name.trainable false` and the feature will be removed from the model.

lastmansleeping: (I understand this is not related to the PR, but leaving here as it is a tiny change)

python/ml4ir/applications/ranking/tests/data/configs/feature_config.yaml
jakemannix: is there any way to carefully verify that the model is actually using these tf functions in the jvm side?  I guess the "check that the ranking and classification scores are the same within 1e-6" test does that?

lastmansleeping: Yes, it should. Here, I've set them to effectively cancel out and no do anything effective because I didn't want to modify all the other tests. I can do that if you think that's better.

python/ml4ir/applications/ranking/tests/data/configs/feature_config.yaml
```diff
@@ -62,6 +62,19 @@ features:
     feature_layer_info:
       type: numeric
       shape: null
+      fn: tf_native_op
+      args:
+        ops:
+          - fn: tf.math.add
+            kwargs:
+              y: 0.
+          - fn: tf.math.subtract  # The goal here is to see the end-to-end functionality of tf_native_op without modifying the tests
```
jakemannix, line 71: so if we add then subtract 0, do we notice?

also, clip_by_value between 0 and 1M?  will our unit tests ever see data outside that range in the first place?

lastmansleeping: No, my goal was to mostly see if this would lead to an error. I commented above, this does not do anything to change the feature values or the resulting scores. I did this because I didn't want to modify all the other gold values from other tests. But I can do that, just to be safe.

jvm/ml4ir-inference/scripts/generate_model.sh
```diff
@@ -30,7 +30,7 @@ RANKING_MODEL=ml4ir/applications/ranking
 
 $EXECUTOR ${RANKING_MODEL}/pipeline.py \
     --data_dir ${RANKING_MODEL}/tests/data/csv \
-    --feature_config ${RANKING_MODEL}/tests/data/configs/feature_config.yaml \
+    --feature_config ${RANKING_MODEL}/tests/data/configs/feature_config_integration_test.yaml \
```
lastmansleeping, line 33: @ducouloa @jakemannix I've created a separate feature config to be used for python-jvm integration test. The goal is to make sure this contains the most extensive version of the feature specification.

python/ml4ir/applications/ranking/model/metrics/metrics_helper.py
```diff
@@ -277,7 +277,7 @@ def summarize_grouped_stats(df_grouped):
     for col in df_grouped_metrics.to_dict().keys():
         if "failure" in col:
             metric_name = col[len("mean_old_") :]
-            df_grouped_metrics["perc_improv_mean_{}".format(metric_name)] = (
+            df_grouped_metrics["perc_improv_mean_{}".format(metric_name)] = 100. * (
```
lastmansleeping, line 280: [Minor] percentage metric did not have a `100. *`

python/ml4ir/applications/ranking/tests/data/configs/feature_config_integration_test.yaml
```diff
@@ -0,0 +1,170 @@
...
+      fn: tf_native_op
+      args:
+        ops:
+          - fn: tf.math.add
+            kwargs:
+              y: 1.
+          - fn: tf.clip_by_value
+            kwargs:
+              clip_value_min: 0.
+              clip_value_max: 10.
```
lastmansleeping, line 74: @jakemannix FYI I've made a disruptive change to the feature layer here, which should be captured in the integration test.

python/ml4ir/applications/ranking/tests/test_metrics.py
```diff
@@ -26,12 +26,12 @@
...
-                'perc_improv_mean_name_match_failure_all_rank': -0.380,
-                'perc_improv_mean_name_match_failure_any_rank': -1.244,
-                'perc_improv_mean_name_match_failure_any_count': -1.068,
-                'perc_improv_mean_name_match_failure_any_fraction': -0.754}
+                'perc_improv_mean_name_match_failure_all': -40.217,
+                'perc_improv_mean_name_match_failure_any': -96.407,
+                'perc_improv_mean_name_match_failure_all_rank': -38.028,
+                'perc_improv_mean_name_match_failure_any_rank': -124.478,
+                'perc_improv_mean_name_match_failure_any_count': -106.854,
+                'perc_improv_mean_name_match_failure_any_fraction': -75.401}
```
lastmansleeping, line 34: [Minor] This is because I fixed the percentage metric by doing `100. *`

python/ml4ir/applications/ranking/tests/test_serving.py
```diff
@@ -19,6 +19,7 @@ def test_model_serving(self):
 
         # Test model training on TFRecord SequenceExample data
         data_dir = os.path.join(self.root_data_dir, "tfrecord")
+        self.feature_config_fname = "feature_config_integration_test.yaml"
```
lastmansleeping, line 22: @jakemannix @ducouloa Using the same feature config for both python-jvm integration tests and internal python(training)-python(serving) integration test.

python/ml4ir/base/features/feature_fns/tf_native.py
```diff
@@ -0,0 +1,59 @@
+import tensorflow as tf
+
+from ml4ir.base.io.file_io import FileIO
+
+
+def tf_native_op(feature_tensor: tf.Tensor, feature_info: dict, file_io: FileIO):
```
lastmansleeping, line 6: This is the main addition of this PR.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -119,7 +119,7 @@ class TensorFlowInferenceIT extends TestData {
       )
     }
     if (query.predictedScores != null) {
-      assertArrayEquals("scores aren't close enough: ", docScores, query.predictedScores, 1e-6f)
+      assertArrayEquals("scores aren't close enough: ", docScores, query.predictedScores, 1e-4f)
```
lastmansleeping, line 122: @jakemannix @ducouloa FYI

lastmansleeping: Was failing for this reason
```
org.junit.internal.ArrayComparisonFailure: scores aren't close enough: : arrays first differed at element [0]; expected:<0.63819975> but was:<0.63819844>
```

balikasg: do we understand why? You may want to add an inline comments saying we had 1e-6f but it was too much.

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -156,10 +156,10 @@ def evaluate(
                 label_col=self.feature_config.get_label("node_name"),
                 old_rank_col=self.feature_config.get_rank("node_name"),
                 new_rank_col=RankingConstants.NEW_RANK,
-                group_keys=self.feature_config.get_group_metrics_keys(
-                    "node_name"),
-                secondary_labels=self.feature_config.get_secondary_labels(
-                    "node_name"),
+                group_keys=list(set(self.feature_config.get_group_metrics_keys(
```
balikasg, line 159: is it expected to have duplicates and you are taking the `set`? is it a sign of an error?

lastmansleeping: I think I know what the issue is. I can try and fix it, but it will be messy. I think we should keep it as a separate PR. I will file a bug for this if that works for you. This small change suppresses a lot of pandas warnings for now.

I can also remove this change from this PR and keep it in that bug itself, if that's better.

python/ml4ir/applications/ranking/tests/test_feature_layer.py
```diff
@@ -686,3 +688,30 @@ def test_global_1d_pooling(self):
...
+                            "ops": [
+                                {"fn": "tf.math.add", "kwargs": {"y": 1.}},
+                                {"fn": "tf.math.log"}
+                            ]
+                        }
+                    }
+                },
+                file_io=None
+            )
+        expected_tensor = tf.expand_dims(tf.math.log(tf.math.add(input_tensor, 1.)), axis=-1)
```
balikasg, line 715: why are we `expand_dims` here? batch?

lastmansleeping: Even I was confused why I had to do it, then I realized that we have an `expand_dims` in feature_layer, but it is gated with those features that do not have a feature fn specified -> https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/base/features/feature_layer.py#L171

I did try removing the `if` condition, but it breaks every other feature fn like embedding, LSTM etc. I will need to modify everyone of them. I didn't want to devolve this PR into that at the time being. So I am adhering to the `expand_dims` we have in other functions as well.

python/ml4ir/base/features/feature_fns/tf_native.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    feature_node_name = feature_info.get("node_name", feature_info.get("name"))
+    tf_ops = feature_info.get("feature_layer_info", {}).get("args", {}).get("ops", {})
+
+    if not tf_ops:
+        return feature_tensor
+
+    for tf_op in tf_ops:
+        try:
+            fn_, kwargs = eval(tf_op["fn"]), tf_op.get("kwargs", {})
```
balikasg, line 47: I think the usage of `kwargs` is problematic here. This term has a somewhat special meaning in python, and what we are passing are not really `kwargs` but actual required arguments in these functions. You probably do it because we are passing them as a dictionary but I am afraid we are overloading and already overloaded term. Wouldn't `args` or `fn_args` make it simpler or at least minimize the risk of understanding something else?

balikasg: I see how you are using it below, I get it.

balikasg: but essentially these are not kwargs (at least not only kwargs)

docs/source/misc/changelog.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.1] - 2021-05-20
+
+### Added
+
+- Support for using native tf/keras feature functions from the feature config YAML
```
ducouloa, line 12: 👍

jvm/ml4ir-inference/scripts/generate_model.sh
ducouloa: Thanks for the callout

python/ml4ir/base/features/feature_fns/tf_native.py
lastmansleeping: Yes, I can definitely rename to `args`

python/ml4ir/base/features/feature_fns/tf_native.py
```diff
@@ -0,0 +1,59 @@
...
+        except AttributeError as e:
+            raise KeyError("Invalid fn specified for tf_native_op : {}\n{}".format(tf_op["fn"], e))
+        
+        try:
+            feature_tensor = fn_(feature_tensor, **kwargs)
+        except Exception as e:
+            raise Exception("Error while applying {} to {} feature:\n{}".format(tf_op["fn"], feature_node_name, e))
```
balikasg, line 54: Nice error messages!

docs/source/misc/changelog.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.1] - 2021-05-20
+
+### Added
```
balikasg, line 10: do you think we should detail this to any other part of the README or are we good?

python/ml4ir/applications/ranking/model/ranking_model.py
balikasg: It works. I let you decide what you believe the best approach is. If you go with keeping it, just make a note to the bug to simplify these lines once we solve it.

docs/source/misc/changelog.md
lastmansleeping: The neat thing about this is I've also added it here -> https://ml4ir.readthedocs.io/en/latest/misc/changelog.html

We can have a version of this in the main python README as well, but that would mean we will probably need to maintain two versions.

python/ml4ir/applications/ranking/tests/test_feature_layer.py
balikasg: We do it there to support for ranking type of datasets, where a query can be coupled with several retrieved docs (if I remember well).. Yes, it is a tough thing to remember that one, it is everywhere, even in the metrics..

python/ml4ir/base/features/feature_fns/tf_native.py
ducouloa: Do we have to also add a short paragraph on the public doc (https://ml4ir.readthedocs.io/en/latest/) about this new capability?

ducouloa (dismissed):

python/ml4ir/base/features/feature_fns/tf_native.py
lastmansleeping: Good call. Let me do that.

lastmansleeping: Good call. Let me do that.

python/ml4ir/applications/ranking/tests/test_feature_layer.py
lastmansleeping: Yes, that is the reason.

ducouloa (approved):

balikasg (approved):

PR #126, opened 2021-05-28
Mo/positional bias handling (MLranker)

mohazahran:
Adding MLranker's method to handle positional bias.
To trigger it add this section to the model_config.yaml file

```
positional_bias_handler:
  key: fixed_additive_positional_bias
  max_ranks_count: 25
```
 
Example:

```
architecture_key: dnn
layers:
  - type: dense
    name: first_dense
    units: 64
    activation: relu
  - type: dropout
    name: first_dropout
    rate: 0.0
  - type: dense
    name: second_dense
    units: 16
    activation: relu
  - type: dropout
    name: second_dropout
    rate: 0.0
  - type: dense
    name: final_dense
    units: 1
    activation: 

positional_bias_handler:
  key: fixed_additive_positional_bias
  max_ranks_count: 25
  
optimizer:
  key: adam
  gradient_clip_value: 5.0
lr_schedule:
  key: constant
  learning_rate: 0.01
```


python/ml4ir/base/data/relevance_dataset.py
```diff
@@ -165,6 +165,12 @@ def create_dataset(self, parse_tfrecord=True):
             as it improved train/test/validation throughput
             by 30% in some real model training.
             """
+            # Setting the one-hot mask of the fr feature to be off for training
```
balikasg, line 168: what does `fr` stand for?

python/ml4ir/base/features/feature_config.py
```diff
@@ -327,6 +347,26 @@ def get_feature(self, name: str):
 
         raise KeyError("No feature named {} in FeatureConfig".format(name))
 
+    def feature_exists(self, name: str, trainable=True):
```
balikasg, line 350: this function does more than checking if a feature simply exists. Also, some params are missing in the docstring. 
I am wondering if there is value in making it two functions `trainable_feature_exists` and `not_trainable_feature_exists`.

python/ml4ir/base/features/preprocessing.py
```diff
@@ -290,6 +291,34 @@ def convert_label_to_clicks(label_vector, dtype):
...
+
+            Returns
+            -------
+            tf tensor
+                one hot vector representation of fr
+    """
+
+    if mask_fr:
+        one_hot_fr = tf.zeros(max_ranks, dtype=tf.dtypes.int64)
+        #one_hot_fr = tf.one_hot(max_ranks+1, depth=max_ranks)
```
balikasg, line 316: consider removing the commented code in the final version.

python/ml4ir/base/features/preprocessing.py
```diff
@@ -290,6 +291,34 @@ def convert_label_to_clicks(label_vector, dtype):
     clicks = tf.dtypes.cast(cond, typ)
     return clicks
 
+@tf.function
+def convert_fr_to_one_hot(fr_value, max_ranks, mask_fr):
+    """Convert the label vector to binary clicks. Documents with the maximum labels are considered clicked and receive
```
balikasg, line 296: An example on this docstring would help. 
Also, your `max_ranks` variable explanation is very good. Why not call it `max_docs_per_query` directly?

lastmansleeping (commented):
Thanks for the early review @balikasg . Mo had created this PR because we were debugging something and we wanted to have same code. 

Good comments overall, worth addressing @mohazahran

python/ml4ir/applications/ranking/config/keys.py
```diff
@@ -35,3 +35,9 @@ class LossTypeKey(Key):
     POINTWISE = "pointwise"
     PAIRWISE = "pairwise"
     LISTWISE = "listwise"
+
+
+class PositionalBiasHandler(Key):
+    """Types of models to handle positional bias"""
+
+    MLRANKER = "mlranker"
```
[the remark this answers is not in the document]
lastmansleeping: 100% agree.

`MLRanker` is probably not even a published work. So referencing in `ml4ir` should be a no-go.

I recommend something like `scalar_rank_bias`

python/ml4ir/applications/ranking/tests/test_fr_one_hot_conversion.py
```diff
@@ -0,0 +1,27 @@
...
+import warnings
+from ml4ir.base.features.preprocessing import convert_fr_to_one_hot
+import numpy as np
+
+warnings.filterwarnings("ignore")
+
+
+class TestFrOneHotConversion(unittest.TestCase):
+
+    def fr_one_hot_conversion(self, fr_value, max_ranks, mask_fr):
```
lastmansleeping, line 11: @mohazahran `fr` is internal lingo and you shouldn't use this in `ml4ir`. Why not just use the feature config `rank` attribute instead. It will get you this and was designed for this exact purpose in fact.

python/ml4ir/base/features/feature_config.py
```diff
@@ -938,6 +978,10 @@ def get_shape(feature_info: dict):
             # Setting size to None for sequence features as the num_records is variable
             if feature_info["tfrecord_type"] == SequenceExampleTypeKey.SEQUENCE:
                 return feature_info.get("shape", (None,))
+            # The fr feature size is pre-defined to max_ranks
+            elif feature_info["node_name"] == "fr_feature":
```
lastmansleeping, line 982: @mohazahran Remove `fr` here as well. You should call it `rank`

lastmansleeping: You are also doing a check on a `node_name`. I don't think this is good coding practice. We are trying to keep `ml4ir` as generic as possible and the only hard-coded keys we expect are `query_key`, `rank`, `label` and `features`.

python/ml4ir/base/model/scoring/interaction_model.py
```diff
@@ -35,9 +35,13 @@ def __call__(self, inputs: Dict[str, Input]):
             computing custom metrics and losses
         """
         train_features, metadata_features = self.feature_layer_op(inputs)
-        train_features, metadata_features = self.transform_features_op(
-            train_features, metadata_features
-        )
+        # separate the default learnable features from the fr_features
+        if self.feature_config.feature_exists('fr', trainable=True):
+            train_fr_features, _ = self.transform_features_op({'fr_feature':train_features['fr_feature']}, metadata_features)
```
lastmansleeping, line 40: @mohazahran Can you explain why you are doing this?

python/ml4ir/base/pipeline.py
```diff
@@ -146,6 +149,24 @@ def __init__(self, args: Namespace):
...
+              'log_at_inference': True,
+              'feature_layer_info': {'type': 'numeric', 'shape': None},
+              'serving_info': {'name': 'fr_feature', 'required': True},
+              'tfrecord_type': 'context',
+              'preprocessing_info': [{'fn': 'convert_fr_to_one_hot','args': {'max_ranks': 25, 'mask_fr': True}}]}
+            feature_config_dict['features'].append(fr_feature)
+
+
+
+
```
lastmansleeping, line 169: @mohazahran Why do you need this? I am very confused by the `fr_feature` feature. Why not dynamically compute it during activation/loss computation?

python/ml4ir/base/features/preprocessing.py
```diff
@@ -290,6 +291,34 @@ def convert_label_to_clicks(label_vector, dtype):
     clicks = tf.dtypes.cast(cond, typ)
     return clicks
 
+@tf.function
+def convert_fr_to_one_hot(fr_value, max_ranks, mask_fr):
```
lastmansleeping, line 295: @mohazahran I don't think you should do this as a preprocessing step. Let's discuss offline if you have questions.

python/ml4ir/base/model/architectures/dnn.py
```diff
@@ -61,18 +77,35 @@ def get_op(layer_type, layer_args):
...
+                    layer_input = layer_op(layer_input)
+
+                # Collapse extra dimensions
+                if isinstance(self.layer_ops[-1], layers.Dense) and (self.layer_ops[-1].units == 1):
+                    scores = tf.squeeze(layer_input, axis=-1)
+                else:
+                    scores = layer_input
 
-            return scores
+                return scores
```
lastmansleeping, line 109: @mohazahran You should move the actual implementation of the positional bias to a module under `applications/ranking/model/positional_bias`.

python/ml4ir/base/model/architectures/dnn.py
```diff
@@ -61,18 +77,35 @@ def get_op(layer_type, layer_args):
...
-            # Pass ranking features through all the layers of the DNN
-            for layer_op in self.layer_ops:
-                layer_input = layer_op(layer_input)
+                #positional_bias = self.positional_bias_layer(ranking_features[1])
+                positional_bias = self.positional_bias_layer(tf.transpose(ranking_features[1], [0, 2, 1]))
+                scores = layers.Add()([positional_bias, layer_input])
+
+                # Collapse extra dimensions
+                if isinstance(self.layer_ops[-1], layers.Dense) and (self.layer_ops[-1].units == 1):
+                    scores = tf.squeeze(scores, axis=-1)
```
lastmansleeping, line 92: @mohazahran This seems like common between the if and else blocks, no?

python/ml4ir/base/model/architectures/dnn.py
```diff
@@ -21,7 +22,22 @@ class DNNLayer:
...
+                                            name="positional_bias_layer",
+                                            activation=None,
+                                            use_bias=False,
+                                            kernel_initializer="glorot_uniform",
+                                            bias_initializer="zeros",
+                                            kernel_regularizer=None,
+                                            bias_regularizer=None,
+                                            activity_regularizer=None,
+                                            kernel_constraint=None,
+                                            bias_constraint=None)
```
lastmansleeping, line 40: @mohazahran These are defaults anyway, right? Why fill them out explicitly?

python/ml4ir/base/model/scoring/interaction_model.py
```diff
@@ -35,9 +35,13 @@ def __call__(self, inputs: Dict[str, Input]):
...
+            train_features, metadata_features = self.transform_features_op({k: train_features[k] for k in train_features if k != 'fr_feature'}, metadata_features)
+            return [train_features, train_fr_features], metadata_features
```
lastmansleeping, line 42: @mohazahran Definitely get rid of this. Why are you returning these two? More importantly, what are these two? Keep in mind you are also introducing inconsistent number of return values within this method. Overall, I don't think we need this `fr_feature` and we could compute it on the fly within the positional bias layer itself.

python/ml4ir/base/model/scoring/interaction_model.py
lastmansleeping, line 40: @mohazahran I have a doubt that this does what you expect it to do. You define a preprocessing function and are then trying to use it from the interaction_model. The interaction_model requires functions to be specified in the `feature_layer_info` part of the config YAML.

lastmansleeping (changes requested):
Okay, one more critical change (or clarification)
- positional bias handling should only be applied during training and not during testing. Looks like this is not being handled currently (or I couldn't find it, at least)

Check [this implementation of dropout in embeddings](https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/base/features/feature_fns/categorical.py#L274_L350) I had done a while back for how to define a layer and gate it within a `call(training=None)` method. This way, `tensorflow` automatically passes `training=True` only when you call `model.fit(...)` and `False` rest of the time (ala inference). This is the most straightforward and unit testable way to do this.

One thing you've done already, which I love is define positional bias as a layer in the DNN. I think that is great! Now you need to make sure you combine the above code snippet with yours (and add one-hot inside it) and then it should all be self contained. What do you think?

python/ml4ir/base/pipeline.py
```diff
@@ -146,6 +149,24 @@ def __init__(self, args: Namespace):
...
+            {
+              'name': 'fr',
+              'node_name': 'fr_feature',
+              'trainable': True,
+              'dtype': 'int64',
+              'log_at_inference': True,
+              'feature_layer_info': {'type': 'numeric', 'shape': None},
+              'serving_info': {'name': 'fr_feature', 'required': True},
+              'tfrecord_type': 'context',
+              'preprocessing_info': [{'fn': 'convert_fr_to_one_hot','args': {'max_ranks': 25, 'mask_fr': True}}]}
```
lastmansleeping, line 164: @mohazahran See, you are hardcoding `25` here. You should get out of the habit of doing that. Check out the `max_sequence_size` argument.

python/ml4ir/base/data/relevance_dataset.py
mohazahran: final rank. It's one of the logged features indicating at which rank this document was shown

python/ml4ir/applications/ranking/tests/test_fr_one_hot_conversion.py
mohazahran: I didn't know that. I can look into it.

python/ml4ir/base/features/feature_config.py
mohazahran: I'm checking the node's name because the 'fr' column is being used now as a rank and as a feature for positional bias.

python/ml4ir/base/model/scoring/interaction_model.py
mohazahran: Don't want to merge the normal features with the positional bias features.

python/ml4ir/base/model/architectures/dnn.py
mohazahran: These were for sanity during development.

python/ml4ir/base/model/scoring/interaction_model.py
mohazahran: I'm returning two because I'm separating the normal features (sim score, pv ...) from the positional bias features (fr)

python/ml4ir/base/model/scoring/interaction_model.py
mohazahran: I believe it should work because the way I set it up is the pre-processing function converts fr to one-hot encoded vector for training and mask the fr for validation and testing. So the input feature to the mode is already processed and don't need any further intervention.

python/ml4ir/base/pipeline.py
mohazahran: Sure, that can be easily changed, I believe I overlooked it while debugging.

python/ml4ir/applications/ranking/tests/test_fixed_additive_positional_bias.py
```diff
@@ -0,0 +1,32 @@
+import unittest
+import warnings
+import tensorflow as tf
+from ml4ir.base.model.architectures.fixed_additive_positional_bias import FixedAdditivePositionalBias
+import numpy as np
+
+warnings.filterwarnings("ignore")
+
+
+class TestFixedAdditivePositionalBias(unittest.TestCase):
```
lastmansleeping, line 10: @mohazahran Worth adding a test for train vs test time difference.

python/ml4ir/base/model/architectures/dnn.py
```diff
@@ -21,7 +23,12 @@ class DNNLayer:
 class DNN:
     def __init__(self, model_config: dict, feature_config: FeatureConfig, file_io):
         self.file_io: FileIO = file_io
+        self.model_config = model_config
+        self.feature_config = feature_config
         self.layer_ops: List = self.define_architecture(model_config, feature_config)
+        if 'positional_bias_handler' in self.model_config and self.model_config['positional_bias_handler'][
+            'key'] == PositionalBiasHandler.FIXED_ADDITIVE_POSITIONAL_BIAS:
+            self.positional_bias_layer = FixedAdditivePositionalBias()
```
lastmansleeping, line 31: @mohazahran For the future, it might be better to create a separate factory method that does this. Will be decoupled from DNN implementation and will have a neat lookup table for all positional bias implementations.

lastmansleeping: It's okay to not do this for now though I think. But make sure you nest the if instead of and here.

python/ml4ir/base/model/architectures/dnn.py
```diff
@@ -60,13 +67,21 @@ def get_op(layer_type, layer_args):
...
                 layer_input = layer_op(layer_input)
 
+            if 'positional_bias_handler' in self.model_config and \
+                    self.model_config['positional_bias_handler']['key'] == \
+                    PositionalBiasHandler.FIXED_ADDITIVE_POSITIONAL_BIAS:
+
+                positional_bias = self.positional_bias_layer(metadata_features[self.feature_config.get_rank()['name']],
+                                                             max_ranks=self.model_config['positional_bias_handler']['max_ranks_count'], training=True)
+                layer_input = layers.Add()([positional_bias, layer_input])
+
```
lastmansleeping, line 84: @mohazahran Why do we need this if we are adding it to DNN layer?

lastmansleeping: Or rather, why do we need the one in the constructor? Just create the object here.

python/ml4ir/base/model/architectures/dnn.py
```diff
@@ -21,7 +23,12 @@ class DNNLayer:
 class DNN:
     def __init__(self, model_config: dict, feature_config: FeatureConfig, file_io):
         self.file_io: FileIO = file_io
+        self.model_config = model_config
+        self.feature_config = feature_config
         self.layer_ops: List = self.define_architecture(model_config, feature_config)
+        if 'positional_bias_handler' in self.model_config and self.model_config['positional_bias_handler'][
```
lastmansleeping, line 29: @mohazahran We discussed to move the positional bias under `layers` and thereby under `define_architecture()`, no? That would be clean I think as it is in essence a layer of the DNN.

lastmansleeping: The added advantage is that it will automatically get added in `get_architecture_op`

lastmansleeping: You can add `positional_bias` to `DNNLayer` keys as well.

python/ml4ir/base/model/architectures/dnn.py
```diff
@@ -60,13 +67,21 @@ def get_op(layer_type, layer_args):
...
 
             # Pass ranking features through all the layers of the DNN
             for layer_op in self.layer_ops:
                 layer_input = layer_op(layer_input)
 
+            if 'positional_bias_handler' in self.model_config and \
+                    self.model_config['positional_bias_handler']['key'] == \
+                    PositionalBiasHandler.FIXED_ADDITIVE_POSITIONAL_BIAS:
+
+                positional_bias = self.positional_bias_layer(metadata_features[self.feature_config.get_rank()['name']],
```
lastmansleeping, line 81: @mohazahran You should replace this with `node_name`.

python/ml4ir/base/model/architectures/fixed_additive_positional_bias.py
```diff
@@ -0,0 +1,60 @@
+import tensorflow as tf
```
lastmansleeping, line 1: @mohazahran You should move this to under ranking directory.

python/ml4ir/base/model/architectures/fixed_additive_positional_bias.py
```diff
@@ -0,0 +1,60 @@
...
+        Returns
+        -------
+        Tensor object
+            A one-hot vector representation of the input rank index. The one-hot entry would
+            be masked in case of non training mode.
+        """
+        features = tf.one_hot(tf.cast(tf.subtract(inputs, 1), dtype=tf.int64), depth=max_ranks, dtype=tf.dtypes.float32)
+        if not training:
+            features = tf.multiply(features, 0.0)
+        return features
```
lastmansleeping, line 38: @mohazahran Why is this a separate function? Can be added to `call()`, no? Just a few lines.

python/ml4ir/base/model/architectures/dnn.py
```diff
@@ -60,13 +67,21 @@ def get_op(layer_type, layer_args):
...
+                                                             max_ranks=self.model_config['positional_bias_handler']['max_ranks_count'], training=True)
```
lastmansleeping, line 82: Ooooh found a bug after breaking my head over this for a long time. You are hardcoding `training=True` here. You shouldn't. Tensorflow/keras will automatically send the right value when you call `fit()` vs `predict()`. What this does is basically hard set it to training=True, regardless of training vs inference - dangerous.

python/ml4ir/base/model/architectures/dnn.py
lastmansleeping, line 31:
```suggestion
            self.positional_bias_layer = FixedAdditivePositionalBias(max_ranks=self.model_config['positional_bias_handler']['max_ranks_count'])
```

lastmansleeping (changes requested):
Positional bias is not working as expected during inference. I have left in line comments explaining and also the changes needed to fix. Let me know if you have questions.

python/ml4ir/base/model/architectures/fixed_additive_positional_bias.py
```diff
@@ -0,0 +1,60 @@
...
+        training : bool
+            If the layer should be run as training or not
+
+        Returns
+        -------
+        Tensor object
+            positional biases resulting from a feedforwrd to the converted one hot tensor through a dense layer
+        """
+        features = self.convert_to_one_hot(inputs, max_ranks, training)
+        return self.dense(features)
```
lastmansleeping, line 59:
```suggestion
        if training:
            features = self.convert_to_one_hot(inputs)
            return self.dense(features)
        else:
            return tf.expand_dims(tf.zeros_like(inputs), axis=-1)
```

Explanation:
```
In [122]: tf.keras.layers.Dense(1)(np.random.randn(128, 25, 25)).shape
Out[122]: TensorShape([128, 25, 1])
```

whereas, rank is of shape `[128, 25]`. So this leads to discrepancy at train vs test time.

python/ml4ir/base/model/architectures/dnn.py
mohazahran: Yeah but we discussed that doing so would put responsibility on the user to specify that the positional bias layer must be the last layer. I don't think it's a good thing to do.

python/ml4ir/base/model/architectures/fixed_additive_positional_bias.py
mohazahran: I'm not sure that I follow because after the one-hot conversion for both training and testing they both go through the `self.dense` so they both should have similar shapes, no?

python/ml4ir/base/model/architectures/fixed_additive_positional_bias.py
mohazahran: Do you have specific dir in mind because to me it fits under architecture.

python/ml4ir/base/model/architectures/dnn.py
```diff
@@ -60,13 +67,20 @@ def get_op(layer_type, layer_args):
...
     def get_architecture_op(self):
-        def _architecture_op(ranking_features):
+        def _architecture_op(ranking_features, metadata_features):
             layer_input = ranking_features
 
             # Pass ranking features through all the layers of the DNN
             for layer_op in self.layer_ops:
                 layer_input = layer_op(layer_input)
 
+            if 'positional_bias_handler' in self.model_config and \
```
lastmansleeping, line 77: @mohazahran Please define this in `ModelConfigKeys`

python/ml4ir/base/model/architectures/fixed_additive_positional_bias.py
```diff
@@ -0,0 +1,60 @@
...
+    def convert_to_one_hot(self, inputs, training=False):
+        """
+        Convert the iuputs to one hot vectors.
+
+        Parameters
+        ----------
+        input : Tensor object
+            rank index tensor
+        max_ranks: int
+            the maximum number of documents per query
```
lastmansleeping, line 26: @mohazahran Remove this.

python/ml4ir/base/model/architectures/fixed_additive_positional_bias.py
```diff
@@ -0,0 +1,60 @@
...
+        training : bool
+            If the layer should be run as training or not
+
+        Returns
+        -------
+        Tensor object
+            positional biases resulting from a feedforwrd to the converted one hot tensor through a dense layer
+        """
+        features = self.convert_to_one_hot(inputs, training)
```
lastmansleeping, line 59: @mohazahran Move the one hot function here itself. It is just a one line method otherwise.

python/ml4ir/applications/ranking/tests/test_fixed_additive_positional_bias.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def one_hot_conversion(self, rank_index, max_ranks, training):
```
lastmansleeping, line 12: @mohazahran You should name this method better. It is not just doing one hot conversion. I see asserts here as well.

python/ml4ir/applications/ranking/tests/test_fixed_additive_positional_bias.py
```diff
@@ -0,0 +1,31 @@
+import unittest
+import warnings
+import tensorflow as tf
+from ml4ir.base.model.architectures.fixed_additive_positional_bias import FixedAdditivePositionalBias
+import numpy as np
+
+warnings.filterwarnings("ignore")
+
+
+class TestFixedAdditivePositionalBias(unittest.TestCase):
```
lastmansleeping, line 10: @mohazahran Can you also add a test for the main `call()` method. I recommended moving the one hot inside the `call()` method itself at another place.

python/ml4ir/base/model/architectures/dnn.py
```diff
@@ -21,7 +23,12 @@ class DNNLayer:
 class DNN:
     def __init__(self, model_config: dict, feature_config: FeatureConfig, file_io):
         self.file_io: FileIO = file_io
+        self.model_config = model_config
+        self.feature_config = feature_config
         self.layer_ops: List = self.define_architecture(model_config, feature_config)
+        if 'positional_bias_handler' in self.model_config and self.model_config['positional_bias_handler'][
+            'key'] == PositionalBiasHandler.FIXED_ADDITIVE_POSITIONAL_BIAS:
+            self.positional_bias_layer = FixedAdditivePositionalBias(max_ranks=self.model_config['positional_bias_handler']['max_ranks_count'])
```
lastmansleeping, line 31: [Minor] @mohazahran Maybe just `max_ranks` in both places?

lastmansleeping (approved):
LGTM

python/ml4ir/applications/ranking/tests/test_fixed_additive_positional_bias.py
```diff
@@ -0,0 +1,31 @@
...
+import tensorflow as tf
+from ml4ir.base.model.architectures.fixed_additive_positional_bias import FixedAdditivePositionalBias
+
+warnings.filterwarnings("ignore")
+
+
+class TestFixedAdditivePositionalBias(unittest.TestCase):
+
+    def calling_additive_positional_bias(self, rank_index, max_ranks, training):
+        """Testing to invoke additive positional bias"""
```
balikasg, line 12: This is a minor comment, feel free to ignore if you think is of no added value. We should improve as a team on what we write and how we write docstrings. If we are to take it one step at a time, we should at least use action verbs..
For instance, the name of this function and the docstring. I think it would make more sense to name them `call_additive_positional_bias` or even better `apply_additive_positional_bias`. Accordingly the dosctring probably needs a verb: `Tests if the positional bias is applied as expected during training (`training=True`) or evaluation`.

python/ml4ir/applications/ranking/tests/test_fixed_additive_positional_bias.py
```diff
@@ -0,0 +1,31 @@
...
+        biases = positional_bias(tf.convert_to_tensor(rank_index), training)
+        if not training:
+            assert tf.math.reduce_sum(biases).numpy() == 0.0
+        else:
+            for i in range(len(rank_index)):
+                assert biases[i][0].numpy() == positional_bias.weights[0][rank_index[i] - 1].numpy()[0]
+
+    def test_one_hot_conversion(self):
+        """calling additive positional bias"""
+        self.calling_additive_positional_bias([5, 4], 5, True)
```
balikasg, line 23: I am ramping-up on scala currently, and a pattern I see a lot is having an iterable and applying a function for each member of it. In scala it is elegant. In Python, we also have map etc. But here, I am wondering if it would have been more readable to write:
```python
for rank_index, max_ranks, does_training in zip([[1,3], [1,2], [2], [2,4,6,8]], [5,5,2,2,10], [True, False, True, False, True]): 
    self.calling_additive_positional_bias(rank_index, max_ranks, does_training)
```
It is a general comment this one, not a blocker.

python/ml4ir/base/features/feature_config.py
balikasg: this was an easy comment to fix, at least the docstring update.

python/ml4ir/base/model/architectures/fixed_additive_positional_bias.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class FixedAdditivePositionalBias(layers.Layer):
+    """
+    Implementing positional bias handling by learning a fixed additive weights
+    """
+    def __init__(self, max_ranks):
+        super(FixedAdditivePositionalBias, self).__init__()
+        self.dense = layers.Dense(1,
+                     name="fixed_additive_positional_bias_layer",
```
balikasg, line 12: should we use the name you defined in the `config`? `FIXED_ADDITIVE_POSITIONAL_BIAS_LAYER`?

python/ml4ir/base/model/architectures/fixed_additive_positional_bias.py
```diff
@@ -0,0 +1,37 @@
+import tensorflow as tf
+from tensorflow.keras import layers
+
+
+class FixedAdditivePositionalBias(layers.Layer):
+    """
```
balikasg, line 6: @mohazahran @lastmansleeping this class is the gist of this PR/contribution. And this trick maybe quite powerful in reality. But notice how cryptic/absent the description is although the code makes many assumptions! You should take 15' and actually describe what it does I think (if not for any other reason, so that you can brag about it!). I am exaggerating on purpose a bit..

```
Implements a solution to handle positional bias in logged data by adding a bias term to the relevance score of a document-query pair. We expect the bias term (learned during training) to be higher for higher ranks. Example: 
Assume 3 doc-query pairs as follows: 

... you give your example here and you show that something is added somewhere ....
  
The process of adding the positional bias is only applied during training because this is a way to de-bias data from logs (typically clicked/no clicked pairs). During inference we do not know the actual ranks, so the addition is turned off. 

This is an approach we have found useful for logged data as these are presented to the user with some order and there is a presentation bias (that this technique tries to model). When having actual relevance annotations you should probably not use this technique. 

Reference: a paper/blog if one exists
```
This is an example of how I would go with this.

mohazahran: Great suggestions @balikasg. I updated the PR and took them into account.

balikasg: thanks! you skipped the example.. 
Nice implementation btw!

balikasg (approved):
LGTM. this PR evolved a lot! Great work!  There is one thing missing: the way to correctly define a `model_config.yaml` is not shown anywhere. We should document it somewhere, it is not even on tests

mohazahran:
Thanks @lastmansleeping and @balikasg for the comments and feedback

PR #127, opened 2021-06-01
Update e2e test script to rebuild the docker image to avoid stale files

ducouloa:
This allows to keep in sync the test files for integration tests

### DIFF
--- a/python/docker-compose.yml
+++ b/python/docker-compose.yml
@@ -11,5 +11,6 @@ services:
       - ./logs:/home/ml4ir/python/logs
       - ./data:/home/ml4ir/python/data
       - ./models:/home/ml4ir/python/models
+      - ./ml4ir/:/home/ml4ir/python/ml4ir
     tty: true # uncomment for python debugging
     stdin_open: true # uncomment for python debugging
--- a/python/docker-compose.yml
+++ b/python/docker-compose.yml
@@ -11,6 +11,5 @@ services:
       - ./logs:/home/ml4ir/python/logs
       - ./data:/home/ml4ir/python/data
       - ./models:/home/ml4ir/python/models
-      - ./ml4ir/:/home/ml4ir/python/ml4ir
     tty: true # uncomment for python debugging
     stdin_open: true # uncomment for python debugging
--- a/python/ml4ir/build/Dockerfile
+++ b/python/ml4ir/build/Dockerfile
@@ -14,11 +14,10 @@ RUN pip config set global.extra-index-url https://pypi.python.org/simple
 RUN pip install --upgrade setuptools
 RUN pip install -r requirements.txt
 
-# Copy ml4ir
-COPY . /home/ml4ir/python/
+# Copy docker entrypoint
+COPY ml4ir/build/run_driver.sh $PYTHONPATH/.
 
 # Set docker entrypoint
-RUN cp $PYTHONPATH/ml4ir/build/run_driver.sh $PYTHONPATH/.
 RUN chmod +x run_driver.sh
 
 ENTRYPOINT ["./run_driver.sh"]
--- a/python/ml4ir/build/Dockerfile
+++ b/python/ml4ir/build/Dockerfile
@@ -14,10 +14,11 @@ RUN pip config set global.extra-index-url https://pypi.python.org/simple
 RUN pip install --upgrade setuptools
 RUN pip install -r requirements.txt
 
-# Copy docker entrypoint
-COPY ml4ir/build/run_driver.sh $PYTHONPATH/.
+# Copy ml4ir
+COPY . /home/ml4ir/python/
 
 # Set docker entrypoint
+RUN cp $PYTHONPATH/ml4ir/build/run_driver.sh $PYTHONPATH/.
 RUN chmod +x run_driver.sh
 
 ENTRYPOINT ["./run_driver.sh"]
--- a/jvm/ml4ir-inference/scripts/generate_model.sh
+++ b/jvm/ml4ir-inference/scripts/generate_model.sh
@@ -13,6 +13,8 @@ if [ -z "$2" ]
     export PYTHONPATH=.
     EXECUTOR="python"
 else
+  # docker-build to ensure to get all the local modifications tested
+  docker-compose build
   EXECUTOR="docker-compose run ml4ir python"
 fi
 


python/ml4ir/build/Dockerfile
```diff
@@ -14,11 +14,10 @@ RUN pip config set global.extra-index-url https://pypi.python.org/simple
 RUN pip install --upgrade setuptools
 RUN pip install -r requirements.txt
 
-# Copy ml4ir
-COPY . /home/ml4ir/python/
+# Copy docker entrypoint
+COPY ml4ir/build/run_driver.sh $PYTHONPATH/.
 
 # Set docker entrypoint
-RUN cp $PYTHONPATH/ml4ir/build/run_driver.sh $PYTHONPATH/.
```
balikasg, line 21: `RUN cp` looks strange here. `COPY`?

lastmansleeping: No, `cp` is right because it is an internal copy. `COPY` is to move files from your machine to the image.

lastmansleeping: (I think that's what was being done)

ducouloa: I just removed this guys 😀
I now only `COPY` (from project directory to Docker image) the `run_driver.sh` directly to where we want to find it.

The full ml4ir copy is not needed as I mount the volume now

lastmansleeping (changes requested):
@ducouloa I understand that this achieves what you are trying to do, but unfortunately don't think this is the right use of docker. Why can you not do the same by checking in the config file to the `data/` (or create another directory called `configs/`) and mount that instead? The issue with this is that the image in itself is not a shippable version of `ml4ir`. It might be fine for us, but think of users who are not developers necessarily.

ducouloa:
@lastmansleeping Ok, I see.
I didn't even know that we have a requirement to ship a Docker image with a working ml4ir app.
Is there a specification somewhere on what we want to deliver? Would it worse adding a test to ensure the integrity of the "ml4ir Docker file"?

lastmansleeping:
> @lastmansleeping Ok, I see.
> I didn't even know that we have a requirement to ship a Docker image with a working ml4ir app.
> Is there a specification somewhere on what we want to deliver? Would it worse adding a test to ensure the integrity of the "ml4ir Docker file"?

I don't think there is any written requirement, but it feels like standard practice to me. I'll let @jakemannix chime in as well to override my request.

jakemannix:
> > @lastmansleeping Ok, I see.
> > I didn't even know that we have a requirement to ship a Docker image with a working ml4ir app.

The point of a Docker image of an open source project, these days, is so users of the project can try it out / use it by just doing ```docker run -it python3 my_pipeline.py \
	--data_dir my/local/data_dir \
	--feature_config my_config_dir/feature_config.yaml \
	--run_id test \
	--data_format tfrecord \
	--execution_mode train_inference_evaluate  \
      ducouloa/ml4ir:0.1```

While we don't really support that yet (I don't think? Or does that just work?), that's the direction we want to go.

balikasg:
How are we advancing on this then? Do we keep it or put the config under a volume you already mount?

ducouloa:
Ok, I think I have a good trade-off.
I'm leaving alone the current Dockerfile recipe that allows what Jake is describing we want to achieve (the ml4ir command) _but_ I'm mounting the local `ml4ir` directory in place with the docker-compose rule file. I've check that this allows to take into account local modification in the Dockerfile (rule is : the last mount rule is taking over the others).

Let me know what do you think.

ducouloa:
@lastmansleeping @balikasg could you tell me what you think about this trade-off of mounting local `ml4ir` on the Docker image for test  purposes?

balikasg:
I don't have a strong preference/problem here, it goes below my knowledge of docker best practises. Usually I am rebuilding images. I am approving.

balikasg (approved):

ducouloa:
Yes, good call on rebuilding images. But this trick avoids any misses on the end_to_end testing that may results in incoherent test output.

python/docker-compose.yml
```diff
@@ -11,5 +11,6 @@ services:
       - ./logs:/home/ml4ir/python/logs
       - ./data:/home/ml4ir/python/data
       - ./models:/home/ml4ir/python/models
+      - ./ml4ir/:/home/ml4ir/python/ml4ir
```
lastmansleeping, line 14: @ducouloa So how does this work if `/home/ml4ir/python/ml4ir` already exists?

lastmansleeping: This is still dangerous IMO, because you might think there is something in the docker file, test it and ship it, and then realize it was only a local change/you forgot to rebuild the image/etc.

Why not just mount the configs? I don't understand the pushback.

ducouloa: I'm not sure I'm push-backing anything here?

My reasoning behind mounting the whole ml4ir dir is exactly to be able to test all the local change made, not only the config change, otherwise you need to `docker-compose build` each time you want to do an integration test with local change on the python code. And obviously, missing a build step while testing is a trap we want to avoid in the future right?

About shipping local change this is a great point until we kept our delivery process beeing manual on a local machine of a designed volunteer, which definitely sounds like a bad practice. I'd rather fix this delivery process IMO

lastmansleeping: > otherwise you need to docker-compose build each time you want to do an integration test with local change on the python code

This is the right way to do it IMO. The code needs to be frozen. You can alter the configs all you want. Mounting the code means there is a high change the docker image and the code which was just integration tested might not be the same. If this change only affects local, I guess it is okay. But think of docker being a way to ship `ml4ir` to end users. Mounting code just messes it up.

tl;dr -> rebuilding each time should be the right thing to do.

ducouloa: You seem strongly opinionated about this change, so I'll made the modification as you wish.

Note that the [official Docker documentation](https://docs.docker.com/storage/bind-mounts/#mount-into-a-non-empty-directory-on-the-container) reference this practice as "can be beneficial:

> If you bind-mount into a non-empty directory on the container, the directory’s existing contents are obscured by the bind mount. **This can be beneficial**, such as **when you want to test a new version of your application without building a new image**

ducouloa:
@lastmansleeping I changed the strategy to leave the current Docker files alone.

I'm now adding a building step to the  e2e script to ensure we include last modification when running those tests. I think this can serve the purpose well.

python/docker-compose.yml
```diff
@@ -12,4 +12,4 @@ services:
       - ./data:/home/ml4ir/python/data
       - ./models:/home/ml4ir/python/models
     tty: true # uncomment for python debugging
-    stdin_open: true # uncomment for python debugging
\ No newline at end of file
+    stdin_open: true # uncomment for python debugging
```
ducouloa, line 15: This is a end_of_line addition that I can't control

lastmansleeping (approved):
LGTM

PR #129, opened 2021-06-17
W-9172445 Adding pre_processing_step

Ullimague:
@W-9172445@
@rev jakemannix@
@rev balikasg@
@rev lastmansleeping@

### DIFF
--- a/python/ml4ir/base/pipeline.py
+++ b/python/ml4ir/base/pipeline.py
@@ -438,8 +438,9 @@ class RelevancePipeline(object):
 
     def pre_processing_step(self):
         """
-        Performs any pre processing step (such as copying data) by custom class inheriting this class, which would
-        override this method.
+        Performs arbitrary pre-processing steps such as copying or transforming data that the rest of the code can not
+        accommodate. It serves as a placeholder without an explicit implementation (returns self) in the base pipeline.
+        We expect that users can extend it in their custom pipelines.
         """
         return self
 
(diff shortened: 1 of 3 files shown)


lastmansleeping:
> @W-9172445@

@Ullimague Good to not have internal artifacts on ml4ir I think.

lastmansleeping (commented):
@Ullimague You might want to do the following while you are it as well:
- bump ml4ir version in setup.py
- Update changelog -> https://github.com/salesforce/ml4ir/blob/master/docs/source/misc/changelog.md

lastmansleeping (approved):
LGTM

python/ml4ir/base/pipeline.py
```diff
@@ -439,6 +436,13 @@ def run(self):
 
         return experiment_tracking_dict
 
+    def pre_processing_step(self):
+        """
+        Performs any pre processing step (such as copying data) by custom class inheriting this class, which would
```
balikasg, line 441: this is a somewhat cryptic docstring. I would go with something like: "Performs arbitrary pre-processing steps such as copying or transforming data that the rest of the code can not accommodate. It serves as a placeholder without an explicit implementation (returns `self`) in the `base` pipeline. We expect that users can extend it in their custom pipelines."

balikasg (approved):
LGTM. I would suggest to take 2' to improve a bit the docstring, but it is not a blocking comment on your implementation.

PR #132, opened 2021-06-24
Adding L1, L2 coefficients to positional bias weight initialization

mohazahran:
Example
`
positional_bias_handler:
  key: fixed_additive_positional_bias
  max_ranks: 25
  kernel_initializer: Zeros
  l1_coeff: 0.1
  l2_coeff: 0.1

`

### DIFF
--- a/python/ml4ir/base/model/architectures/dnn.py
+++ b/python/ml4ir/base/model/architectures/dnn.py
@@ -28,7 +28,10 @@ class DNN:
         self.layer_ops: List = self.define_architecture(model_config, feature_config)
         if 'positional_bias_handler' in self.model_config and self.model_config['positional_bias_handler'][
             'key'] == PositionalBiasHandler.FIXED_ADDITIVE_POSITIONAL_BIAS:
-            self.positional_bias_layer = FixedAdditivePositionalBias(max_ranks=self.model_config['positional_bias_handler']['max_ranks'])
+            self.positional_bias_layer = FixedAdditivePositionalBias(max_ranks=self.model_config['positional_bias_handler']['max_ranks'],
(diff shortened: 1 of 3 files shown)


python/ml4ir/base/model/architectures/dnn.py
```diff
@@ -28,7 +28,10 @@ def __init__(self, model_config: dict, feature_config: FeatureConfig, file_io):
         self.layer_ops: List = self.define_architecture(model_config, feature_config)
         if 'positional_bias_handler' in self.model_config and self.model_config['positional_bias_handler'][
             'key'] == PositionalBiasHandler.FIXED_ADDITIVE_POSITIONAL_BIAS:
-            self.positional_bias_layer = FixedAdditivePositionalBias(max_ranks=self.model_config['positional_bias_handler']['max_ranks'])
+            self.positional_bias_layer = FixedAdditivePositionalBias(max_ranks=self.model_config['positional_bias_handler']['max_ranks'],
```
lastmansleeping, line 31: @mohazahran Awesome! Can you update the tests as well?

mohazahran: The positional bias test focuses on the conversion to one-hot in training and testing. These changes here would not interfere with the current test.

lastmansleeping: I would recommend checking the config for that layer and also check if the initialization is all 0s.

mohazahran: ok will add, extra checks

lastmansleeping (changes requested):
Add another unittest, that way we have
- one for the case when initialization is zeros
- one when it is not

lastmansleeping (approved):
Nice! Thanks for adding the tests @mohazahran

Ullimague (approved):

PR #134, opened 2021-06-25
Adding pyspark to requirements.txt

lastmansleeping:
This is a cleaner version of removing pyspark from required dependencies. This way, if someone clones ml4ir and wants to set it up locally, they still install all required dependencies.

Previous PR -> https://github.com/salesforce/ml4ir/pull/133

### DIFF
--- a/python/optional_requirements.yaml
+++ b/python/optional_requirements.yaml
@@ -1,0 +1,26 @@
+# NOTE: Add dependencies here which the base ml4ir library can run without.
+# Packages mentioned here will be removed from the install_requires for ml4ir setup
+# and instead will be specified under the extras_require argument for an optional install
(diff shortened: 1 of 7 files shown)


python/optional_requirements.yaml
```diff
@@ -0,0 +1,26 @@
...
+#   - package==version
+#   - package==version
+#   ...
+#   - package==version
+#
+# To install a specific optional version of ml4ir, run
+# pip install ml4ir[optional_version]
+#
+# Example
+# -------
```
jakemannix, line 20: very slick.

jakemannix (approved):
looks great!

Ullimague (approved):

PR #136, opened 2021-07-01
Performing paired t-test for click rank distribution between new model and prod model on test set.

mohazahran:


python/ml4ir/applications/ranking/tests/test_t_test_calculation.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class TestTtestCalculation(unittest.TestCase):
+    """
+    Testing the calculation of paired t-test statistic and its p-value.
+    """
+
+    def t_test_calculation(self, a, b, batchzs, pvalue_threshold):
+        n = len(a)
+        expected_t_test_stat, expected_pvalue = stats.ttest_rel(a, b)
```
balikasg, line 17: this is a two-sided test. You are using a related sets test. Why are you not using a standard t-test? From the doc of scipy I see that it is like comparing the same set of students in two exams. The way I read it, a student is a ranker system and here you are comparing them on the same data. I admit, I do not expect that the p-val to be drastically different, I am just curious.

python/ml4ir/applications/ranking/tests/test_t_test_calculation.py
```diff
@@ -0,0 +1,76 @@
...
+        expected_decision = expected_pvalue < pvalue_threshold
+        agg_mean, agg_count, agg_M2 = 0, 0, 0
+        for bcount in range(int(n/batchzs)):
+            batch_a = a[bcount*batchzs : (bcount+1)*(batchzs)]
+            batch_b = b[bcount*batchzs : (bcount+1)*(batchzs)]
+            d = batch_a - batch_b
+            agg_count, agg_mean, agg_M2 = compute_stats_from_stream(d, agg_count, agg_mean, agg_M2)
```
balikasg, line 24: I am a bit confused here, I would need more time to go deep in your calculations, also you are doing some calculations on the `perform_click_rank_dist_paired_t_test` arguments. But in your tests, please do some sanity checks: 
- compare at least the variance analytically calculated with the one you calculate in the streamed approach
- anything else you are doing in stream vs can be done analytically
- Compare the p-values, not the decisions. I think you are using a two sided as expected and you are calculating an 1-sided in your  `perform_click_rank_dist_paired_t_test`, I need to be convinced ;-)

python/ml4ir/applications/ranking/tests/test_t_test_calculation.py
```diff
@@ -0,0 +1,76 @@
...
+        t_test_stat, pvalue = perform_click_rank_dist_paired_t_test(agg_mean, np.sqrt(agg_M2/agg_count), agg_count)
+        decision = pvalue < pvalue_threshold
+        assert decision == expected_decision
+        #assert np.isclose(expected_pvalue, pvalue, atol=0.001)
+
+    def test_1(self):
+        pvalue_threshold = 0.1
+        n = 100
+        batchzs = 10
+        a = np.random.randn(n) + 1
```
balikasg, line 35: `np.random.randn` returns from a standard normal distribution. We want this functionality to compare distributions of ranks, which are much different than the the values you get from a normal distribution (notably they are probability distributions themselves). We should adapt the data we use to something more realistic for our scenario.

balikasg (changes requested):
I left several comments. This is nice stuff, but this PR will need more work.

python/ml4ir/base/model/relevance_model.py
```diff
@@ -504,6 +529,12 @@ def predict(
                 else:
                     # If writing first time, write headers to CSV file
                     predictions_df.to_csv(outfile, mode="w", header=True, index=False)
+
+                # Accumulating statistics for t-test calculation
+                clicked_records = predictions_df[predictions_df['clicked'] == 1.0]
```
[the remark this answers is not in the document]
mohazahran: good suggestion. I think I would move this to the ranking_model

python/ml4ir/base/model/relevance_model.py
```diff
@@ -32,6 +33,28 @@ class RelevanceModelConstants:
...
+def perform_click_rank_dist_paired_t_test(mean, std, n):
+    """
+    Compute the paired t-test statistic and its p-value given mean, standard deviation and sample count
+    """
+    t_test_stat = mean / (std/np.sqrt(n))
+    df = n - 1
+    pvalue = 1 - stats.t.cdf(np.abs(t_test_stat), df=df)
+    return t_test_stat, pvalue
+
+def compute_stats_from_stream(diff, count, mean, M2):
```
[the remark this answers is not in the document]
mohazahran: Will account for these

python/ml4ir/applications/ranking/tests/test_t_test_calculation.py
mohazahran: The way I see it is the two models (or students) are evaluated on the same dataset with the same splits and so a paired t-test is a good fit.

python/ml4ir/applications/ranking/tests/test_t_test_calculation.py
mohazahran: That's true. I'm also adding simulations with exp distribution which is similar as well to the rank distribution

python/ml4ir/applications/ranking/tests/test_t_test_calculation.py
mohazahran: Good point! I believe we should go with a two sided to be more inclusive.

python/build-requirements.txt
```diff
@@ -12,6 +12,7 @@ mypy
 tensorflow==2.0.4
 numpy==1.18.5
 pandas==1.2.1
+scipy==1.3.2
```
lastmansleeping, line 15: @mohazahran I don't see this version as an approved scipy version internally. Any reason you need this specifically? Will slack you the link to internal PyPi.

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -19,6 +21,66 @@
...
+    M2: float
+        The updated squared distance from the mean
+    """
+    for i in range(len(diff)):
+        count += 1
+        delta = diff[i] - mean
+        mean += delta / count
+        delta2 = diff[i] - mean
+        M2 += delta * delta2
+    return count, mean, M2
```
lastmansleeping, line 83: @mohazahran Move these out to a separate module.

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -19,6 +21,66 @@
...
+    t_test_stat = mean / (np.sqrt(variance/n))
+    df = n - 1
+    pvalue = 1 - stats.t.cdf(np.abs(t_test_stat), df=df)
+    return t_test_stat*2, pvalue*2  # multiplying by 2 for two sided t-test
+
+
+def compute_stats_from_stream(diff, count, mean, M2):
+    """
+    Compute the running mean, variance for a stream of data.
+    src: Welford's online algorithm: https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance
```
lastmansleeping, line 55: Thanks!

python/ml4ir/applications/ranking/tests/test_t_test_calculation.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class TestTtestCalculation(unittest.TestCase):
+    """
+    Testing the calculation of paired t-test statistic and its p-value.
+    """
+
+    def t_test_calculation(self, a_bucket, b_bucket, batch_size):
+        n = len(a_bucket)
+        expected_t_test_stat, expected_pvalue = stats.ttest_rel(a_bucket, b_bucket)
```
lastmansleeping, line 17: @mohazahran Shouldn't you be doing `ttest_ind` here?

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -56,15 +118,72 @@ def predict(
...
+
+            batch_count += 1
+            if batch_count % logging_frequency == 0:
+                self.logger.info("Finished predicting scores for {} batches".format(batch_count))
+
+        predictions_df = None
+        if logs_dir:
+            self.logger.info("Model predictions written to -> {}".format(outfile))
+        else:
+            predictions_df = pd.concat(predictions_df_list)
```
lastmansleeping, line 174: @mohazahran Why is this whole thing being duplicated here? Is this the only new line being added?
```agg_count, agg_mean, agg_M2 = compute_stats_from_stream(diff, agg_count, agg_mean, agg_M2)```

File: python/ml4ir/applications/ranking/model/ranking_model.py [hunk 11]
```diff
@@ -56,15 +118,72 @@ def predict(
...
+
+        # performing click rank distribution t-test
```
lastmansleeping, line 176: @mohazahran You should add a new argument to `--run-t-test` and not run this always.

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -56,15 +118,72 @@ def predict(
...
+        if agg_count >= 2:
+            self.logger.info("Performing a paired t-test between the click rank distribution of new model and the old model:\n\tNull hypothesis: There is no difference between the two click distributions.\n\tAlternative hypothesis: There is a difference between the two click distributions")
+            t_test_stat, pvalue = perform_click_rank_dist_paired_t_test(agg_mean, np.sqrt(agg_M2/(agg_count-1)), agg_count)
+            self.logger.info("t-test statistic={}, p-value={}".format(t_test_stat, pvalue))
+            if pvalue < RankingConstants.rank_distribution_t_test_pvalue_threshold:
+                self.logger.info("With p-value threshold={} > p-value --> we reject the null hypothesis. The click rank distribution of the new model is significantly different from the old model".format(RankingConstants.rank_distribution_t_test_pvalue_threshold))
+            else:
+                self.logger.info("With p-value threshold={} < p-value --> we cannot reject the null hypothesis. The click rank distribution of the new model is not significantly different from the old model".format(RankingConstants.rank_distribution_t_test_pvalue_threshold))
```
lastmansleeping, line 184: @mohazahran Worth moving to a separate method for future readability.

python/ml4ir/applications/ranking/model/ranking_model.py
lastmansleeping: I think a better way to do this would be to pass all the aggregate stats you need for the t-test as feature functions into `additional_features`. There should be some stateful variables you can use to update existing values IMO. Let me know if you want to sync up about this. It's been a while for me as well.

python/ml4ir/applications/ranking/tests/test_t_test_calculation.py
```diff
@@ -0,0 +1,119 @@
...
+        for bcount in range(int(n/batch_size)):
+            batch_a = a_bucket[bcount*batch_size : (bcount+1)*(batch_size)]
+            batch_b = b_bucket[bcount*batch_size : (bcount+1)*(batch_size)]
+            d = batch_a - batch_b
+            agg_count, agg_mean, agg_M2 = compute_stats_from_stream(d, agg_count, agg_mean, agg_M2)
+
+        t_test_stat, pvalue = perform_click_rank_dist_paired_t_test(agg_mean, (agg_M2/(agg_count-1)), agg_count)
+        assert np.isclose(expected_pvalue, pvalue, atol=0.0001)
+
+    def test_1(self):
```
lastmansleeping, line 28: @mohazahran This is great. I would also add a unit test for the two methods individually.

python/ml4ir/applications/ranking/model/ranking_model.py
lastmansleeping: I found an easier alternative for you - move all of this to `evaluate()` which already does this and is a more logical place for doing a t-test anyway.

lastmansleeping (commented):
Some great feedback from Georgios already. I left some comments on better structuring this for the future. Let me know if you have questions about any of the requested changes.

python/build-requirements.txt
mohazahran: Good catch. I didn't check it against that list. I will update it.

python/ml4ir/applications/ranking/tests/test_t_test_calculation.py
mohazahran: Since the two models are tested on the same splits, then a paired t-test would be suitable and hence ttest_rel. What do you think?

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -19,6 +21,7 @@
 
 class RankingConstants:
     NEW_RANK = "new_rank"
+    rank_distribution_t_test_pvalue_threshold = 0.1
```
lastmansleeping, line 24:
```suggestion
    T_TEST_P_VAL_THRESHOLD = 0.1
```

python/ml4ir/applications/ranking/tests/test_t_test_calculation.py
lastmansleeping: ping

lastmansleeping (changes requested):
- Add t-stat and p-val to the `metrics_dict` in `evaluate()`. This will get saved in the `_SUCCESS` file and can be used later.
- Add test for `compute_stats_from_stream` and `perform_click_rank_dist_paired_t_test` separately
- Define a CLI argument for t-test. Not a constant.

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -1,6 +1,7 @@
 import tensorflow as tf
 from tensorflow import data
 import os
+import sys
```
balikasg, line 4: minor: but the order of the imports here should follow core python packages, pip-installed and ml4ir.
So, `os`, `sys` should go to the top.

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -74,6 +74,7 @@ def evaluate(
         logs_dir: Optional[str] = None,
         logging_frequency: int = 25,
         compute_intermediate_stats: bool = True,
+        rank_distribution_t_test_pvalue_threshold: float = 0.1,
```
balikasg, line 77: Please add it to the docstring

balikasg: Since you touch the docstring: 
"Evaluate the RelevanceModel" can become something more explicit!

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -173,6 +182,16 @@ def evaluate(
...
 
+        # performing click rank distribution t-test
+        if rank_distribution_t_test_pvalue_threshold > 0:
+            t_test_stat, pvalue = perform_click_rank_dist_paired_t_test(agg_mean, (agg_M2 / (agg_count - 1)), agg_count)
+            t_test_log_results(t_test_stat, pvalue, rank_distribution_t_test_pvalue_threshold, self.logger)
+
+            # Log t-test statistic and p-value
+            metrics_dict['Rank distribution t-test statistic'] = t_test_stat
+            metrics_dict['Rank distribution t-test pvalue'] = pvalue
+            metrics_dict['Rank distribution t-test statistical significant'] = pvalue < rank_distribution_t_test_pvalue_threshold
```
balikasg, line 193:
```suggestion
            metrics_dict['Rank distribution t-test: change is statistically significant'] = pvalue < rank_distribution_t_test_pvalue_threshold
```

balikasg: not a blocker

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -149,9 +150,17 @@ def evaluate(
 
         batch_count = 0
         df_grouped_stats = pd.DataFrame()
+        # defining variables to compute running mean and variance for t-test computations
+        agg_count, agg_mean, agg_M2 = 0, 0, 0
```
balikasg, line 154: you may want to say something about `agg_M2`

python/ml4ir/applications/ranking/t_test.py
```diff
@@ -0,0 +1,90 @@
...
+    pvalue: float
+        The p-value of the t-test statistic
+    """
+    t_test_stat = mean / (np.sqrt(variance/n))
+    df = n - 1
+    pvalue = 1 - stats.t.cdf(np.abs(t_test_stat), df=df)
+    return t_test_stat*2, pvalue*2  # multiplying by 2 for two sided t-test
+
+
+def compute_stats_from_stream(diff, count, mean, M2):
```
balikasg, line 30: having a capital `M2` feels unnatural here

python/ml4ir/applications/ranking/t_test.py
```diff
@@ -0,0 +1,90 @@
...
+    for i in range(len(diff)):
+        count += 1
+        delta = diff[i] - mean
+        mean += delta / count
+        delta2 = diff[i] - mean
+        M2 += delta * delta2
+    return count, mean, M2
+
+
+def t_test_log_results(t_test_stat, pvalue, rank_distribution_t_test_pvalue_threshold, logger):
```
balikasg, line 64: I understand that this was an ask from @lastmansleeping, still having a separate logging function feels strange. Not a blocker.

python/ml4ir/base/config/parse_args.py
```diff
@@ -353,6 +353,13 @@ def define_args(self):
                  "(All info after the '#' in the format [key = val]).",
         )
 
+        self.add_argument(
```
balikasg, line 356: I want to argue we could completely remove this complexity. I mean, always calculate the test (your implementation should scale and I do not expect it to make things slow). Doing so: 
- removes the need for this argument
- removes the addition of the arguments in the base relevance model and base relevance pipeline (which is good, because they are not that generic)
- Reports the p-value in the dictionaries. The user, when post-processing has all the info needed to decide on stat. significance and add the boolean column `is_stat_sig`. 

What do you think?

balikasg (commented):
Great progress!

python/ml4ir/base/config/parse_args.py
lastmansleeping: @mohazahran Why don't you add this in https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/applications/ranking/config/parse_args.py ?

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -149,9 +150,17 @@ def evaluate(
...
         for predictions_dict in test_dataset.map(_predict_fn).take(-1):
             predictions_df = pd.DataFrame(predictions_dict)
 
+            # Accumulating statistics for t-test calculation
+            clicked_records = predictions_df[predictions_df[self.feature_config.get_label("node_name")] == 1.0]
+            diff = (clicked_records[RankingConstants.NEW_RANK] - clicked_records[
```
lastmansleeping, line 160: @mohazahran Just noticed - We should probably compute the t-test with 1 / rank instead of rank because we are interested in the mean of the reciprocal ranks.

python/ml4ir/applications/ranking/model/ranking_model.py
lastmansleeping: ping

python/ml4ir/applications/ranking/t_test.py
lastmansleeping: I just didn't want to overcrowd the `evaluate()` function. That's all.

lastmansleeping: I agree there is almost no functionality here. :|

python/ml4ir/applications/ranking/t_test.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+    logger.info(
+        "Performing a paired t-test between the click rank distribution of new model and the old model:\n\tNull hypothesis: There is no difference between the two click distributions.\n\tAlternative hypothesis: There is a difference between the two click distributions")
+    logger.info("t-test statistic={}, p-value={}".format(t_test_stat, pvalue))
+    if pvalue < rank_distribution_t_test_pvalue_threshold:
+       logger.info(
+            "With p-value threshold={} > p-value --> we reject the null hypothesis. The click rank distribution of the new model is significantly different from the old model".format(
+                rank_distribution_t_test_pvalue_threshold))
+    else:
+       logger.info(
```
lastmansleeping, line 88:
```suggestion
       logger.warning(
```

python/ml4ir/applications/ranking/t_test.py
lastmansleeping: +1

python/ml4ir/applications/ranking/t_test.py
lastmansleeping: @mohazahran Sorry, can we move it back? I assumed you would be able to move the entire t-test (all calls+logging) to a separate helper method. But I guess, that's not simple.

python/ml4ir/base/pipeline.py
```diff
@@ -329,6 +329,7 @@ def run(self):
                     group_metrics_min_queries=self.args.group_metrics_min_queries,
                     logs_dir=self.logs_dir_local,
                     compute_intermediate_stats=self.args.compute_intermediate_stats,
+                    rank_distribution_t_test_pvalue_threshold=self.args.rank_distribution_t_test_pvalue_threshold,
```
lastmansleeping, line 332: @mohazahran This is why your test is failing.

python/ml4ir/applications/ranking/t_test.py
lastmansleeping: I think the better way to do this is define a `run_ttest()` in `RelevanceModel` and then implement it in `RankingModel` with the code here and in the current ttest block in `RankingModel`.

Populate and return a ttest metrics dictionary from that above function.

In evaluate(), simply call that method and do a `metrics_dict.update(whatever dictionary was returned)`

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -65,6 +65,39 @@ def predict(
...
+            P-value threshold for student t-test
+        metrics_dict: dict
+            dictonary of metrics to keep track
+
+        Returns
+        -------
+        t_test_stat: float
+            The t-test statistic
+        pvalue: float
+            The p-value of the t-test statistic
```
lastmansleeping, line 90: @mohazahran I don't think you return anything here, right?

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -173,6 +217,10 @@ def evaluate(
                 self.logger.info(
                     "Finished evaluating {} batches".format(batch_count))
 
+        # performing click rank distribution t-test
+        if ttest_pvalue_threshold > 0:
+            self.run_ttest(agg_mean, (agg_M2 / (agg_count - 1)), agg_count, ttest_pvalue_threshold, metrics_dict)
```
lastmansleeping, line 222: @mohazahran This is not a good coding practice - passing a dictionary into a function which modifies it in place. A cleaner way to do this is to return `t_test_metrics_dict` from `run_ttest` and then do a `metrics_dict = metrics_dict.update(...)` here. That way someone reading this function can already follow all the values that metrics_dict contains. Thoughts?

python/ml4ir/applications/ranking/t_test.py
lastmansleeping: ping

python/ml4ir/applications/ranking/t_test.py
```diff
@@ -0,0 +1,90 @@
...
+    -------
+    t_test_stat: float
+        The t-test statistic
+    pvalue: float
+        The p-value of the t-test statistic
+    """
+    t_test_stat = mean / (np.sqrt(variance/n))
+    df = n - 1
+    pvalue = 1 - stats.t.cdf(np.abs(t_test_stat), df=df)
+    return t_test_stat*2, pvalue*2  # multiplying by 2 for two sided t-test
```
lastmansleeping, line 27: @mohazahran Should you multiply the t-stat by 2 as well? I might be wrong.

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -149,9 +185,17 @@ def evaluate(
...
         batch_count = 0
         df_grouped_stats = pd.DataFrame()
+        # defining variables to compute running mean and variance for t-test computations
+        agg_count, agg_mean, agg_M2 = 0, 0, 0
         for predictions_dict in test_dataset.map(_predict_fn).take(-1):
             predictions_df = pd.DataFrame(predictions_dict)
 
+            # Accumulating statistics for t-test calculation
+            clicked_records = predictions_df[predictions_df[self.feature_config.get_label("node_name")] == 1.0]
+            diff = (1/clicked_records[RankingConstants.NEW_RANK] - 1/clicked_records[
```
lastmansleeping, line 195: @mohazahran Can you add a comment about using inverse click ranks here?

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -149,9 +181,17 @@ def evaluate(
...
+        # defining variables to compute running mean and variance for t-test computations
+        agg_count, agg_mean, agg_M2 = 0, 0, 0
         for predictions_dict in test_dataset.map(_predict_fn).take(-1):
             predictions_df = pd.DataFrame(predictions_dict)
 
+            # Accumulating statistics for t-test calculation using 1/rank
+            clicked_records = predictions_df[predictions_df[self.feature_config.get_label("node_name")] == 1.0]
+            diff = (1/clicked_records[RankingConstants.NEW_RANK] - 1/clicked_records[
+                self.feature_config.get_rank("node_name")]).to_list()
+            agg_count, agg_mean, agg_M2 = compute_stats_from_stream(diff, agg_count, agg_mean, agg_M2)
```
lastmansleeping, line 193: @mohazahran You should probably put this within an if block as well -> `if ttest_pvalue_threshold > 0:`

lastmansleeping (approved):
Thanks for making all the changes @mohazahran :)

python/ml4ir/applications/classification/model/classification_model.py
```diff
@@ -25,6 +25,7 @@ def evaluate(
         logs_dir: Optional[str] = None,
         logging_frequency: int = 25,
         compute_intermediate_stats: bool = True,
+        ttest_pvalue_threshold: float = 0.1,
```
balikasg, line 28: @mohazahran @lastmansleeping do you see the problem with this?

python/ml4ir/applications/ranking/t_test.py
```diff
@@ -0,0 +1,90 @@
...
+    for i in range(len(diff)):
+        count += 1
+        delta = diff[i] - mean
+        mean += delta / count
+        delta2 = diff[i] - mean
+        m2 += delta * delta2
+    return count, mean, m2
+
+
+def t_test_log_results(t_test_stat, pvalue, ttest_pvalue_threshold, logger):
```
balikasg, line 64: It is a bit strange design decision we made the we need to lass a logger to the function. I will check this in more detail, but ideally a logger should be defined for the whole project once and not being passed with every function.

python/ml4ir/applications/ranking/tests/test_t_test_calculation.py
```diff
@@ -0,0 +1,121 @@
...
+        n = 100
+        batch_size = 10
+        batch_a = np.random.randn(n) + 1
+        batch_b = np.random.randn(n)
+        self.t_test_calculation(batch_a, batch_b, batch_size)
+
+    def test_2(self):
```
balikasg, line 37: Probably minor: there is a lot of DRY in `test_1` ... `test_9`

python/ml4ir/base/config/parse_args.py
balikasg: this comment still remains. I appreciate the hard work on this PR, but I am not convinced of the argument's need. See my first comment above. Also, what @lastmansleeping suggests, seems to be a good compromise (if we actually need to compromise).

mohazahran: @balikasg  Yeah I tried Ashish's suggestion. Classification testcase fails because the evaluate function still needs `ttest_pvalue_threshold` to be passed to it. So calling it with classificationModel would crash

python/ml4ir/applications/ranking/t_test.py
lastmansleeping: Yeah, I agree. I can file a WI to track this if you'd like.

balikasg (approved):
Approving to unblock development after the related discussion at: https://platformcloud.slack.com/archives/C01JH622PMF/p1626964404022000

PR #137, opened 2021-07-14
[WIP] Adding Kfold cross validation run

mohazahran:
Adding kfold cross validation run and aggregate the results.

### DIFF
--- a/python/ml4ir/applications/ranking/pipeline.py
+++ b/python/ml4ir/applications/ranking/pipeline.py
@@ -1,5 +1,9 @@
 import sys
 import ast
+import os
+import pathlib
+import pandas as pd
+from scipy import stats
 from argparse import Namespace
 from tensorflow.keras.metrics import Metric
 from tensorflow.keras.optimizers import Optimizer
@@ -26,7 +30,7 @@ class RankingPipeline(RelevancePipeline):
     """Base class that defines a pipeline to train, evaluate and save
     a RankingModel using ml4ir"""
 
-    def __init__(self, args: Namespace):
+    def __init__(self, args: Namespace,):
         """
         Constructor to create a RelevancePipeline object to train, evaluate
         and save a model on ml4ir.
@@ -150,13 +154,81 @@ class RankingPipeline(RelevancePipeline):
             )
 
 
+def kfold_analysis(original_logs_dir, run_id, num_folds, logger):
+    """
+    Aggregate results of the k-fold runs and perform t-test on the results.
+    """
+    rows = []
+    pvalue_threshold = 0.1
+    metrics = ['train_old_MRR', 'train_new_MRR', 'val_old_MRR', 'val_new_MRR', 'test_old_MRR', 'test_new_MRR']
+    for i in range(num_folds):
+        row = {"fold":i}
+        logs_dir = pathlib.Path(original_logs_dir) / run_id / "fold_{}".format(i) / run_id
+        with open(logs_dir/"_SUCCESS", 'r') as f:
+            for line in f:
+                parts = line.split(',')
+                if parts[0] in metrics:
+                    row[parts[0]] = float(parts[1])
+        rows.append(row)
+    results = pd.DataFrame(rows)
+    logger.info(results)
+    # calculate statistical paired t-test
+    for dataset in ['train', 'val', 'test']:
+        t_test_stat, pvalue = stats.ttest_rel(results["{}_old_MRR".format(dataset)], results["{}_new_MRR".format(dataset)])
+        logger.info("{}_t_test_stat={}".format(dataset, t_test_stat))
+        logger.info("{}_pvalue={} --> (statistically significant={})".format(dataset, pvalue, pvalue<pvalue_threshold))
+
+
+def kfold_cross_validation_run(args):
+    """
+    Performs K-fold cross validation
+    """
+    rp = RankingPipeline(args=args)
+    rp.logger.info("K-fold Cross Validation mode starting with k={}".format(args.kfold))
+    rp.logger.info("Reading datasets ...")
+    relevance_dataset = rp.get_relevance_dataset()
+    rp.logger.info("Relevance Dataset created")
+    all_data = relevance_dataset.train.concatenate(relevance_dataset.validation)
+    all_data = all_data.unbatch()
+    folds = args.kfold
+    original_logs_dir = str(rp.args.logs_dir)
+    original_models_dir = str(rp.args.models_dir)
+    original_run_id = rp.run_id
+    rp.logger.info("Starting K-fold Cross Validation with k={}".format(args.kfold))
+    for i in range(folds):  # indexes (folds)
+        rp.logger.info("fold={}".format(i))
+        logs_dir = pathlib.Path(original_logs_dir) / rp.args.run_id / "fold_{}".format(i)
+        models_dir = pathlib.Path(original_models_dir) / rp.args.run_id / "fold_{}".format(i)
+        args.logs_dir = pathlib.Path(logs_dir).as_posix()
+        args.models_dir = pathlib.Path(models_dir).as_posix()
+        validation = all_data.shard(folds, i)
+        training_idx = list(range(folds))
+        training_idx.remove(i)
+        training = None
+        for j in training_idx:
+            if not training:
+                training = all_data.shard(folds, j)
+            else:
+                training = training.concatenate(all_data.shard(folds, j))
+
+        relevance_dataset.validation = validation.batch(args.batch_size, drop_remainder=False)
+        relevance_dataset.train = training.batch(args.batch_size, drop_remainder=False)
+
+        pipeline = RankingPipeline(args=args)
+        pipeline.run(relevance_dataset=relevance_dataset)
+    rp.local_io.rm_dir(os.path.join(rp.data_dir_local, "tfrecord"))
+    kfold_analysis(original_logs_dir, original_run_id, folds, rp.logger)
+
+
 def main(argv):
     # Define args
     args: Namespace = get_args(argv)
-
-    # Initialize Relevance Pipeline and run in train/inference mode
-    rp = RankingPipeline(args=args)
-    rp.run()
+    if args.kfold > 1:
+        kfold_cross_validation_run(args=args)
+    else:
+        # Initialize Relevance Pipeline and run in train/inference mode
+        rp = RankingPipeline(args=args)
+        rp.run()
 
 
 if __name__ == "__main__":
--- a/python/ml4ir/base/config/parse_args.py
+++ b/python/ml4ir/base/config/parse_args.py
@@ -353,6 +353,13 @@ class RelevanceArgParser(ArgumentParser):
                  "(All info after the '#' in the format [key = val]).",
         )
 
+        self.add_argument(
+            "--kfold",
+            type=int,
+            default=1,
+            help="The number of folds for K-fold Cross Validation.",
+        )
+
     def set_default_args(self):
         pass
 
--- a/python/ml4ir/base/pipeline.py
+++ b/python/ml4ir/base/pipeline.py
@@ -265,7 +265,7 @@ class RelevancePipeline(object):
         """
         raise NotImplementedError
 
-    def run(self):
+    def run(self, relevance_dataset=None):
         """
         Run the pipeline to train, evaluate and save the model
 
@@ -288,8 +288,9 @@ class RelevancePipeline(object):
             test_metrics = dict()
 
             # Build dataset
-            relevance_dataset = self.get_relevance_dataset()
-            self.logger.info("Relevance Dataset created")
+            if not relevance_dataset:
+                relevance_dataset = self.get_relevance_dataset()
+                self.logger.info("Relevance Dataset created")
 
             # Build model
             relevance_model = self.get_relevance_model()
@@ -311,6 +312,7 @@ class RelevancePipeline(object):
                     monitor_metric=self.args.monitor_metric,
                     monitor_mode=self.args.monitor_mode,
                     patience=self.args.early_stopping_patience,
+                    #kfold=self.args.kfold,
                 )
 
             if self.args.execution_mode in {
@@ -469,7 +471,7 @@ class RelevancePipeline(object):
             f.write(job_info)
 
         # Delete temp data directories
-        if self.data_format == DataFormatKey.CSV:
+        if self.data_format == DataFormatKey.CSV and self.args.kfold <= 1:
             self.local_io.rm_dir(os.path.join(self.data_dir_local, "tfrecord"))
         self.local_io.rm_dir(DefaultDirectoryKey.TEMP_DATA)
         self.local_io.rm_dir(DefaultDirectoryKey.TEMP_MODELS)
--- a/python/ml4ir/base/model/relevance_model.py
+++ b/python/ml4ir/base/model/relevance_model.py
@@ -404,6 +404,7 @@ class RelevanceModel:
         )
 
         if self.is_compiled:
+
             history = self.model.fit(
                 x=dataset.train,
                 validation_data=dataset.validation,


python/build-requirements.txt
```diff
@@ -12,6 +12,7 @@ mypy
 tensorflow==2.0.4
 numpy==1.18.5
 pandas==1.2.1
+scipy==1.5.4
```
lastmansleeping, line 15: @mohazahran 1.6.0 seems to be internally approved for py37. FYI

lastmansleeping: You can update `requirements.txt` as well.

python/ml4ir/base/config/parse_args.py
```diff
@@ -353,6 +353,20 @@ def define_args(self):
                  "(All info after the '#' in the format [key = val]).",
         )
 
+        self.add_argument(
+            "--kfold",
```
lastmansleeping, line 357: @mohazahran Maybe k_fold_cv?

lastmansleeping: Or num_k_fold_cv (this might be redundant, I agree)

python/ml4ir/applications/ranking/pipeline.py
```diff
@@ -1,5 +1,9 @@
 import sys
 import ast
+import os
+import pathlib
```
lastmansleeping, line 4: [Minor] @mohazahran `from pathlib import Path`

lastmansleeping (changes requested):
This is awesome. Very neat tricks. Few things I recommend top of my head to help future dev
- KFoldCV should be usable for all ml4ir jobs. Nothing stopping us from doing that, right?
- Do not call get_relevance_model() repeatedly. There is a tf-keras bug which does not release memory for garbage collection. So all these models stay in memory and every gradient step increases memory. Just store the best_model and current_model at a time (I guess). More importantly, try to find a way to reset the weights back to original state (maybe even make a deepcopy if possible).
- Log all the results at the end that can be used by downstream jobs like HPO

Good time to do some small refactoring - I would recommend defining kfold as part of RelevanceModel.run() itself. So how? Thinking out loud...
Break up the `run()` method so that you can separate dataset and model creation and the remaining steps (you can call the latter `run_internal()` or something). So now in `run()`, you can implement the logic to do kfold CV in RelevancePipeline itself. We can discuss more about how to do the analysis and everything. But you probably also need a few more arguments like `optimize_mode` (maximize or minimize) and `optimize_metric` (the actual metrics from your handy `metrics_dict`).


Will ping a reference idea on slack from something internal.

balikasg (commented):
few comments, I noticed it is WIP

python/ml4ir/applications/ranking/pipeline.py
```diff
@@ -150,13 +154,95 @@ def validate_args(self):
...
+    rp.logger.info("K-fold Cross Validation mode starting with k={}".format(args.kfold))
+    rp.logger.info("Reading datasets ...")
+    relevance_dataset = rp.get_relevance_dataset()
+    rp.logger.info("Relevance Dataset created")
+    if args.include_testset_in_kfold:
+        all_data = relevance_dataset.train.concatenate(relevance_dataset.validation).concatenate(relevance_dataset.test)
+    else:
+        all_data = relevance_dataset.train.concatenate(relevance_dataset.validation)
+    all_data = all_data.unbatch()
+    all_data = all_data.shuffle(args.batch_size * 2)
```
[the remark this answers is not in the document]
mohazahran: That's true, it's a memory friendly way to shuffle. However, this has to be removed as there is a bug when shuffle() is called before shard().

python/build-requirements.txt
lastmansleeping: ping

python/ml4ir/base/pipeline.py
```diff
@@ -265,9 +269,209 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+            all_data = relevance_dataset.train.concatenate(relevance_dataset.validation).concatenate(
+                relevance_dataset.test)
+        else:
+            all_data = relevance_dataset.train.concatenate(relevance_dataset.validation)
+
+        # un-batch and shuffle all queries
+        all_data = all_data.unbatch()
+        # shuffling before using the shard method gives unexpected results. Should be avoided
+        # all_data = all_data.shuffle(batch_size * 2)
+        return all_data
```
lastmansleeping, line 350: @mohazahran I think you should move these two methods to `RelevanceDataset`. Maybe even create a `KFoldDataset` class and have methods to fetch, perhaps?

python/ml4ir/base/pipeline.py
```diff
@@ -265,9 +269,209 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+            # We apply prefetch as it improved train/test/validation throughput by 30% in some real model training.
+            relevance_dataset.training = training.prefetch(tf.data.experimental.AUTOTUNE)
+            relevance_dataset.validation = validation.prefetch(tf.data.experimental.AUTOTUNE)
+            if args.include_testset_in_kfold:
+                test = test.batch(args.batch_size, drop_remainder=False)
+                relevance_dataset.test = test.prefetch(tf.data.experimental.AUTOTUNE)
+
+            # run the pipeline over the created folds.
+            pipeline = self.create_pipeline_for_kfold(args)
+            pipeline.run_pipeline(relevance_dataset=relevance_dataset)
```
lastmansleeping, line 459: @mohazahran I don't think you've still addressed the issue of creating multiple models. It will definitely blow up memory I have tried this.

python/ml4ir/base/pipeline.py
```diff
@@ -265,9 +269,209 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+
+        # removing intermediate directory and run kfold analysis
+        self.local_io.rm_dir(os.path.join(self.data_dir_local, "tfrecord"))
+        self.run_kfold_analysis(base_logs_dir, base_run_id, num_folds)
+
+
+
+    def run_pipeline(self, relevance_dataset=None):
```
lastmansleeping, line 467: @mohazahran You shouldn't rename this method. I'm not sure what impact on other jobs depending on this is. Have you tested that it doesn't break anything?

lastmansleeping: If you implement a `KFoldDataset` class, you can even have something like a `fetch()` method that does this for you without having to pass the dataset as an argument.

python/ml4ir/base/pipeline.py
mohazahran: I thought the memory issues were related to creating multiple RelevanceDatasets.
will look into that

python/ml4ir/base/pipeline.py
mohazahran: it did not, because the function "run" calls "run_pipeline"

python/ml4ir/applications/classification/pipeline.py
```diff
@@ -160,6 +160,16 @@ def get_relevance_dataset(
 
         return relevance_dataset
 
+    def create_pipeline_for_kfold(self, args):
+        """
+        Create a ClassificationPipeline object used in running kfold cross validation.
+        """
+        return ClassificationPipeline(args=args)
```
arvindsrikantan, line 167: Why is this needed if all we are doing is calling `__init__`?

python/ml4ir/applications/ranking/tests/test_ranking_kfold_cv.py
```diff
@@ -0,0 +1,170 @@
...
+            train_qids = set([q[0]['query_id'].numpy()[0] for q in train])
+            validation_qids = set([q[0]['query_id'].numpy()[0] for q in validation])
+            if use_testset_in_folds:
+                test_qids = set([q[0]['query_id'].numpy()[0] for q in test])
+            if use_testset_in_folds:
+                assert len(set.intersection(train_qids, test_qids)) == 0
+                assert len(set.intersection(validation_qids, test_qids)) == 0
+            assert len(set.intersection(train_qids, validation_qids)) == 0
+
+    def test_merge_datasets_1(self):
```
arvindsrikantan, line 110: The test names aren't descriptive. At the very least, we need a doc string explaining what this function is trying to test

python/ml4ir/base/data/kfold_relevance_dataset.py
```diff
@@ -0,0 +1,98 @@
...
+        self.keep_additional_info = keep_additional_info
+        self.non_zero_features_only = non_zero_features_only
+
+        self.train: Optional[tf.data.TFRecordDataset] = None
+        self.validation: Optional[tf.data.TFRecordDataset] = None
+        self.test: Optional[tf.data.TFRecordDataset] = None
+        if read_data_sets:
+            self.create_dataset(parse_tfrecord)
+
+    def create_folds(self, fold_id, all_data):
```
arvindsrikantan, line 60: [Minor] `fold_idx` seems to be more unambiguous

python/ml4ir/base/data/kfold_relevance_dataset.py
```diff
@@ -0,0 +1,98 @@
...
+            test_pcent_split: float = -1,
+            use_part_files: bool = False,
+            parse_tfrecord: bool = True,
+            logger: Optional[Logger] = None,
+            keep_additional_info: int = 0,
+            non_zero_features_only: int = 0,
+            num_folds: int = 3,
+            include_testset_in_kfold: bool = False,
+            read_data_sets: bool = False
+    ):
```
arvindsrikantan, line 32: This method needs a doc string explaining the params

python/ml4ir/base/pipeline.py
```diff
@@ -243,6 +249,53 @@ def get_relevance_dataset(self, preprocessing_keys_to_fns={}) -> RelevanceDatase
 
         return relevance_dataset
 
+    def get_kfold_relevance_dataset(self, num_folds, include_testset_in_kfold, read_data_sets, preprocessing_keys_to_fns={}) -> RelevanceDataset:
+        """
+        Create RelevanceDataset object by loading train, test data as tensorflow datasets
+
+        Parameters
+        ----------
```
arvindsrikantan, line 257: Some args are missing

python/ml4ir/base/pipeline.py
```diff
@@ -265,9 +318,128 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
         """
         raise NotImplementedError
 
+    def kfold_analysis(self, base_logs_dir, run_id, num_folds):
+        """
+        Aggregate results of the k-fold runs and perform t-test on the results.
+        """
+        raise NotImplementedError
```
arvindsrikantan, line 325: Why are there 2 methods with the same name?

python/ml4ir/base/pipeline.py
```diff
@@ -265,9 +318,128 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+
+    def create_pipeline_for_kfold(self, args):
+        raise NotImplementedError
```
arvindsrikantan, line 328: Can't we have a generic definition here?
I feel this method is redundant if it always calls the respective init

python/ml4ir/base/pipeline.py
```diff
@@ -265,9 +318,128 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+            current fold number
+        num_folds: int
+            Total number of folds
+        metrics: set()
+            Set of metrics to include in the kfold analysis
+        """
+        if not metrics:
+            return
+        rows = []
+        pvalue_threshold = 0.1
```
arvindsrikantan, line 349: Should this be configurable? Another option is to just report it at 0.1, 0.05 and 0.01

python/ml4ir/base/pipeline.py
```diff
@@ -265,9 +318,128 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+        for i in range(num_folds):
+            row = {"fold": i}
+            logs_dir = pathlib.Path(base_logs_dir) / run_id / "fold_{}".format(i) / run_id
```
arvindsrikantan, line 352: Why `run_id` twice?

python/ml4ir/base/pipeline.py
```diff
@@ -265,9 +318,128 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+            with open(logs_dir / "_SUCCESS", 'r') as f:
```
arvindsrikantan, line 353: @lastmansleeping , we should probably start writing to a better filename considering we don't use the old HPO anymore

python/ml4ir/base/pipeline.py
```diff
@@ -265,9 +318,128 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+            Total number of folds
+        run_id : int
+            current fold number
+        num_folds: int
+            Total number of folds
+        metrics: set()
+            Set of metrics to include in the kfold analysis
+        """
+        if not metrics:
+            return
```
arvindsrikantan, line 347: Shouldn't we throw an error or log a warning here?

python/ml4ir/base/pipeline.py
```diff
@@ -265,9 +318,128 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+            return self.run_pipeline()
+
+        if self.args.include_testset_in_kfold:
+            if self.args.kfold < 3:
+                raise Exception("Number of folds must be > 2")
+        else:
+            if self.args.kfold < 2:
+                raise Exception("Number of folds must be > 1")
+
+        args = copy.deepcopy(self.args)
```
arvindsrikantan, line 396: Why?

python/ml4ir/base/pipeline.py
```diff
@@ -265,9 +318,128 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+
+        # reading, parsing the dataset (train, validation, test)
+        self.logger.info("Reading datasets ...")
+        relevance_dataset = self.get_relevance_dataset()
+        self.logger.info("Relevance Dataset created")
+
+        all_data = relevance_dataset.merge_datasets(include_testset_in_merge=args.include_testset_in_kfold)
```
arvindsrikantan, line 403: Why use args here and self.args right below?

python/ml4ir/base/pipeline.py
arvindsrikantan: I think I see it below.

python/ml4ir/base/pipeline.py
```diff
@@ -265,9 +318,128 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+
+        num_folds = self.args.kfold
+        base_logs_dir = str(self.args.logs_dir)
+        base_models_dir = str(self.args.models_dir)
```
arvindsrikantan, line 407: Curious: why `str()` and later `Path()` below?

python/ml4ir/base/pipeline.py
```diff
@@ -265,9 +318,128 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+            fold_relevance_dataset = self.get_kfold_relevance_dataset(args.kfold,
+                                                                      args.include_testset_in_kfold,
+                                                                      read_data_sets=False)
+            fold_relevance_dataset.create_folds(fold_id, all_data)
+            pipeline = self.create_pipeline_for_kfold(args)
+            fold_result = pipeline.run_pipeline(fold_relevance_dataset)
+            folds_results[fold_id] = fold_result
+
+        # removing intermediate directory and run kfold analysis
+        self.local_io.rm_dir(os.path.join(self.data_dir_local, "tfrecord"))
```
arvindsrikantan, line 430: Should extract to a method and reuse here and in `run_pipeline`

python/ml4ir/applications/classification/pipeline.py
mohazahran: because we can be creating either a classification pipeline or a ranking pipeline

python/ml4ir/base/pipeline.py
mohazahran: I think we can throw a warning here

python/ml4ir/base/pipeline.py
mohazahran: later on I will modify the copied args object to initiate a pipeline run

python/ml4ir/base/pipeline.py
mohazahran: for that line it doesn't matter, but I agree I will change it for consistency.

python/ml4ir/applications/classification/pipeline.py
```diff
@@ -160,6 +160,16 @@ def get_relevance_dataset(
...
+
+    def run_kfold_analysis(self, base_logs_dir, base_run_id, num_folds):
+        # TODO: Implement the kfold CV analysis for classification
```
lastmansleeping, line 170: @mohazahran What is stopping us from doing this? I would like this to be generic enough for all ml4ir pipelines ideally.

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -21,6 +21,7 @@
 
 class RankingConstants:
     NEW_RANK = "new_rank"
+    KFOLD_METRICS = {'train_old_MRR', 'train_new_MRR', 'val_old_MRR', 'val_new_MRR', 'test_old_MRR', 'test_new_MRR'}
```
lastmansleeping, line 24: @mohazahran As we discussed, I'm not sure why you are hardcoding this. You should create an argument and make these default at the least. How do you plan to make this work for classification?

python/ml4ir/base/data/kfold_relevance_dataset.py
lastmansleeping: +1

python/ml4ir/base/data/kfold_relevance_dataset.py
lastmansleeping: Also maybe rename to `fetch_fold_dataset`?

python/ml4ir/base/data/kfold_relevance_dataset.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def create_folds(self, fold_id, all_data):
+        """
+        Create training, validation and test set according to the passed fold id.
+
+        Parameters
+        ----------
+        fold_id : int
+            current fold number
+        all_data: Tensorflow Dataset
```
lastmansleeping, line 68: @mohazahran It seems unnecessary to have an argument for `all_data` when you have a class that does KFold specifically. Can you set `self.merged_data` in the `__init__()`? You should create a separate method for merging if needed.

python/ml4ir/base/data/kfold_relevance_dataset.py
```diff
@@ -0,0 +1,98 @@
...
+        self.use_part_files: bool = use_part_files
+
+        self.keep_additional_info = keep_additional_info
+        self.non_zero_features_only = non_zero_features_only
+
+        self.train: Optional[tf.data.TFRecordDataset] = None
+        self.validation: Optional[tf.data.TFRecordDataset] = None
+        self.test: Optional[tf.data.TFRecordDataset] = None
+        if read_data_sets:
+            self.create_dataset(parse_tfrecord)
```
lastmansleeping, line 58: @mohazahran Why do you need to do this step again if you already have and plan to use the merged dataset in the method below?

python/ml4ir/base/pipeline.py
```diff
@@ -265,9 +324,122 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+            rows.append(row)
+        results = pd.DataFrame(rows)
+        self.logger.info(results)
+        # calculate statistical paired t-test
+        for dataset in ['train', 'val', 'test']:
+            t_test_stat, pvalue = stats.ttest_rel(results["{}_old_MRR".format(dataset)],
+                                                  results["{}_new_MRR".format(dataset)])
+            self.logger.info("{}_t_test_stat={}".format(dataset, t_test_stat))
+            self.logger.info(
+                "{}_pvalue={} --> (statistically significant={})".format(dataset, pvalue, pvalue < pvalue_threshold))
```
lastmansleeping, line 367: @mohazahran You need to also find a way to add this to `metrics_dict`.

python/ml4ir/base/pipeline.py
lastmansleeping: True. Will create a WI for this.

python/ml4ir/base/pipeline.py
lastmansleeping: The reason I keep pushing for this is because you are not thinking in terms of training models at scale. How would you fetch these values from the logs using, say a script?

python/ml4ir/base/pipeline.py
```diff
@@ -265,9 +324,122 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+        rows = []
+        for i in range(num_folds):
+            row = {"fold": i}
+            logs_dir = pathlib.Path(base_logs_dir) / run_id / "fold_{}".format(i) / run_id
+            with open(logs_dir / "_SUCCESS", 'r') as f:
+                for line in f:
+                    parts = line.split(',')
+                    if parts[0] in metrics:
+                        row[parts[0]] = float(parts[1])
+            rows.append(row)
```
lastmansleeping, line 358: @mohazahran Does this print the metrics for every single KFold run?

python/ml4ir/base/pipeline.py
```diff
@@ -265,9 +324,122 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+        results = pd.DataFrame(rows)
+        self.logger.info(results)
+        # calculate statistical paired t-test
+        for dataset in ['train', 'val', 'test']:
```
lastmansleeping, line 362: @mohazahran Can you also leave a TODO item for selecting the best model as well? I think that should be the last piece of this. I understand the complications though.

python/ml4ir/applications/classification/pipeline.py
arvindsrikantan: But aren't you creating an instance of the respective class in this method? Why would you create ranking pipeline in ClassificationPipeline class?

python/ml4ir/applications/ranking/model/ranking_model.py
arvindsrikantan: This might be a PITA comment: Can we consistently use `"` for strings?

python/ml4ir/base/data/kfold_relevance_dataset.py
mohazahran: I believe returning the merged_data might be better as it can be used with other newly created Kfold objects, so that each object would slice it according to its fold_idx

python/ml4ir/base/pipeline.py
mohazahran: no, just once after all folds are completed

python/ml4ir/base/data/kfold_relevance_dataset.py
mohazahran: The reason for this check is needing to read data from files (creating the dataset) once before iterating through the folds. Then for each fold a new KfoldRelevanceDataset is created (without the need re-read from files). The new object just takes the merged data and slice it according to its fold_idx

python/ml4ir/applications/classification/pipeline.py
mohazahran: As discussed in the call, for ranking pipeline, I compare old vs. new models with ttest. Classification -as far as I know- does not have old, new model results.

PR #142, opened 2021-10-01
Adding softmax cross entropy loss

lastmansleeping:


python/ml4ir/applications/ranking/model/losses/listwise_losses.py
```diff
@@ -6,11 +6,10 @@
...
-class RankOneListNet(ListwiseLossBase):
+class SoftmaxCrossEntropy(ListwiseLossBase):
     def get_loss_fn(self, **kwargs):
         """
-        Define a masked rank 1 ListNet loss
-        Additionally can pass in record positions to handle positional bias
+        Define a softmax cross entropy loss
 
         Returns
         -------
```
Ullimague, line 15: I don't think line 17 is valid anymore.

lastmansleeping: My bad. Good catch!

python/ml4ir/applications/ranking/model/losses/listwise_losses.py
```diff
@@ -33,18 +32,7 @@ def _loss_fn(y_true, y_pred):
...
-
-            # Reshape the tensors so that we sum the losses from each record
-            y_true = tf.expand_dims(tf.squeeze(y_true), axis=-1)
-            y_pred = tf.expand_dims(tf.squeeze(y_pred), axis=-1)
-
-            # Scale the sum of losses down by number of queries in the batch
-            return tf.math.divide(bce(y_true, y_pred), batch_size)
+            return cce(y_true, tf.math.multiply(y_pred, mask))
 
         return _loss_fn
```
Ullimague, line 38: Now get_final_activation_op is part of another class. There is no breaking changes or tests about it?

lastmansleeping: I am inheriting that class, right?

lastmansleeping: Both the losses use the same activation. They just use a different method to compute the final loss.

Ullimague: Oh I missed that. Do we want to have a common parent class for it? Or are we fine keeping it that way?

lastmansleeping: I think it should be fine because Rank One ListNet loss is a variation of the softmax cross entropy loss, that's all.

Ullimague (approved):

python/ml4ir/applications/ranking/tests/test_losses.py
```diff
@@ -62,3 +62,12 @@ def test_rank_one_listnet(self):
         loss = self.run_default_pipeline(loss_key="rank_one_listnet")
 
         assert np.isclose(loss, 2.00879, rtol=0.05)
+
+    def test_softmax_cross_entropy(self):
+        """
+        Test model training and evaluate Rank One ListNet loss
```
balikasg, line 68: This docstring should be about softmax cross-entropy loss.

python/ml4ir/applications/ranking/tests/test_losses.py
```diff
@@ -62,3 +62,12 @@ def test_rank_one_listnet(self):
...
+        """
+
+        loss = self.run_default_pipeline(loss_key="softmax_cross_entropy")
+
+        assert np.isclose(loss, 1.19955, rtol=0.05)
```
balikasg, line 73: These tests, where we run the pipeline and get `np.isclose` to something, have value, in the sense that the pipeline runs. They boost coverage too! 
This is not the whole story: ideally we want to eval the actual code in `SoftmaxCrossEntropy`  and `get_loss_fn` where the tensor calculations happen. Would you consider adding such a "toy" test?

python/ml4ir/applications/ranking/model/losses/listwise_losses.py
```diff
@@ -6,23 +6,22 @@
 from ml4ir.applications.ranking.model.losses.loss_base import ListwiseLossBase
 
 
-class RankOneListNet(ListwiseLossBase):
+class SoftmaxCrossEntropy(ListwiseLossBase):
```
balikasg, line 9: Are there any references we can add? Not paragraphs but at least a pointer for someone to understand a bit more. For instance, to me the difference with `categorical_cross_entropy` is not evident (I am lacking ranking context). Is is suitable both for binary and more fine-grained relevance?

python/ml4ir/applications/ranking/model/losses/loss_factory.py
```diff
@@ -26,5 +26,7 @@ def get_loss(loss_key, scoring_type) -> RelevanceLossBase:
         return pointwise_losses.SigmoidCrossEntropy(loss_key=loss_key, scoring_type=scoring_type)
     elif loss_key == LossKey.RANK_ONE_LISTNET:
         return listwise_losses.RankOneListNet(loss_key=loss_key, scoring_type=scoring_type)
+    elif loss_key == LossKey.SOFTMAX_CROSS_ENTROPY:
```
balikasg, line 29: the separation in poitwise_losses and listwise_losses reads nice here ;-)

balikasg (approved):
Few comments, but nothing that blocks this PR!

python/ml4ir/applications/ranking/tests/test_losses.py
lastmansleeping: I agree. I am going to change all of them to just unit tests actually. It will help with memory leak as well. I think the current build is failing because of memory overflow.

balikasg: Did you have memory issues when training actual models (not toy in tests)? I have been training for several days but without issues for instance (which is good). Perhaps I had more memory.. But I remember having issues with tests (locally)

lastmansleeping: Not on my local box. I think it only breaks on the autobuilds because of the docker container hard limitation. Our local boxes have a lot of swap size for handling this (is my guess). 

The main issue is that in tf2, keras models are not releasing memory. So for a test like this, we create the pipeline many times (so we are creating and deleting keras models) and the model is not getting "garbage collected" fully. :|

PR #144, opened 2021-10-21
Adding reduce_lr_on_plateau scheduling learning rate

mohazahran:


python/ml4ir/applications/ranking/pipeline.py
```diff
@@ -124,6 +126,20 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+            if lr_schedule_key == LearningRateScheduleKey.REDUCE_LR_ON_PLATEAU:
+                monitor_metric = self.args.monitor_metric
+                if not monitor_metric.startswith("val_"):
+                    monitor_metric = "val_{}".format(monitor_metric)
+                reduce_lr = tf.keras.callbacks.ReduceLROnPlateau(monitor=monitor_metric,
+                                                                 factor=lr_schedule.get('factor', 0.5),
+                                                                 patience=lr_schedule.get('patience', 1),
+                                                                 min_lr=lr_schedule.get('min_lr', 0.0001), verbose=1)
+                relevance_model.callbacks_list.append(reduce_lr)
+
```
lastmansleeping, line 142: @mohazahran I had asked you to move this out of the pipeline. Add it to the model class that way we can use it with ml4ir as a library. pipelines are very restrictive.

python/ml4ir/base/model/relevance_model.py
```diff
@@ -403,6 +405,10 @@ def fit(
             patience=patience,
         )
 
+        if hasattr(self, 'callbacks_list'):
```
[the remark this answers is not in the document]
lastmansleeping: ping

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -216,8 +225,100 @@ def test_cyclic_lr_in_training_pipeline(self):
 
         assert np.all(np.isclose(lr_gold, lr_list))
 
+    def test_reduce_lr_on_plateau_in_training_pipeline(self):
```
lastmansleeping, line 228: @mohazahran I think you should probably add a test for the callback function alone. Creating new tensorflow models has been very memory intensive on the autobuilds and we should avoid it if possible.

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -216,8 +225,88 @@ def test_cyclic_lr_in_training_pipeline(self):
...
+        relevance_model: RelevanceModel = RankingModel(
+            feature_config=feature_config,
+            tfrecord_type=TFRecordTypeKey.SEQUENCE_EXAMPLE,
+            scorer=scorer,
+            optimizer=optimizer,
+            model_file=None,
+            file_io=io,
+            logger=Logger,
+            model_config=model_config,
+        )
```
lastmansleeping, line 291: @mohazahran Can you test this without creating a RankingModel? I ask this because creating new models is resulting in memory overflow on the autobuilds.

python/ml4ir/base/model/relevance_model.py
```diff
@@ -350,6 +361,27 @@ def from_univariate_interaction_model(
...
+                                                                     min_lr=lr_schedule.get('min_lr', 0.0001),
+                                                                     verbose=1)
+                elif not monitor_metric.startswith("val_"):
+                    monitor_metric = "val_{}".format(monitor_metric)
+                    reduce_lr = tf.keras.callbacks.ReduceLROnPlateau(monitor=monitor_metric,
+                                                                     factor=lr_schedule.get('factor', 0.5),
+                                                                     patience=lr_schedule.get('patience', 1),
+                                                                     min_lr=lr_schedule.get('min_lr', 0.0001),
+                                                                     verbose=1)
+                self.callbacks_list.append(reduce_lr)
```
lastmansleeping, line 383: @mohazahran Move this code out of this module. You can have the callback addition here (down). The backend logic should be in a new folder, perhaps `callbacks/`

python/ml4ir/base/model/relevance_model.py
```diff
@@ -404,6 +437,9 @@ def fit(
             patience=patience,
         )
 
+        callbacks_list.extend(self.callbacks_list)
+        self.callbacks_list = []
```
lastmansleeping, line 441: @mohazahran Add a comment on what is happening here. I'm confused.

lastmansleeping (commented):
Looking good. Left a few restructuring comments.

python/ml4ir/base/model/callbacks/debugging.py
```diff
@@ -29,6 +29,7 @@ def on_epoch_begin(self, epoch, logs=None):
         self.epoch = epoch + 1
         self.logger.info("Starting Epoch : {}".format(self.epoch))
         self.logger.info(logs)
+        self.logger.info('lr={}'.format(str(self.model.optimizer._decayed_lr(float).numpy())))
```
lastmansleeping, line 32: @mohazahran Will `model.optimizer` have a `_decayed_lr` attribute when running with other LR settings?

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
lastmansleeping: One (admittedly convoluted) way to do this is by just calling `on_epoch_end()` multiple times to test the learning rate updates. I agree that this is essentially testing `ReduceLrOnPlateau` and not whether it works in ml4ir. So for now, you can leave it as is I think. 

I'm also wondering if we even need an expensive test such as this one from our end. Maybe the real test should be to see if the right monitoring metric was added by the `add_scheduler_as_callback` and everything was configured right. That's the ml4ir part really, right?

https://github.com/keras-team/keras/blob/v2.7.0/keras/callbacks.py#L2603-L2724

lastmansleeping (commented):
Made a few comments. Let me know once you've addressed them. Looks simple and good though. :)

python/ml4ir/base/model/optimizers/optimizer.py
```diff
@@ -103,6 +103,8 @@ def choose_scheduler(model_config):
                 decay_steps=10000000,
                 decay_rate=1.0,
             )
+        elif lr_schedule_key == LearningRateScheduleKey.REDUCE_LR_ON_PLATEAU:
+            learning_rate_schedule = lr_schedule.get('learning_rate', OptimizerDefaultValues.CONSTANT_LR)
```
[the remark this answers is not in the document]
mohazahran: Yes

docs/source/quickstart/model_config.md
```diff
@@ -87,3 +94,15 @@ lr_schedule:
   learning_rate_decay_steps: 100000   #default value is 100000
   learning_rate_decay: 0.96              #default value is 0.96
 ```
+
+
+reduce_lr_on_plateau Learning Rate Schedule
```
balikasg, line 99: Thanks for adding this here, very practical. 
Also, thanks for being concise and to the point.

python/ml4ir/applications/ranking/pipeline.py
```diff
@@ -122,6 +123,8 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
             output_name=self.args.output_name,
             file_io=self.local_io,
             logger=self.logger,
+            model_config=self.model_config,
```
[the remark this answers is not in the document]
balikasg: do we need move the model_file and the model_config? Unless the model_file is the target to save the model..

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -27,6 +27,7 @@
 
 INPUT_DIR = "ml4ir/applications/ranking/tests/data/"
 MODEL_CONFIG = "ml4ir/applications/ranking/tests/data/configs/model_config_cyclic_lr.yaml"
+MODEL_CONFIG_REDUCE_LR_ON_PLATEAU = "ml4ir/applications/ranking/tests/data/configs/model_config_reduce_lr_on_plateau.yaml"
```
balikasg, line 30: What do you think about not committing actual files in git (like the config you committed) and simply writing/removing then in setUp, tearDown ? The advantage being that you have them in front of your eyes when the debug the test.

python/ml4ir/applications/ranking/tests/data/configs/model_config_reduce_lr_on_plateau.yaml
```diff
@@ -0,0 +1,15 @@
...
+    units: 1
+    kernel_initializer: Ones
+    use_bias: False
+    activation:
+optimizer:
+  key: sgd
+  gradient_clip_value: 100.0
+lr_schedule:
+  key: reduce_lr_on_plateau
+  learning_rate: 50.0
```
balikasg, line 14: A somewhat extreme value even for a test...!

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
balikasg: @mohazahran I hate to be a party pooper here but these comments that are left unanswered (perhpas you synced offline) is not reassuring. 
There are two issues with tests in ml4ir: 
- We run out of memory
- They take too long
The tests here contribute to both problems. Can we do something else to validate the functionality?
If not:
- Just use 3 epochs
- Ensure you are using toy data like 20 examples and not 2K 
Thoughts?

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
```diff
@@ -216,8 +224,90 @@ def test_cyclic_lr_in_training_pipeline(self):
...
+            epochs=10,
+            verbose=True,
+            callbacks=callback_list,
+        )
+        lr_list = my_callback_object.get_lr_reduce_on_plateau_list()
+        lr_gold = [50.0, 50.0, 25.0, 12.5, 6.25, 3.125, 1.5625, 1.0, 1.0, 1.0]
+
+        assert np.all(np.isclose(lr_gold, lr_list))
+
     def tearDown(self):
```
balikasg, line 308: Do we really need a `tearDown` that passes?

python/ml4ir/base/model/relevance_model.py
```diff
@@ -350,6 +357,27 @@ def from_univariate_interaction_model(
             logger=logger,
         )
 
+    def add_scheduler_as_callback(self):
+        """Adding reduce lr on plateau as a callback if specified"""
```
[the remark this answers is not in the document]
balikasg: Let;s add the docstring and rename to `define_scheduler_as_callback`

python/ml4ir/base/model/relevance_model.py
```diff
@@ -350,6 +357,28 @@ def from_univariate_interaction_model(
...
+            lr_schedule = model_config['lr_schedule']
+            lr_schedule_key = lr_schedule['key']
+            if lr_schedule_key == LearningRateScheduleKey.REDUCE_LR_ON_PLATEAU:
+                if monitor_metric is None:
+                    reduce_lr = tf.keras.callbacks.ReduceLROnPlateau(factor=lr_schedule.get('factor', 0.5),
+                                                                     patience=lr_schedule.get('patience', 1),
+                                                                     min_lr=lr_schedule.get('min_lr', 0.0001),
+                                                                     mode=lr_schedule.get('mode', 'auto'),
+                                                                     verbose=1)
+                elif not monitor_metric.startswith("val_"):
```
balikasg, line 372: This is a personal taste comment; using if/elif in this way (without else) leaves the reader wondering if anything is missing. Here, we coud have 
```python
if monitor_metric is None:
  return tf.keras...()
if not monitor_metric.startswith("val_"):
  return tf.keras...()
```
which still makes me wondering if we are missing the case that the user acutally passes a valid monitor metric. it seems like passing `val_accuracy` is not treated here. 
- Are we treating the case where the user passes a valid monitoring metric?

python/ml4ir/base/model/relevance_model.py
```diff
@@ -350,6 +357,27 @@ def from_univariate_interaction_model(
...
+                if monitor_metric is None:
+                    reduce_lr = tf.keras.callbacks.ReduceLROnPlateau(factor=lr_schedule.get('factor', 0.5),
+                                                                     patience=lr_schedule.get('patience', 1),
+                                                                     min_lr=lr_schedule.get('min_lr', 0.0001),
+                                                                     verbose=1)
+                elif not monitor_metric.startswith("val_"):
+                    monitor_metric = "val_{}".format(monitor_metric)
+                    reduce_lr = tf.keras.callbacks.ReduceLROnPlateau(monitor=monitor_metric,
+                                                                     factor=lr_schedule.get('factor', 0.5),
+                                                                     patience=lr_schedule.get('patience', 1),
```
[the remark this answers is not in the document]
balikasg: Resolving this, as the change is implemented.

python/ml4ir/base/model/relevance_model.py
```diff
@@ -828,6 +858,11 @@ def _build_callback_hooks(
         # Debugging/Logging
         callbacks_list.append(DebuggingCallback(self.logger, logging_frequency))
 
+        # Adding lr scheduler as a callback
```
balikasg, line 861:
```suggestion
        # Adding lr scheduler as a callback; used for `ReduceLROnPlateau` which we treat today as a callback
```

balikasg: this is possible PITA, but it may be useful for our future selves.

balikasg (commented):
@mohazahran :
- First thanks for the clean PR that comes with tests and documentation. 
- Please let me know what you think about the tests. It is an issue because of the reasons mentioned in the comments. 
- Also, take a look at the actual definition of the callback; there is a chance we are missing a case when the user provides the expercted metric (prepended with val). Our doc does not specify how this metric should look so in any case it is a gap.

python/ml4ir/applications/ranking/tests/data/configs/model_config_reduce_lr_on_plateau.yaml
mohazahran: Sure :) but I intentionally wanted the learning rate to be extremely bad so that the optimizer hits a plateau and lr reductions kicks in.

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
mohazahran: was used as a place holder of what might be needed to be cleaned, but left. 
I will take it out

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
lastmansleeping: We have not synced offline. Even I am awaiting replies to move forward with the PR.

python/ml4ir/base/model/relevance_model.py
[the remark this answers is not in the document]
lastmansleeping: It is not resolved. My question is should the patience be 1 or more?

lastmansleeping (changes requested):
@mohazahran 
- Remove `model_config` and `monitor_metric` arguments from `RelevanceModel` constructor and calls as it is not needed anymore.
- Add docstrings to `define_scheduler_as_callback`

python/ml4ir/base/model/relevance_model.py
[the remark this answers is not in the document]
mohazahran: I don't think there is a right answer here. the default number by tf is 10 which I think is too high. I will put it to 5 as a compromise.

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
lastmansleeping: @balikasg I don't think the data size really matters other than adding a few (milli) seconds to the training time. The fact that we create yet another model is the real issue and should be looked at in more detail (not part of this PR)

python/ml4ir/applications/ranking/pipeline.py
```diff
@@ -1,6 +1,7 @@
 import sys
 import ast
 from argparse import Namespace
+import tensorflow as tf
```
lastmansleeping, line 4: @mohazahran Do you need this?

python/ml4ir/base/model/relevance_model.py
```diff
@@ -350,6 +351,44 @@ def from_univariate_interaction_model(
...
+
+        if model_config and 'lr_schedule' in model_config:
+            lr_schedule = model_config['lr_schedule']
+            lr_schedule_key = lr_schedule['key']
+            if lr_schedule_key == LearningRateScheduleKey.REDUCE_LR_ON_PLATEAU:
+                if monitor_metric is None:
+                    reduce_lr = tf.keras.callbacks.ReduceLROnPlateau(factor=lr_schedule.get('factor', 0.5),
+                                                                     patience=lr_schedule.get('patience', 5),
+                                                                     min_lr=lr_schedule.get('min_lr', 0.0001),
+                                                                     mode=lr_schedule.get('mode', 'auto'),
```
lastmansleeping, line 379: @mohazahran Shouldn't this be `monitor_mode` from `fit()`?

mohazahran: I defined this mode through the model config file. but I see your point, It could be yes.

python/ml4ir/applications/ranking/tests/test_lr_schedule.py
mohazahran: I see your point, but since this is the way ml4ir is operating (reading config files) I followed it.

python/ml4ir/base/model/relevance_model.py
lastmansleeping: Should be okay for now.

lastmansleeping (approved):
LGTM. Thanks for the changes. I think it is at a state where it can be merged and used for training. Please make a note of clean up we should do in the future, if any.

python/ml4ir/base/model/relevance_model.py
mohazahran: I'm using the passed metric, it should be fine now.

python/ml4ir/applications/ranking/pipeline.py
arvindsrikantan: Can be removed right?

python/ml4ir/base/model/callbacks/debugging.py
arvindsrikantan: @mohazahran , waiting on your confirmation here. Based on tf code it seems to be the case

python/ml4ir/base/tests/test_lr_schedule.py
```diff
@@ -216,8 +224,86 @@ def test_cyclic_lr_in_training_pipeline(self):
 
         assert np.all(np.isclose(lr_gold, lr_list))
 
-    def tearDown(self):
-        pass
+    def test_reduce_lr_on_plateau_in_training_pipeline(self):
```
arvindsrikantan, line 227: I see a lot of repetition across all test functions. This should be made cleaner by extracting out the common parts

python/ml4ir/base/model/callbacks/debugging.py
mohazahran: it seems that way yes. I tried with both sgd and adam

python/ml4ir/base/tests/test_lr_schedule.py
mohazahran: There is only two tests that share some modules, however since these modules are dependent and changes from one test to the other is across different modules, I think there is no very clean way to combine both.

arvindsrikantan (approved):
Thanks for the change and response!

PR #147, opened 2021-11-17
Splitting tests and creating new autobuild @rev mohamed-m@

lastmansleeping:


python/ml4ir/base/tests/test_ml4ir_sanity.py
```diff
@@ -203,14 +188,8 @@ def test_linear_ml4ir_sanity_7(self):
                         log_regression_mrr=0.4938108725899423,
                         working_dir=pathlib.Path(self.working_dir.path), log_dir=pathlib.Path(self.log_dir))
 
-    def _test_linear_ml4ir_sanity_8(self):
+    def test_linear_ml4ir_sanity_8(self):
```
lastmansleeping, line 191: Some of these tests were deactivated earlier due to memory issues. I am reactivating them now.

python/ml4ir/applications/ranking/tests/test_base.py
```diff
@@ -33,53 +34,7 @@
 FEATURE_CONFIG_FNAME = "feature_config.yaml"
 
 
-class RankingTestBase(unittest.TestCase):
```
lastmansleeping, line 36: Moving this to a `RelevanceTestBase` class for ease of use.

python/ml4ir/base/tests/test_ml4ir_sanity.py
balikasg: These tests may run for up to 150 epochs with batch size of 1. This is in line 46-48. 
How about doing 2-3 epochs with larger batches? Can be a followup WI. 
We should be also reconsidering if all these tests (with probably copy-paste numbers) indeed add value. 
Thoughts? 
cc: @mohazahran since you have added them. If you have a fresher look on this, please chime in

lastmansleeping: Yep, would like to get @mohazahran 's thoughts on updating these.

I'm for deprecating a few of these.

.circleci/config.yml
```diff
@@ -129,7 +162,7 @@ jobs:
             . venv/bin/activate
             coverage combine reports/.coverage.ranking reports/.coverage.classif
             coverage report
-            coverage html --fail-under=85
+            coverage html --fail-under=75
```
tayyabj, line 165: Is there any way we can keep the old threshold (85)?

.circleci/config.yml
```diff
@@ -35,6 +35,39 @@ jobs:
       - store_test_results:
           path: jvm/ml4ir-inference/target
 
+  build_test_python_base:
```
tayyabj, line 38: this needs to be added to `workflows` below

python/ml4ir/base/tests/test_base.py
```diff
@@ -0,0 +1,85 @@
...
+
+warnings.filterwarnings("ignore")
+
+
+OUTPUT_DIR = "ml4ir/applications/ranking/tests/test_output"
+ROOT_DATA_DIR = "ml4ir/applications/ranking/tests/data"
+FEATURE_CONFIG_FNAME = "feature_config.yaml"
+
+
+class RelevanceTestBase(unittest.TestCase):
```
tayyabj, line 36: add a docstring mentioning this is the base test class for the repo - `Relevance` seems to throw me off

python/ml4ir/base/tests/test_feature_layer.py
```diff
@@ -1,14 +1,14 @@
...
 from ml4ir.base.features.feature_fns import tf_native as tf_native_fns
 from ml4ir.base.config.keys import SequenceExampleTypeKey
+from ml4ir.base.tests.test_base import RelevanceTestBase
 
 import tensorflow as tf
 import numpy as np
 
 
-class RankingModelTest(RankingTestBase):
+class RankingModelTest(RelevanceTestBase):
```
tayyabj, line 11: If this is not a ranking test, please rename it to something else

.circleci/config.yml
lastmansleeping: I'm not sure why the code coverage decreased. My guess is that it is now scanning the `base/` paths as well whereas it might not have earlier, but I am not confident.

tayyabj (approved):

python/ml4ir/base/tests/test_ml4ir_sanity.py
mohazahran: Initially this PR was training all the models (linear regression, perceptron and ml4ir0 on the spot, so it was easier to configure the comparison between them by changing the parameters (batch size, training set size ...). but since LR and perceptron required extra dependencies only their weights are used after training them offline. So updating these will required re-training these models and updating the weights.
One possible this we can do is we can retire the perceptron (as it's inferior to the other two models) but i dont think this will really help the run time.

mohazahran (approved):
Looks good, let's discuss probably cutting down some sanity tests in a separate PR if they contribute to the memory issues.

PR #148, opened 2021-11-23
Refactoring ranking metrics helper and adding unit tests

lastmansleeping:
Refactoring the ranking metrics helper into smaller unit testable modules to add NDCG metric in follow up PR.

- [x] Refactor and verify everything works manually
- [x] Add docstrings
- [x] Add unittests

I will update the TODO above, but let me know your initial thoughts @mohazahran @arvindsrikantan 

Follow up PR to add NDCG -> https://github.com/salesforce/ml4ir/pull/149

**Note** - There is no new functionality added in this PR. It is a refactor/code cleanup. I also added some unittests (which were missing) as a bonus.


python/ml4ir/applications/ranking/tests/test_metrics.py
```diff
@@ -12,26 +12,26 @@
...
-                'mean_old_name_match_failure_any_rank': 0.351,
-                'mean_old_name_match_failure_any_count': 0.165,
-                'mean_old_name_match_failure_any_fraction': 0.087,
-                'mean_new_name_match_failure_all': 0.086,
-                'mean_new_name_match_failure_any': 0.218,
-                'mean_new_name_match_failure_all_rank': 0.196,
-                'mean_new_name_match_failure_any_rank': 0.788,
-                'mean_new_name_match_failure_any_count': 0.342,
-                'mean_new_name_match_failure_any_fraction': 0.153,
+                'old_name_match_failure_all_mean': 0.061,
```
lastmansleeping, line 15: If you notice, none of the values have changed. I just changed "mean" to the suffix so that it is easier to handle in a generic way.

python/ml4ir/applications/ranking/model/metrics/metrics_helper.py
```diff
@@ -2,143 +2,156 @@
...
+        "{}{}_failure_any".format(prefix, secondary_label): failure_any,
+        "{}{}_failure_all_rank".format(prefix, secondary_label): click_rank
+        if failure_all
+        else 0,
+        "{}{}_failure_any_rank".format(prefix, secondary_label): click_rank
+        if failure_any
+        else 0,
+        "{}{}_failure_any_count".format(prefix, secondary_label): failure_count,
+        "{}{}_failure_any_fraction".format(prefix, secondary_label): failure_fraction,
+    }
```
lastmansleeping, line 90: This is not new code. I am simply removing it into a separate method, so that it is cleaner, unit testable and also easy to add new metrics like NDCG.

python/ml4ir/applications/ranking/tests/test_metrics_helper.py
```diff
@@ -0,0 +1,264 @@
+import unittest
+import json
+
+import pandas as pd
+from pandas import testing as pd_testing
+
+from ml4ir.applications.ranking.model.metrics.metrics_helper import *
+
+
+class ComputeSecondaryMetricsTest(unittest.TestCase):
```
lastmansleeping, line 10: Some of these tests might seem redundant. But in some ways the point is that the existing metrics do not capture all scenarios and will be better captured with a metric that uses the actual values as opposed to binary assessments of the secondary label.

python/ml4ir/applications/ranking/model/metrics/metrics_helper.py
```diff
@@ -2,143 +2,156 @@
...
+        Name of the secondary label used to define metric name
+    prefix: str
+        Prefix attached to the metric name
+
+    Returns
+    -------
+    dict
+        Key value pairs of the metric names and the associated computed values
+        using the secondary label
+    """
```
lastmansleeping, line 50: Add example or explanation for how to use secondary labels.

lastmansleeping: or when

arvindsrikantan (approved):
Looks good to me

mohazahran (approved):
Looks good.

PR #149, opened 2021-11-23
Adding NDCG for secondary label

lastmansleeping:
The actual NDCG implementation is straight forward and the refactoring helps keep the code clean.

- [x] Add NDCG and verify everything works manually
- [x] Add docstrings
- [x] Add unittests

I will update the TODO above, but let me know your initial thoughts @mohazahran @arvindsrikantan


mohazahran (approved):
It looks good to me

PR #157, opened 2022-02-07
Update to tensorflow==2.7.0 - update model

lastmansleeping:
1. https://github.com/salesforce/ml4ir/pull/156
**2. https://github.com/salesforce/ml4ir/pull/157**
3. https://github.com/salesforce/ml4ir/pull/158
4. https://github.com/salesforce/ml4ir/pull/159

I was able to train both Ranking and EP models successfully with the following commands.
```
--run_id test \
--data_format tfrecord \
--loss_key softmax_cross_entropy \
--num_epochs 2 \
--execution_mode train_evaluate
```
and
```
python ml4ir/applications/classification/pipeline.py \
--batch_size 32 \
--run_id test \
--data_format tfrecord \
--execution_mode train_evaluate \
--num_epochs 2
```

There are still a few pending items regardless and worth doing in follow up items before merging with master.

- [ ] Fix ranking *validation* metrics. Test metrics are working fine.
- [ ] Fix tests
- [ ] [FOLLOW UP] `model.summary()` doesn't print the nested layers, this causes issues with temperature scaling and L1 which require accessing weights from nested layers.
- [ ] [FOLLOW UP] Fix TF2.7 issue with handling duplicate metric names for old vs new metrics
- [ ] [FOLLOW UP]  Verify backward compatibility with `ml4ir-inference`


python/ml4ir/base/model/architectures/dnn.py
```diff
@@ -10,56 +11,85 @@
...
+                                                                     l2_coeff=self.model_config[DNNLayerKey.POSITIONAL_BIAS_HANDLER].get("l2_coeff", 0))
 
     def define_architecture(self, model_config: dict, feature_config: FeatureConfig):
         """
         Convert the model from model_config to a List of tensorflow.keras.layer
 
         :param model_config: dict corresponding to the model config
         :param feature_config: dict corresponding to the feature config, only used in case of classification if the last
-            layer of the model_config doesn't have a units number defined (or set to -1). In which case we retrieve the
+            layer of the model_config doesn"t have a units number defined (or set to -1). In which case we retrieve the
```
arvindsrikantan, line 71: Find/replace typo

lastmansleeping: Haha, good catch.

python/ml4ir/base/model/losses/loss_base.py
```diff
@@ -1,39 +1,48 @@
...
+
+
+class RelevanceLossBase(layers.Layer):
     """
     Abstract class that defines the loss and final activation function
     used to train a RelevanceModel
     """
 
-    def get_loss_fn(self, **kwargs):
+    def call(self, inputs, y_true, y_pred, training=None):
```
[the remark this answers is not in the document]
lastmansleeping: You should be able to find examples under `ml4ir/applications/ranking/model/losses/listwise_losses.py`.

I think I have found a way to not pass the inputs anymore. I will file a follow up work item for this. But for context, this is not a new argument. Earlier, we were using `kwargs.get(...)`. I have just made it explicit now.

python/ml4ir/base/model/relevance_model.py
```diff
@@ -117,31 +117,43 @@ def __init__(
...
             metrics_impl: List[Union[str, kmetrics.Metric]] = get_metrics_impl(
                 metrics=metrics, feature_config=feature_config, metadata_features=metadata_features
             )
 
-            # Compile model
-            """
-            NOTE:
-            Related Github issue: https://github.com/tensorflow/probability/issues/519
-            """
             self.model.compile(
```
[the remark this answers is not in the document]
lastmansleeping: Nope. That's the great news about all of my effort. There are no keras symbolic tensors or input placeholders anymore. Everything is a tensorflow tensor end-to-end.

python/ml4ir/applications/classification/model/metrics/metrics_factory.py
```diff
@@ -21,9 +18,7 @@ def get_metric(metric_key: str) -> Metric:
     Metric class
         Class defining the metric computation logic
     """
-    if metric_key == MetricKey.CATEGORICAL_ACCURACY:
```
arvindsrikantan, line 24: Why was this removed?

arvindsrikantan: Never mind. I see it below

lastmansleeping: I removed unnecessary boilerplate. You can simply use `categorical_accuracy` and use tensorflow/keras native loss without overriding and wrapping.

python/ml4ir/applications/ranking/model/metrics/metrics_impl.py
```diff
@@ -67,105 +45,26 @@ def update_state(self, y_true, y_pred, sample_weight=None):
...
         `y_true` and `y_pred` should have the same shape.
         """
-        query_scores: Tensor = self._fn(y_true, y_pred, **self._fn_kwargs)
-        return super(MeanMetricWrapper, self).update_state(
-            query_scores, sample_weight=sample_weight
+        # Convert predicted ranking scores into ranks for each record per query
+        y_pred_ranks = tf.add(
+            tf.argsort(
+                tf.argsort(y_pred, axis=-1, direction="DESCENDING", stable=True), stable=True
+            ),
```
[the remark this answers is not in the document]
lastmansleeping: `tf.argsort` essentially returns a tensor with indices pointing to the original tensor. So, to assign the ranks back to the original tensor, you need to do another argsort.

I think this example should clear it up ->

```
In [61]: tf.argsort([1, 3, 2], direction="DESCENDING")
Out[61]: <tf.Tensor: shape=(3,), dtype=int32, numpy=array([1, 2, 0], dtype=int32)>

In [62]: tf.argsort(tf.argsort([1, 3, 2], direction="DESCENDING"))
Out[62]: <tf.Tensor: shape=(3,), dtype=int32, numpy=array([2, 0, 1], dtype=int32)>
```

python/ml4ir/base/tests/test_lr_schedule.py
```diff
@@ -217,7 +218,7 @@ def test_cyclic_lr_in_training_pipeline(self):
         lr_list = my_callback_object.get_lr_list()
         lr_gold = [0.001,0.020800006,0.040599994,0.0604,0.080199994,0.1,0.080199994,0.0604,0.040599994,0.020800006,0.001,
                    0.010900003,0.020800006,0.030699994,0.040599994,0.050499998,0.040599994,0.030699994,0.020800006,
-                   0.010900003,0.001,0.0059499955,0.010900003,0.015849996,0.020800006,0.02575,0.020800006,0.015849996,
+                   0.010900003,0.001,0.0059499955,0.010900003,0.015849996,0.020800006,0.02575,0.02575,0.020800006,0.015849996,
```
arvindsrikantan, line 221: Thoughts on why this changed?

lastmansleeping: This was surprising to me too. If you notice the values, it is just one single value that got changed. In the new version, there are 51 values compared to 50 previously. 

Here are a few guesses:
- There is a new implementation on tensorflow side in `tf.keras.optimizers.schedules.LearningRateSchedule`
- I do call `model(...)` once before calling `model.fit(...)`. Maybe the learning rate scheduler also gets triggered with this.

I haven't investigated further here though because the change looked very small to me. But might be good to hear from @mohazahran too as he wrote this piece originally.

arvindsrikantan (approved):
Will appreciate @mohazahran confirming the change in the LR test. Looks good to me otherwise!

PR #158, opened 2022-02-07
Update to tensorflow==2.7.0 - update requirements and build

lastmansleeping:
1. https://github.com/salesforce/ml4ir/pull/156
2. https://github.com/salesforce/ml4ir/pull/157
**3. https://github.com/salesforce/ml4ir/pull/158**
4. https://github.com/salesforce/ml4ir/pull/159

I was able to train both Ranking and EP models successfully with the following commands.
```
--run_id test \
--data_format tfrecord \
--loss_key softmax_cross_entropy \
--num_epochs 2 \
--execution_mode train_evaluate
```
and
```
python ml4ir/applications/classification/pipeline.py \
--batch_size 32 \
--run_id test \
--data_format tfrecord \
--execution_mode train_evaluate \
--num_epochs 2
```

There are still a few pending items regardless and worth doing in follow up items before merging with master.

- [ x ] Fix ranking *validation* metrics. Test metrics are working fine.
- [ x ] Fix tests
- [ x ] [FOLLOW UP] `model.summary()` doesn't print the nested layers, this causes issues with temperature scaling and L1 which require accessing weights from nested layers.
- [ x ] [FOLLOW UP] Fix TF2.7 issue with handling duplicate metric names for old vs new metrics
- [ ] [FOLLOW UP]  Verify backward compatibility with `ml4ir-inference`
- [ ] [FOLLOW UP] Fix temperature scaling
- [ ] [FOLLOW UP] Fix transfer learning


.circleci/config.yml
```diff
@@ -148,6 +152,7 @@ jobs:
             python3 -m venv venv
             . venv/bin/activate
             cd python
+            pip install --upgrade pip
```
arvindsrikantan, line 155: Why was this necessary?

arvindsrikantan (approved):
Looks good!

.circleci/config.yml
lastmansleeping: The base pip version on the circleci docker image wasn't able to install a package. I forget which one.

python/ml4ir/applications/classification/tests/test_classification_model.py
```diff
@@ -42,7 +42,7 @@ def test_group_metrics_df(self):
         Test the dimensions of the grouped metrics
         """
         # Metrics cardinality and names
-        metrics = self.classification_pipeline.metrics_keys  # Metrics during training
+        metrics = ["loss"] + self.classification_pipeline.metrics_keys  # Metrics during training
```
arvindsrikantan, line 45: Did you get a chance to train classification models with these changes @lastmansleeping ? Should we have a follow up item to ensure end-to-end flow works as expected?

python/ml4ir/applications/ranking/tests/test_losses.py
```diff
@@ -28,38 +28,49 @@ def setUp(self):
...
+
+        y_pred = loss_op.final_activation_op({
+            "logits": self.logits,
+            "metadata": {
+                "mask": self.mask
+            }
+        })
         assert np.isclose(y_pred[0][0].numpy(), 0.54905695, atol=1e-5)
-        assert np.isclose(y_pred[2][4].numpy(), 0.64832306, atol=1e-5)
+        assert np.isclose(y_pred[2][4].numpy(), 0., atol=1e-5)
```
arvindsrikantan, line 40: Why did this change?

lastmansleeping: I think this should be expected to change really because we've modified underlying tensorflow, numpy versions which could be impacting the random seeds or even the implementations. This is my guess though.

python/ml4ir/applications/classification/tests/test_classification_model.py
lastmansleeping: I have verified classification models too. You should be able to verify with
```
python ml4ir/applications/classification/pipeline.py \
--batch_size 32 \
--run_id test \
--data_format tfrecord \
--execution_mode train_evaluate \
--num_epochs 2
```

Please note that the last step of this will fail with the default model_config because temperature scaling is not yet supported.

All other classification, ranking and base tests are passing though.

python/ml4ir/applications/ranking/tests/test_losses.py
lastmansleeping: (Ignore my previous reply)

This is 0., and should be 0., because we mask the predictions too. We weren't doing this earlier. It wasn't a breaking issue though because it was being masked in the loss. I'm doing the masking in the activation now to make sure in the next version update, we can remove passing input features to the loss function like we discussed.

lastmansleeping: If you notice, the final loss computed is still the same in the next line.

arvindsrikantan (approved):
LGTM!

PR #162, opened 2022-02-16
[DEPRECATED] Adding setrank architecture

lastmansleeping:
This PR has been deprecated. Here's the new PR with SetRank written using `tf-models-official` -> https://github.com/salesforce/ml4ir/pull/206


python/ml4ir/base/model/architectures/transformer.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def point_wise_feed_forward_network(model_dim, feed_forward_dim):
+    return tf.keras.Sequential([
+        tf.keras.layers.Dense(feed_forward_dim, activation='relu'),  # (batch_size, seq_len, feed_forward_dim)
+        tf.keras.layers.Dense(model_dim)  # (batch_size, seq_len, model_dim)
+    ])
+
+
+class MultiHeadAttention(tf.keras.layers.Layer):
```
ArmandGiraud, line 62: Why we don't use a [standard transformer layer](https://www.tensorflow.org/api_docs/python/tfm/nlp/layers/TransformerEncoderBlock) here?

python/ml4ir/applications/ranking/model/metrics/metrics_helper.py
```diff
@@ -5,12 +5,12 @@
 
 # Metrics where higher value is better/desirable
 POSITIVE_METRIC_SUFFIXES = [
-    "MRR",
-    "NDCG_mean"
+    "MRR"
```
ArmandGiraud, line 8: I'm not sure to understand what the changes here have to do with the PR, but 1 - NDCG, could be named inverse NDCG , or reciprocal dcg, i'm not sure intrinsic_failure is very indicative

python/ml4ir/applications/classification/model/classification_model.py
```diff
@@ -245,7 +245,9 @@ def predict(
         predictions_df[self.output_name] = [x for x in predictions_]
         if logs_dir:
             np.set_printoptions(formatter={'all':lambda x: str(x.decode('utf-8')) if isinstance(x, bytes) else str(x)},
-                                linewidth=sys.maxsize, threshold=sys.maxsize)  # write the full vector in the csv not ...
+                                linewidth=sys.maxsize,
+                                threshold=sys.maxsize,  # write the full vector in the csv not ...
+                                legacy="1.13")  # enables 1.13 legacy printing mode
```
ArmandGiraud, line 250: Why is legacy printing necessary here?

python/ml4ir/applications/ranking/model/metrics/metrics_helper.py
lastmansleeping: I will remove these changes. I do not remember why they are here either.

python/ml4ir/base/model/architectures/transformer.py
lastmansleeping: Let me look into this. I believe we didn't have it back when I wrote this.

PR #166, opened 2022-03-09
Adding support for flexible dense model

arvindsrikantan:
Some TODOs:
- [x] Adding tests
- [x] Example config

In a subsequent PR (https://github.com/salesforce/ml4ir/pull/168):
- [x] Ability to tie weights to reuse layers
- [x] Visualizing the dependency graph


python/ml4ir/base/config/keys.py
```diff
@@ -21,6 +21,7 @@ class ArchitectureKey(Key):
     DNN = "dnn"
     LINEAR = "linear"
     RNN = "rnn"
+    DENSE_MODEL = "dense-model"
```
lastmansleeping, line 24: @arvindsrikantan Let's rename this. `dense-model` is quite meaningless to what you are really doing. Maybe `dnn-dag` or `dense-dag`?

lastmansleeping: I still like `hierarchical-dnn` too.

python/ml4ir/base/model/architectures/flexible_dense_model.py
```diff
@@ -0,0 +1,253 @@
+from __future__ import annotations
```
lastmansleeping, line 1: @arvindsrikantan Worth renaming this module name too. `flexible_dense_model` is in some ways all encompassing.

arvindsrikantan: That is the idea though, haha

arvindsrikantan:
Example config:

```yaml
architecture_key: auto-dag-network
layers:
  - type: keras.layers.merge.Concatenate
    name: features_concat
    inputs:
      - query_start
      - sim_score
      - pv_s
    aslist: true
    axis: -1
  - type: keras.layers.core.dense.Dense
    name: first_dense
    inputs:
      - features_concat
    units: 512
    activation: relu
  - type: keras.layers.core.dense.Dense
    name: second_dense
    inputs:
      - first_dense
    units: 256
    activation: relu
  - type: keras.layers.core.dense.Dense
    name: final_dense
    inputs:
      - second_dense
    units: 1
    activation: null
optimizer:
  key: sgd
  gradient_clip_value: 0.25
lr_schedule:
  key: reduce_lr_on_plateau
  learning_rate: 0.1
  factor: 0.5
  patience: 2
  min_lr: 0.0001
  mode: max
```

python/ml4ir/base/data/tfrecord_writer.py
```diff
@@ -42,19 +42,20 @@
 ... --keep-single-files
```
arvindsrikantan, line 42: Ignore this file (part of another PR)

python/ml4ir/base/model/architectures/dnn.py
```diff
@@ -122,6 +122,7 @@ def call(self, inputs, training=None):
             Logits tensor computed with the forward pass of the architecture layer
         """
         train_features = inputs[FeatureTypeKey.TRAIN]
+        metadata_features = inputs[FeatureTypeKey.METADATA]
```
arvindsrikantan, line 125: This was a bug in the previous update. Let me know if this looks right @lastmansleeping

python/ml4ir/applications/ranking/tests/config/auto_dag_network_sample_config.yaml
```diff
@@ -0,0 +1,38 @@
+architecture_key: auto-dag-network
+layers:
+  - type: keras.layers.merge.Concatenate
+    name: features_concat
+    inputs:
+      - query_text
+      - text_match_score
+      - page_views_score
+    aslist: true
+    axis: -1
```
lastmansleeping, line 10: @arvindsrikantan Worth putting the keras layers args inside a nested struct here. Maybe `layer_kwargs`?

python/ml4ir/applications/ranking/tests/test_base.py
```diff
@@ -86,7 +87,7 @@ def get_ranking_model(
 
         # Define optimizer
         optimizer: Optimizer = get_optimizer(
-            model_config=self.file_io.read_yaml(self.args.model_config),
+            model_config=self.model_config,
```
lastmansleeping, line 90: @arvindsrikantan Thank you!

python/ml4ir/applications/ranking/tests/test_serving.py
```diff
@@ -119,14 +125,28 @@ def _filter_records(x, mask):
         )
 
         # Compare the scores from the different versions of the model
-        assert np.isclose(model_predictions, default_signature_predictions, rtol=0.01,).all()
+        assert np.isclose(model_predictions, default_signature_predictions, rtol=0.01, ).all()
```
lastmansleeping, line 128: [Minor] @arvindsrikantan Looks like dangling `,`

python/ml4ir/applications/ranking/tests/config/auto_dag_network_sample_config.yaml
arvindsrikantan: I'll stick to `args` instead of `layer_kwargs` to be consistent with feature layer.

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -0,0 +1,355 @@
...
+from ml4ir.base.config.keys import FeatureTypeKey
+from ml4ir.base.features.feature_config import FeatureConfig
+from ml4ir.base.io.file_io import FileIO
+
+
+def get_layer_subclasses() -> Dict[str, Type[layers.Layer]]:
+    """Get mapping of {subclass-name: subclass} for all derivative classes of keras.layers.Layer"""
+
+    def full_class_name(cls):
+        """Get {package_name}.{class_name}"""
```
lastmansleeping, line 18: @arvindsrikantan Can you add an example for keras native classes and custom ml4ir classes here?

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -0,0 +1,355 @@
...
+        if module == "builtins":
+            return cls.__qualname__  # avoid outputs like 'builtins.str'
+        return f"{module}.{cls.__qualname__}"
+
+    def get_sub_classes(cls):
+        """DFS traversal to get all subclasses of `cls` class"""
+        for subcls in cls.__subclasses__():
+            if subcls not in subclasses:
+                # Handles duplicates
+                subclasses[full_class_name(subcls)] = subcls
```
lastmansleeping, line 29: @arvindsrikantan Updating a global variable in a recursive loop is probably not the best coding practice, but it should be fine here as this is not really reusable code.

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -0,0 +1,355 @@
...
+        inputs: list of str
+            All inputs to the `layer` to be passed during forward prop
+        inputs_as_list: bool
+            Boolean to indicate if the set of input tensors to the layer should be converted to list
+        """
+        self.name = name
+        self.layer = layer
+        self.dependent_children = dependent_children if dependent_children is not None else []
+        self.inputs = inputs
+        self.is_input_node = True if inputs is None else False
```
lastmansleeping, line 73: [Minor] @arvindsrikantan `is_input_node` is probably confusing terminology given that you also have `inputs` in this same class. With that said, I don't have ideas for a better name.

lastmansleeping: But my point is "input" means two different things in this class.

python/ml4ir/base/model/architectures/auto_dag_network.py
arvindsrikantan: Agreed. Passing by reference is an option, but not necessary unless we want to reuse

arvindsrikantan: It's not global will be my excuse :P

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -0,0 +1,355 @@
...
+        for node in self.nodes.values():
+            if node.name not in visited:
+                path.add(node.name)
+                dfs(node)
+                path.remove(node.name)
+
+        return order[::-1]
+
+
+class AutoDagNetwork(keras.Model):
```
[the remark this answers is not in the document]
arvindsrikantan: Not at this point I'm afraid. I'm open to suggestions for follow-up work

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -0,0 +1,362 @@
...
+            All inputs to the model from the interaction model
+        """
+        self.nodes = self.create_nodes(layer_ops, inputs)
+        self.create_dependency_graph()
+        # TODO: Add graph visualization for easier debugging
+        output_nodes = self.get_output_nodes()
+        if len(output_nodes) == 0:
+            raise CycleFoundException("No output nodes found because of cycle in DAG")
+        elif len(output_nodes) > 1:
+            raise NotImplementedError(f"1 output node expected, found {len(output_nodes)}: {output_nodes}")
```
lastmansleeping, line 125: @arvindsrikantan You could also create a `DanglingNodeFound` exception while you are at it.

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -0,0 +1,362 @@
...
+        Parameters
+        ----------
+        layer_ops: dict
+            Dictionary which contains information for all the layer nodes in the graph. Expected schema:
+            {
+                "NodeName":  # Name of the layer node
+                {
+                    "inputs": ["InputNodeName1", "InputNodeName2"],  # Name of the input layer nodes for this layer
+                    "op": keras.layers.Layer.ChildClass(),  # Layer instance which defines the operation of the node
+                    "aslist": True  # If inputs to the "op" should be a list of tensors
```
lastmansleeping, line 112: @arvindsrikantan Update docs with `args`

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -0,0 +1,362 @@
...
+        Returns
+        -------
+        keras.layers.Layer
+            Instance of current_layer_type
+        """
+        try:
+            return self.available_layers[current_layer_type](**current_layer_args)
+        except KeyError:
+            self.file_io.logger.error("Layer type: '%s' is not supported or not found in %s",
+                                      current_layer_type, str(self.available_layers))
```
lastmansleeping, line 273: @arvindsrikantan Is `available_layers` going to print every single layer under `tf.keras`?

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -0,0 +1,362 @@
...
+
+        Returns
+        -------
+        dict
+            Dictionary with the layer instance, inputs required for the layer and the format of inputs to the layer
+        """
+        return {
+            self.INPUTS: layer_args[self.INPUTS],
+            self.OP_IDENTIFIER: self.instantiate_op(layer_args[self.LAYER_TYPE], layer_args.get(self.LAYER_KWARGS, {})),
+            self.INPUTS_AS_LIST: layer_args.get(self.INPUTS_AS_LIST, False)
```
lastmansleeping, line 294: @arvindsrikantan I'm still not very clear on why we need `as_list`.

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -0,0 +1,362 @@
...
+
+        Returns
+        -------
+        LayerGraph
+            Dependency DAG for the given model config
+        """
+
+        layer_ops = {layer_args[self.NAME]: self.get_layer_op(layer_args) for layer_args in
+                     model_config[self.LAYERS_IDENTIFIER]}
+        # While sorting is not mandatory, I would advise not removing it for the sake of reproducibility
```
lastmansleeping, line 314: @arvindsrikantan "I would advise" ~-> "it is recommended"

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -0,0 +1,362 @@
...
+        inputs = sorted(set([input_name for layer_op in layer_ops.values()
```
lastmansleeping, line 315: [Minor] @arvindsrikantan Why not sort in `instantiate_op` itself? Seems slightly cleaner.

lastmansleeping: With that, I think you can combine the two methods as well. I don't see why we need a separate method for `get_layer_op` which is essentially a wrapper around `instantiate_op`

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -0,0 +1,362 @@
...
+            File input output handler
+        """
+        super().__init__(**kwargs)
+
+        self.file_io: FileIO = file_io
+        self.model_config = model_config
+        self.feature_config = feature_config
+
+        # Get all available layers (including keras layers)
+        self.available_layers = get_layer_subclasses()
```
lastmansleeping, line 241: @arvindsrikantan Worth making this an argument to the constructor with the default being `get_layer_subclasses()`

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -0,0 +1,362 @@
...
+        model_graph = self.define_architecture(model_config)
```
lastmansleeping, line 242: @arvindsrikantan Why not `self.model_graph = ...`? Maybe you don't need to register layers if you do that?

lastmansleeping: (Not sure)

But I feel you should store the actual graph itself in the object's state.

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -0,0 +1,362 @@
...
+            # Input nodes don't need any execution
+            if node.is_input_node:
+                if node.name not in train_features:
+                    raise KeyError(f"Input feature {node.name} cannot be found in the feature ops outputs")
+            else:
+                layer_input = {k: outputs[k] for k in node.inputs}
+                if node.inputs_as_list or len(layer_input) == 1:
+                    layer_input = list(layer_input.values())
+                    if len(layer_input) == 1:
+                        layer_input = layer_input[0]
```
lastmansleeping, line 351: @arvindsrikantan Why would you need to do this? For `concat`?

python/ml4ir/base/tests/test_auto_dag_network.py
```diff
@@ -0,0 +1,117 @@
...
+        self.layer_subclasses = get_layer_subclasses()
+
+    def test_local_userdefined_subclass(self):
+        """Test to show local classes are also accessible but need to be in scope to be found"""
+
+        class UserDefinedTestLayerLocal(Layer):
+            pass
+
+        self.assertNotIn(UserDefinedTestLayerLocal, self.layer_subclasses.values())
+        self.assertIn(UserDefinedTestLayerLocal, get_layer_subclasses().values())
```
lastmansleeping, line 25: @arvindsrikantan Worth adding a test for the string to be used in model config to access this.

python/ml4ir/base/tests/test_auto_dag_network.py
```diff
@@ -0,0 +1,117 @@
...
+
+from ml4ir.base.model.architectures.auto_dag_network import LayerNode, get_layer_subclasses, CycleFoundException, \
+    LayerGraph
+
+
+class UserDefinedTestLayerGlobal(Layer):
+    pass
+
+
+class TestGetLayerSubclasses(unittest.TestCase):
```
lastmansleeping, line 13: @arvindsrikantan Can also add a test for a keras native layer to be "complete"

lastmansleeping (commented):
@arvindsrikantan Left a few comments. Nothing major, looks clean to me. Maybe we should add unittest for AutoDagNetwork as well?

python/ml4ir/base/model/architectures/auto_dag_network.py
arvindsrikantan: There are no args here

python/ml4ir/base/model/architectures/auto_dag_network.py
arvindsrikantan: Dangling nodes will be considered input nodes. Also, we cannot create a non input node without inputs. The idea here is that we might want to eventually support multiple outputs, so I feel NotImplementedError is more appropriate.

python/ml4ir/base/model/architectures/auto_dag_network.py
arvindsrikantan: `self.model_graph = ...` doesn't register layers. I've made this change in #168

python/ml4ir/base/model/architectures/auto_dag_network.py
arvindsrikantan: Yes, not only tf.keras, but all children classes in the context (example layers from tensorflow_probability are also shown)

python/ml4ir/base/model/architectures/auto_dag_network.py
arvindsrikantan: Layers can take 3 forms of inputs -> tensor, dict, list. If there is only 1 input to a node, we automatically choose to only send the tensor. If the layer accepts multiple inputs, this flag controls what the format of the input should be (dict being the default). Let me know if I'm missing something

python/ml4ir/base/model/architectures/auto_dag_network.py
arvindsrikantan: Good point on combining the two. It was an artifact of "levels" which is not needed anymore

arvindsrikantan: > Why not sort in get_layer_op itself? Seems slightly cleaner.

I'm not sorting the inputs to each node here. I'm sorting just the feature layer nodes. The final order is still controlled by the model_config

python/ml4ir/base/model/architectures/auto_dag_network.py
arvindsrikantan: Not just for concat. This will be needed by all layers which have only a single tensor input or layers like concat which need list of inputs (e.g.: BERT - two segments)

python/ml4ir/base/tests/test_auto_dag_network.py
arvindsrikantan: @lastmansleeping , didn't quite get you. What do you mean by complete? Have all methods defined?

arvindsrikantan: Added keras native layers

python/ml4ir/base/model/architectures/auto_dag_network.py
lastmansleeping: ping

python/ml4ir/base/model/architectures/auto_dag_network.py
lastmansleeping: Yeah, I realized that later.

python/ml4ir/base/model/architectures/auto_dag_network.py
lastmansleeping: > Dangling nodes will be considered input nodes

I'm not sure I understand this.

python/ml4ir/base/model/architectures/auto_dag_network.py
lastmansleeping: > I'm not sorting the inputs to each node here. I'm sorting just the feature layer nodes.

How are they different?

lastmansleeping: Or rather, why sort then?

lastmansleeping (approved):
LGTM, but left a few comments for further clarification.

python/ml4ir/base/model/architectures/auto_dag_network.py
arvindsrikantan: To keep the order of self.nodes consistent across runs. This also ensures the topological sort result doesn't change every time. Note that there is no unique ordering for topological sort.

python/ml4ir/base/model/architectures/auto_dag_network.py
arvindsrikantan: All nodes which are dangling will have two properties:
1. They have no inputs (Feature layer nodes) -> In this case, these nodes have is_input_node=True. These are not dangling nodes
2. They have no dependent children (output nodes) -> In this case, we throw NotImplementedError because we want to eventually support multiple outputs for the model

If in case we have a non-feature layer node with 0 inputs and 0 dependent children, it will be ignored similar to a feature node which is not used by any layer which should be okay I feel

lastmansleeping (approved):

PR #167, opened 2022-03-16
Support strings as interaction model output

arvindsrikantan:
This is useful when using layers in model config which deal with strings and not just numeric types


lastmansleeping (approved):
@arvindsrikantan Can you add a note in the comment?

PR #168, opened 2022-03-18
Adding visualization and ability to tie weights

arvindsrikantan:
Follow up to https://github.com/salesforce/ml4ir/pull/166


python/ml4ir/applications/ranking/features/feature_config.py
```diff
@@ -1,3 +1,4 @@
+# TODO: Fix or remove this file (no usages found)
```
lastmansleeping, line 1: @arvindsrikantan Delete it.

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -81,12 +84,8 @@ def __init__(
...
 
     def __str__(self):
-        """Node is identified by its name"""
-        return self.name
-
-    def __repr__(self):
-        """Get the obj representation"""
-        return self.name
+        """Get a readable representation of the node"""
+        return f"{self.name}" if self.is_input_node else f"{self.name}\n[{self.layer.__class__.__qualname__}]"
```
lastmansleeping, line 88: @arvindsrikantan Could add input node names too.

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -239,15 +274,24 @@ def __init__(
 
         # Get all available layers (including keras layers)
         self.available_layers = get_layer_subclasses()
-        model_graph = self.define_architecture(model_config)
-        self.execution_order: List[LayerNode] = model_graph.topological_sort()
+        self.model_graph = self.define_architecture(model_config)
+
+        # Get the first file handler for the logger
+        file_handlers = [handler for handler in self.file_io.logger.handlers if isinstance(handler, FileHandler)]
+        logging_dir = Path(file_handlers[0].baseFilename).parent if file_handlers else Path(self.DEFAULT_VIZ_SAVE_PATH)
```
lastmansleeping, line 281: @arvindsrikantan This seems unnecessarily complicated. I recommend you move it to `RelevanceScorer` perhaps or even `RelevanceModel`?

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -274,23 +318,33 @@ def instantiate_op(self, current_layer_type: str, current_layer_args: Dict) -> l
...
             All the arguments from model config which are needed to define a layer op
+        existing_op: instance of type keras.layers.Layer
+            If specified, reuse the layer op
 
         Returns
         -------
         dict
             Dictionary with the layer instance, inputs required for the layer and the format of inputs to the layer
         """
+        if existing_op and existing_op.__class__.__name__ != layer_args[self.LAYER_TYPE].split(".")[-1]:
```
lastmansleeping, line 341: @arvindsrikantan Add a comment on the check being made here. I find it very hard to understand how you conclude "incompatible" here.

lastmansleeping (changes requested):
Looks fine to me, but make sure you add/update the serving test as well to make sure this works end-to-end.

python/ml4ir/applications/ranking/features/feature_config.py
arvindsrikantan: The file you mean?

python/ml4ir/base/model/architectures/auto_dag_network.py
arvindsrikantan: That makes the visualization very clunky.

python/ml4ir/base/model/architectures/auto_dag_network.py
arvindsrikantan: Move the entire visualization code to those classes you mean? It's very specific to AutoDagNework though

lastmansleeping (changes requested):
@arvindsrikantan Considering you had to install non `pip` packages to add this, I think you should add `pygraphviz` as an optional dependency. Check here -> https://github.com/salesforce/ml4ir/blob/master/python/optional_requirements.yaml

(Haven't reviewed the rest again yet)

python/ml4ir/applications/ranking/tests/data/configs/auto_dag_network_sample_config.yaml
```diff
@@ -39,4 +55,6 @@ lr_schedule:
   factor: 0.5
   patience: 2
   min_lr: 0.0001
-  mode: max
\ No newline at end of file
+  mode: max
+tie_weights:
+  - ["sample_tie_weight_1", "sample_tie_weight_2"]
```
lastmansleeping, line 60: [Minor] @arvindsrikantan You don't need to have `"` here.

python/ml4ir/base/tests/test_auto_dag_network.py
```diff
@@ -123,4 +128,218 @@ def test_simple_topological(self):
...
+            self.assertEqual(model.model_graph.get_node("global1").layer, model.model_graph.get_node("global2").layer)
+        with self.subTest("Inputs should be different"):
+            self.assertNotEqual(model.model_graph.get_node("global1").inputs,
+                                model.model_graph.get_node("global2").inputs)
+
+    def test_without_tie_weights(self):
+        model_config = copy.deepcopy(self.model_config)
+        model_config.pop("tie_weights")
+        model = AutoDagNetwork(model_config=model_config, feature_config=self.feat_config, file_io=self.file_io)
+        self.assertIsNotNone(model.model_graph)
```
lastmansleeping, line 328: @arvindsrikantan This feels like an insufficient test to me. Can we check if the actual connections match with the model config or something similar to tests for the graph?

python/ml4ir/build/Dockerfile
```diff
@@ -9,6 +9,8 @@ ENV PYTHONPATH=/home/ml4ir/python
 
 # Install python dependencies
 COPY requirements.txt /home/ml4ir/python/
+RUN sudo apt-get --allow-releaseinfo-change update
```
lastmansleeping, line 12: @arvindsrikantan Can you add a note here?

python/optional_requirements.yaml
```diff
@@ -21,6 +21,9 @@
 # To install ml4ir `all`, run pip install ml4ir[all]
 all:
   - pyspark==3.0.1  # required to run ml4ir.base.pipeline
+  - pygraphviz==1.7  # required to visualize ml4ir.base.model.architectures.auto_dag_network.LayerGraph
 pyspark:
   - pyspark==3.0.1  # required to support pyspark data read
+visualization:
```
lastmansleeping, line 27: @arvindsrikantan Thanks!

lastmansleeping (changes requested):
@arvindsrikantan Looks good to me. Left a few comments. The only other check I would request is to create a dummy PR on `master` to verify if the dockerfile updates work.

.circleci/config.yml
```diff
@@ -13,6 +13,8 @@ jobs:
       - run:
           name: Install Python deps in a venv
           command: |
+            sudo apt-get --allow-releaseinfo-change update
```
lastmansleeping, line 16: @arvindsrikantan Why did you give up on adding this to the circleCI Dockerfile?

lastmansleeping:
@arvindsrikantan Maybe I am missing something but I don't see that the review comments are addressed. Specifically, I am looking for these
- getting the path from the `logger` seems hacky to me. Did you try a workaround with saving this in `RelevanceModel` or `RelevanceScorer`?
- create a dummy PR on master to verify if the dockerfile updates work.

arvindsrikantan:
> @arvindsrikantan Maybe I am missing something but I don't see that the review comments are addressed.

Missed your previous review notification (didn't keep an eye on my personal email)

.circleci/config.yml
arvindsrikantan: We're using the base circleci image everywhere, so I chose to add it directly

python/ml4ir/applications/ranking/features/feature_config.py
lastmansleeping: Yep.

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -1,7 +1,14 @@
 from __future__ import annotations
 
+from logging import FileHandler
+from pathlib import Path
 from typing import List, Dict, Union, Type
 
+try:
+    import pygraphviz as pgv
+except ImportError:
```
lastmansleeping, line 9: @arvindsrikantan You could also mock the import?

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -201,6 +222,23 @@ def dfs(node: LayerNode):
 
         return order[::-1]
 
+    def visualize(self, path: str):
+        """
+        Utility function to visualize the DAG
```
lastmansleeping, line 227: @arvindsrikantan Worth mentioning "save to disk" as well.

python/build-requirements.txt
```diff
@@ -29,4 +29,7 @@ testfixtures==6.17.1
 pyspark==3.0.1
 
 #calibration
-tensorflow-probability==0.15.0
\ No newline at end of file
+tensorflow-probability==0.15.0
+
+# DAG visualization
+pygraphviz==1.7
```
lastmansleeping, line 35: @arvindsrikantan Do we need this in the `build-requirements.txt`? I forget, but I think this is only used for tests.

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -239,15 +280,29 @@ def __init__(
...
         self.available_layers = get_layer_subclasses()
-        model_graph = self.define_architecture(model_config)
-        self.execution_order: List[LayerNode] = model_graph.topological_sort()
+
+        # Visualize with pygraphviz if available
+        self.viz_path = None
+        if pgv:
+            # Get the first file handler for the logger
+            file_handlers = [handler for handler in self.file_io.logger.handlers if isinstance(handler, FileHandler)]
+            logging_dir = Path(file_handlers[0].baseFilename).parent if file_handlers else Path(self.DEFAULT_VIZ_SAVE_PATH)
```
lastmansleeping, line 289: @arvindsrikantan I remember leaving a comment about not fetching the folder from the logger. Why can't we move this logging higher up the call to `RelevanceScorer` perhaps?

lastmansleeping: You could have a `draw(...)` method for both `DNN` and `AutoDagNetwork` and leave it unimplemented for DNN

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -274,6 +329,37 @@ def instantiate_op(self, current_layer_type: str, current_layer_args: Dict) -> l
...
+        dict
+            Dictionary with the layer instance, inputs required for the layer and the format of inputs to the layer
+        """
+        # Ensure the class names for the current layer and the existing op are the same
+        if existing_op and existing_op.__class__.__name__ != layer_args[self.LAYER_TYPE].split(".")[-1]:
+            raise TypeError(f"Cannot reuse existing layer of type {existing_op.__class__.__name__}. "
+                            f"{layer_args[self.LAYER_TYPE]} is incompatible")
+        return {
+            self.INPUTS: layer_args[self.INPUTS],
+            self.OP_IDENTIFIER: existing_op if existing_op
```
lastmansleeping, line 358: @arvindsrikantan So tying weights is just a matter of using the same layer? I assumed there would be two different layers sharing the tensorflow variables somehow. Glad that this seems to be much more straightforward.

lastmansleeping: This feels very RNN backprop like. Correct me if my understanding is wrong.

python/ml4ir/base/tests/test_auto_dag_network.py
lastmansleeping: Thanks!

lastmansleeping (commented):
Left a few comments, but looks good overall. :)

python/ml4ir/base/model/architectures/auto_dag_network.py
arvindsrikantan: Yeah, pretty much like an RNN. In this case it boils down to using the same layer. We currently don't support word2vec type weight tying

python/build-requirements.txt
arvindsrikantan: The circle CI build will need pygraphviz too for running the test (from what I understand)

python/ml4ir/base/model/architectures/auto_dag_network.py
arvindsrikantan: Were you suggesting to mock here or in the tests? This is not done just for testing. In case the optional dependency is not installed, it has to be ignored.

python/ml4ir/base/model/architectures/auto_dag_network.py
arvindsrikantan: Comment got lost in the refactoring I guess. IIRC, I was talking about how the visualization was specific to this architecture type, but I now see what you meant

python/ml4ir/base/model/architectures/auto_dag_network.py
lastmansleeping: You could mock here too. I don't think that is a better/worse solution than what you have though.

python/build-requirements.txt
lastmansleeping: None of the tests actually use `pygrahviz`, right?

python/ml4ir/base/model/architectures/auto_dag_network.py
lastmansleeping: Okay, let me know if you are addressing this. Otherwise I can approve this PR. Everything else seems to be handled to me.

python/ml4ir/applications/classification/pipeline.py
```diff
@@ -84,6 +84,7 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
             output_name=self.args.output_name,
             logger=self.logger,
             file_io=self.file_io,
+            logs_dir=self.args.logs_dir
```
lastmansleeping, line 87: @arvindsrikantan You should pass `logs_dir_local` here too, right?

python/ml4ir/base/model/architectures/auto_dag_network.py
```diff
@@ -304,6 +292,15 @@ def __init__(
         self.file_io.logger.info("Execution order: %s", self.execution_order)
         self.output_node = self.model_graph.output_node
 
+    def plot_abstract_model(self, plot_dir: str):
```
lastmansleeping, line 292: @arvindsrikantan Why "abstract" or rather why not just `plot_model`?

python/ml4ir/base/model/scoring/scoring_model.py
```diff
@@ -60,6 +61,8 @@ def __init__(
             Name of the output that captures the score computed by the model
         logger : Logger, optional
             Logging handler
+        logs_dir : str, optional
+            Path to the logging directory
```
lastmansleeping, line 65: @arvindsrikantan Can you also add why it is needed.

python/ml4ir/base/pipeline.py
```diff
@@ -61,26 +61,19 @@ def __init__(self, args: Namespace):
...
         self.local_io = LocalIO()
-        self.models_dir_hdfs = None
-        self.logs_dir_hdfs = None
-        self.data_dir_hdfs = None
-        if self.args.file_handler == FileHandlerKey.SPARK:
-            self.models_dir = os.path.join(self.args.models_dir, self.run_id)
-            self.logs_dir = os.path.join(self.args.logs_dir, self.run_id)
-            self.data_dir = self.args.data_dir
 
+        self.models_dir = self.models_dir_local = os.path.join(self.args.models_dir, self.run_id)
```
lastmansleeping, line 65: @arvindsrikantan Thanks!

python/ml4ir/base/model/architectures/auto_dag_network.py
lastmansleeping: Oh I didn't know this existed. Cool.

python/ml4ir/applications/classification/pipeline.py
arvindsrikantan: My bad, it has to be `logs_dir_local`

lastmansleeping (approved):
Thanks for being patient and working on this with a lot of back and forth! @arvindsrikantan :)

PR #173, opened 2022-04-25
Update tfrecord_reader.py

mohazahran:
Uncommenting this line to skip bad queries without throwing an error.
.apply(data.experimental.ignore_errors())

### DIFF
--- a/python/ml4ir/base/data/tfrecord_reader.py
+++ b/python/ml4ir/base/data/tfrecord_reader.py
@@ -797,7 +797,7 @@ def read(
         # Parallel calls set to AUTOTUNE: improved training performance by 40% with a classification model
         dataset = (
             dataset.map(parse_fn, num_parallel_calls=tf.data.experimental.AUTOTUNE)
-            # .apply(data.experimental.ignore_errors())
+             .apply(data.experimental.ignore_errors())
         )
 
     # Create BatchedDataSet


python/ml4ir/base/data/tfrecord_reader.py
```diff
@@ -797,7 +797,7 @@ def read(
         # Parallel calls set to AUTOTUNE: improved training performance by 40% with a classification model
         dataset = (
             dataset.map(parse_fn, num_parallel_calls=tf.data.experimental.AUTOTUNE)
-            # .apply(data.experimental.ignore_errors())
+             .apply(data.experimental.ignore_errors())
```
balikasg, line 800: Can you share some context on why this was commented and why we are now uncommenting? 

LGTM though. 

A separate issue: 
- On the creation of the dataset here, are we applying any randomization? In other words does the model see the data always in the same order or not? I am looking at this [for example](https://www.tensorflow.org/api_docs/python/tf/data/Dataset#shuffle)

balikasg (approved):
LGTM; please consider adding some description on the PR to better stream-line this. 
Also, I am wondering on shuffling, what do you think?

python/ml4ir/base/data/tfrecord_reader.py
mohazahran: I'm uncommenting this line to skip bad queries without throwing an error.

Offline, I believe the model see the data in the same order. However, running on Flowsnake, I believe there is some "un-planned" shuffling that happens perhaps due to spark.

mohazahran: FWIW, I think shuffling the data per epoch would produce a better model.

balikasg: Do you want to add it?

arvindsrikantan: +1 on shuffling

PR #177, opened 2022-05-04
Fixing evaluation error when no clicked document is present

arvindsrikantan:


python/ml4ir/applications/ranking/model/metrics/metrics_helper.py
```diff
@@ -202,7 +207,9 @@ def compute_secondary_labels_metrics_on_query_group(
                 compute_secondary_label_metrics(
                     secondary_label_values=query_group[secondary_label],
                     ranks=query_group[old_rank_col],
-                    click_rank=query_group[query_group[label_col] == 1][old_rank_col].values[0],
+                    click_rank=query_group[query_group[label_col] == 1][old_rank_col].values[0]
+                    if (query_group[label_col] == 1).sum() != 0
+                    else float("inf"),
```
mohazahran, line 212: `inf` would throw an exception down the path in `compute_secondary_label_metrics` which would be caught in try, catch. I don't see a way around that currently.

python/ml4ir/applications/ranking/model/metrics/metrics_helper.py
```diff
@@ -224,13 +233,13 @@ def compute_secondary_labels_metrics_on_query_group(
...
 
 def get_grouped_stats(
-    df: pd.DataFrame,
-    query_key_col: str,
-    label_col: str,
-    old_rank_col: str,
-    new_rank_col: str,
-    group_keys: List[str] = [],
-    secondary_labels: List[str] = [],
+        df: pd.DataFrame,
```
mohazahran, line 236: I notice a lot of indentations added. Is it auto formatting?

mohazahran (approved):
Looks good.

python/ml4ir/applications/ranking/model/metrics/metrics_helper.py
arvindsrikantan: Yes, auto formatting indeed

PR #184, opened 2022-07-06
Adding RankMatchFailure metric

arvindsrikantan:


arvindsrikantan:
@mohazahran , I added the tests too

python/ml4ir/applications/ranking/model/losses/listwise_losses.py
```diff
@@ -36,18 +36,24 @@ def _loss_fn(y_true, y_pred):
             mask : [batch_size, num_classes]
```
arvindsrikantan, line 36: Only auto formatting

python/ml4ir/applications/ranking/model/metrics/metrics_impl.py
```diff
@@ -297,3 +302,257 @@ def _get_matches_hook(self, y_pred_click_ranks):
...
+        `y_pred` and `y_true` should be passed in as vectors of probabilities, rather
+        than as labels. If necessary, use `tf.one_hot` to expand `y_true` as a vector.
+        If `sample_weight` is `None`, weights default to 1.
+        Use `sample_weight` of 0 to mask values.
+        """
+        name = "{}_{}".format(state, name)
+        # TODO: Handle Example dataset without mask and rank fields
+        rank = metadata_features[feature_config.get_rank("node_name")]
+        mask = metadata_features[feature_config.get_mask("node_name")]
+        if not feature_config.aux_label:
```
arvindsrikantan, line 348: Dependency on the aux_label can be relaxed to allow any field.

Limitation currently is the callbacks and the feature config population

python/ml4ir/applications/ranking/model/metrics/metrics_impl.py
```diff
@@ -297,3 +302,257 @@ def _get_matches_hook(self, y_pred_click_ranks):
...
+        num_match = tf.cast(tf.math.count_nonzero(scores > 0, axis=-1), tf.float32)
+        match_failure = 1 - tf.cast(
+            RankMatchFailure.normalized_discounted_cumulative_gain(rank_scores, ranks_above_click),
+            tf.float32,
+        )
+        # If all records<=click have a name match, then it is not an NMF
+        # If number of scores>0 is same as the clicked rank, all ranks have a name match
+        match_failure = tf.where(
+            tf.equal(metric_click_ranks, num_match),
+            tf.constant(0, dtype=tf.float32),
```
arvindsrikantan, line 459: Check if this is supposed to be masked to `-np.inf`

python/ml4ir/base/model/metrics/metrics_impl.py
```diff
@@ -12,10 +13,11 @@ class MetricState:
...
 
 def get_metrics_impl(
-    metrics: List[Union[str, Type[Metric]]],
-    feature_config: FeatureConfig,
-    metadata_features: Dict,
-    **kwargs
+        metrics: List[Union[str, Type[Metric]]],
+        feature_config: FeatureConfig,
+        metadata_features: Dict,
+        for_aux_output: bool = False,
```
mohazahran: This needs to be added to the Parameters list doc string.

python/ml4ir/applications/ranking/model/metrics/metrics_impl.py
```diff
@@ -1,14 +1,19 @@
...
+from tensorflow.keras import metrics
+from tensorflow.python.ops import math_ops
 
-from ml4ir.base.model.metrics.metrics_impl import MetricState
 from ml4ir.base.features.feature_config import FeatureConfig
+from ml4ir.base.model.metrics.metrics_impl import MetricState
 
-from typing import Optional, Dict
+
+class CombinationMetric:
```
mohazahran: Can you remind me, why this is needed?

python/ml4ir/base/model/metrics/metrics_impl.py
```diff
@@ -41,6 +43,9 @@ def get_metrics_impl(
     metrics_impl: List[Union[Metric, str]] = list()
 
     for metric in metrics:
+        if isinstance(metric, ranking_metrics_impl.CombinationMetric) and for_aux_output:
```
mohazahran: Here, it's no allowing RankMatchFailure to be a metric for aux_ouput, but it allows the other metrics as (MRR) to be a metric for aux output, right?
It seems to me that MRR and the other non NMF metrics shouldn't be computed for the aux output because they don't convey any info. In fact it's useless to learn the value of train_aux_ranking_score_new_MRR as it's being measured against the title scores which are not clicks (correct me if I'm wrong here) . However, the NMF metric should be part of either the primary output or the aux output (if it's allowed for both, then both should give the same metric value, right? i.e. for example: train_ranking_score_new_RankMatchFailure should be equals to train_aux_ranking_score_new_RankMatchFailure.

python/ml4ir/applications/ranking/tests/test_auxiliary_loss.py
```diff
@@ -0,0 +1,80 @@
...
+        feature_config_path = os.path.join(ROOT_DATA_DIR, "configs", "feature_config_aux_loss.yaml")
+        model_config_path = os.path.join(ROOT_DATA_DIR, "configs", "model_config_cyclic_lr.yaml")
+        data_dir = os.path.join(ROOT_DATA_DIR, "tfrecord")
+        train_ml4ir(data_dir, feature_config_path, model_config_path, self.log_dir)
+
+        ml4ir_results = pd.read_csv(os.path.join(self.log_dir, 'test_aux_loss', '_SUCCESS'), header=None)
+        primary_training_loss = float(ml4ir_results.loc[ml4ir_results[0] == 'train_ranking_score_loss'][1])
+        assert np.isclose(primary_training_loss, 1.1877643, atol=0.0001)
+        aux_training_loss = float(ml4ir_results.loc[ml4ir_results[0] == 'train_aux_ranking_score_loss'][1])
+        assert np.isclose(aux_training_loss, 2.3386843, atol=0.0001)
```
mohazahran, line 105: This expected value "2.3386843" is not correct and it's failing the test case. I checked, this value is changed from my PR. This is the value from my PR:
assert np.isclose(aux_training_loss, 1.2242277, atol=0.0001)

python/ml4ir/applications/ranking/model/metrics/metrics_impl.py
arvindsrikantan: This is used to distinguish single label metrics from multi-label metrics. Multi-label metrics should be computed only for one of the outputs

python/ml4ir/base/model/metrics/metrics_impl.py
arvindsrikantan: To keep the implementation straight-forward, combination metrics are defined only for the "main output". The other metrics can/cannot be relevant to the aux-output (depending on the actual aux label). That's the reasoning behind the current design. Let me know if you feel another approach is more appropriate. 

> However, the NMF metric should be part of either the primary output or the aux output (if it's allowed for both, then both should give the same metric value, right? i.e. for example: train_ranking_score_new_RankMatchFailure should be equals to train_aux_ranking_score_new_RankMatchFailure.

This is accurate, but an implementation nightmare at this point. Open to suggestions here

python/ml4ir/applications/ranking/tests/test_auxiliary_loss.py
arvindsrikantan: Thanks for pointing this out. I might have messed up a merge conflict resolution

python/ml4ir/base/model/metrics/metrics_impl.py
mohazahran: In that case, it seems to me that no metrics other than loss should be tracked for the aux_output. and all other metrics loss, MRR, ACR and RankNMF should be tracked for the main output. In other words, the only change from training a model without aux_labels is that we should we should add RankNMF to the set of metrics tracked by the main output whenever aux_label is given.

arvindsrikantan: This makes sense for the particular aux feature we're talking about. We might want to measure MRR on a different aux output, no?

mohazahran: Generally it's possible, but that aux source has to be clicks, no?
If we want to make this account for all different cases of possible aux targets/outputs, then perhaps we need to make the metrics to be tracked a user input. What do you think?

mohazahran (approved):
Thanks Arvind for this PR!
I recommend a follow up story to better select metrics for each output (primary and aux.)

python/ml4ir/base/model/metrics/metrics_impl.py
arvindsrikantan: Yes, that's what I was thinking too. Will track this in a follow up

PR #186, opened 2022-08-19
Adding Ranking explainer for ml4ir (Demo)

tanmaylaud:
This PR adds a notebook demo for using omnixai library for generating per-query valid explanations for ml4ir models

### DIFF
--- a/python/notebooks/Ranking_Explanations.ipynb
+++ b/python/notebooks/Ranking_Explanations.ipynb
@@ -1,0 +1,4226 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Learning to Rank Expanations Demo 2022"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Overview\n",
+    "\n",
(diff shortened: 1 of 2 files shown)


lastmansleeping (approved):

lastmansleeping (changes requested):
@tanmaylaud Can you add `omnixai` (and specify a version) here? https://github.com/salesforce/ml4ir/blob/master/python/optional_requirements.yaml

python/optional_requirements.yaml
```diff
@@ -21,6 +21,7 @@
 # To install ml4ir `all`, run pip install ml4ir[all]
 all:
   - pyspark==3.0.1  # required to run ml4ir.base.pipeline
+  - omnixai==1.1.4  #required for running explanations demo. Upgrade to 1.1.5 when it is available
```
lastmansleeping, line 24: I meant to add it as another optional install too. So add something like this below.

```
explainer:
  - omnixai==1.1.4 #required for running explanations demo. Upgrade to 1.1.5 when it is available
```

lastmansleeping, line 24:
```suggestion
  - omnixai==1.1.4  # required for running explanations demo. Upgrade to 1.1.5 when it is available
```

tanmaylaud: Done

lastmansleeping (approved):

PR #193, opened 2022-11-19
Allowing custom classificationmodel classes

lastmansleeping:


python/ml4ir/applications/classification/pipeline.py
```diff
@@ -41,6 +41,8 @@ def __init__(self, args: Namespace):
         self.loss_key = args.loss_key
         super().__init__(args)
 
+        self.classification_model_cls = ClassificationModel
```
balikasg, line 44: It is not clear to me what this buys us at this point. I am sure there is a reason behind it, I am wondering if having a class method e.g., `get_default_classification_class` would be more clear. At this point on can not override this from the constructor either.

balikasg (approved):
LGTM

Ullimague (approved):

PR #196, opened 2023-01-11
Upgrade version to 0.0.6-SNAPSHOT

ArmandGiraud:


lastmansleeping (approved):
Thanks for the update @ArmandGiraud

ArmandGiraud:
@lastmansleeping I believe we should update python code, publish on pypi, and freeze the code by making github tag/release. Then move to new snapshot.

lastmansleeping:
> @lastmansleeping I believe we should update python code, publish on pypi, and freeze the code by making github tag/release. Then move to new snapshot.

We already have tagged releases. The python and java version are not necessarily in sync in terms of version numbers. That should be fine I think. We can maybe add a table mapping python release to jvm release in the README. I think you should merge this for now.

PR #202, opened 2023-01-20
Adding support for auxiliary loss

lastmansleeping:


ArmandGiraud (approved):
LGTM given it was previous reviewed independently

ArmandGiraud (approved):

ArmandGiraud (approved):

python/ml4ir/applications/ranking/config/keys.py
```diff
@@ -9,6 +9,9 @@ class LossKey(Key):
     RANK_ONE_LISTNET = "rank_one_listnet"
     SOFTMAX_CROSS_ENTROPY = "softmax_cross_entropy"
 
+    AUX_ONE_HOT_CROSS_ENTROPY = "aux_one_hot_cross_entropy"
```
lastmansleeping, line 12: @mohazahran I've renamed the losses, but at a high level, they are the same.

The one hot aux loss converts `y_true` to one hot labels and the softmax loss converts them to softmax probabilities.

python/ml4ir/applications/ranking/model/losses/listwise_losses.py
```diff
@@ -91,6 +91,99 @@ def final_activation_op(self, inputs, training=None):
         return self.final_activation_fn(logits)
 
 
+class AuxiliaryOneHotCrossEntropy(SoftmaxCrossEntropy):
+    """
+    Compute the one-hot softmax cross entropy loss on the auxiliary label
+    """
+
+    def call(self, inputs, y_true, y_pred, training=None):
```
lastmansleeping, line 99: @mohazahran The function definitions are now different because they inherit from tensorflow layers itself.

python/ml4ir/applications/ranking/model/losses/listwise_losses.py
```diff
@@ -91,6 +91,99 @@ def final_activation_op(self, inputs, training=None):
         return self.final_activation_fn(logits)
 
 
+class AuxiliaryOneHotCrossEntropy(SoftmaxCrossEntropy):
```
lastmansleeping, line 94: Both the aux losses are inherited from `SoftmaxCrossEntropy` because they are using the same activation function on `y_pred` / logits.

python/ml4ir/applications/ranking/model/losses/listwise_losses.py
```diff
@@ -91,6 +91,99 @@ def final_activation_op(self, inputs, training=None):
...
+        y_true = tf.cast(y_true, y_pred.dtype)
+
+        # Convert y_true to 1-hot labels
+        y_true_one_hot = tf.equal(y_true, tf.expand_dims(tf.math.reduce_max(y_true, axis=1), axis=1))
+        y_true_one_hot = tf.cast(y_true_one_hot, dtype=y_pred.dtype)
+
+        # Scale down the loss of a query by 1 / (number of ties)
+        sample_weight = tf.math.divide(tf.constant(1, dtype=tf.float32), tf.reduce_sum(y_true_one_hot, axis=1))
+
+        return self.loss_fn(y_true=tf.math.multiply(y_true_one_hot, mask),
```
lastmansleeping, line 137: I think I replaced `y_true` -> `tf.math.multiply(y_true_one_hot, mask)` across the file and even here.

python/ml4ir/applications/ranking/model/losses/listwise_losses.py
```diff
@@ -91,6 +91,99 @@ def final_activation_op(self, inputs, training=None):
...
+        mask = tf.cast(inputs[FeatureTypeKey.MASK], y_pred.dtype)
+        y_true = tf.cast(y_true, y_pred.dtype)
+
+        # Convert y_true to a probability distribution
+        y_true_softmax = self.final_activation_op({
+            FeatureTypeKey.METADATA: {
+                FeatureTypeKey.MASK: mask
+            },
+            FeatureTypeKey.LOGITS: y_true
+        }, training=training)
```
lastmansleeping, line 181: @mohazahran I'm using the activation function to get the masked softmax.

python/ml4ir/applications/ranking/pipeline.py
```diff
@@ -66,6 +67,23 @@ def get_loss(self):
...
+        RelevanceLossBase object
+        """
+        if self.aux_loss_key:
+            return loss_factory.get_loss(
+                loss_key=self.aux_loss_key,
+                scoring_type=self.scoring_type,
+                output_name="aux_{}".format(self.args.output_name)
+            )
+        else:
+            return None
```
lastmansleeping, line 85: @mohazahran I've also removed the `is_aux_loss` flag in the losses and made them separate classes. You can specify which aux loss you want similar to before with the `aux_loss_key` arg.

python/ml4ir/applications/ranking/tests/data/configs/feature_config_aux_loss.yaml
```diff
@@ -0,0 +1,155 @@
...
+  serving_info:
+    name: clicked
+  tfrecord_type: sequence
+features:
+  - name: text_match_score
+    node_name: text_match_score
+    trainable: true
+    dtype: float
+    log_at_inference: true
+    is_aux_label: true
```
lastmansleeping, line 44: @mohazahran FYI - this is a major change compared to earlier. There is no separate `aux_label` key anymore. Replaced with a sub field called `is_aux_label`. This simplified some of the TFRecord parsing logic IMO.

python/ml4ir/applications/ranking/tests/test_auxiliary_loss.py
```diff
@@ -0,0 +1,128 @@
...
+
+from ml4ir.applications.ranking.config.parse_args import get_args
+from ml4ir.applications.ranking.pipeline import RankingPipeline
+
+warnings.filterwarnings("ignore")
+
+ROOT_DATA_DIR = "ml4ir/applications/ranking/tests/data"
+
+
+def train_ml4ir(data_dir, feature_config, model_config, logs_dir, aux_loss):
```
lastmansleeping, line 20: @mohazahran FYI - old test
https://github.com/salesforce/ml4ir/blob/1f4ef603f132949c2c88a3ed852733fcf9de95f3/python/ml4ir/applications/ranking/tests/test_auxiliary_loss.py

Two changes that might be impacting the results (I've verified this by simply changing some of it and running the old tests successfully).
- Changing `y_true` -> `tf.multiply(y_true, mask)`
- Masked softmax

python/ml4ir/applications/ranking/model/losses/listwise_losses.py
```diff
@@ -91,6 +91,99 @@ def final_activation_op(self, inputs, training=None):
...
+
+        return self.loss_fn(y_true=tf.math.multiply(y_true_softmax, mask),
```
lastmansleeping, line 183: @mohazahran I replaced the manual cross entropy computation with tensorflow's CCE loss which implicitly handles `log(0)` and nulls and such. It is much safer and stable to do this from my experience.

lastmansleeping: I've verified that it should give the same result too using your sanity test.

python/ml4ir/applications/ranking/tests/test_losses.py
```diff
@@ -80,4 +80,55 @@ def test_rank_one_list_net(self):
         assert np.isclose(y_pred[2][4].numpy(), 0.0, atol=1e-5)
 
         loss = loss_op({"mask": self.mask}, self.y_true, y_pred)
-        assert np.isclose(loss, 2.1073625, atol=1e-5)
\ No newline at end of file
+        assert np.isclose(loss, 2.1073625, atol=1e-5)
+
+    def test_aux_one_hot_hot_cross_entropy(self):
```
lastmansleeping, line 85: @mohazahran FYI - old test -> https://github.com/salesforce/ml4ir/blob/1f4ef603f132949c2c88a3ed852733fcf9de95f3/python/ml4ir/applications/ranking/tests/test_losses.py

python/ml4ir/applications/ranking/tests/test_losses.py
```diff
@@ -80,4 +80,55 @@ def test_rank_one_list_net(self):
...
+            "metadata": {
+                "mask": self.mask
+            }
+        })
+
+        assert np.isclose(y_pred[0][0].numpy(), 0.19868991, atol=1e-5)
+        assert np.isclose(y_pred[2][4].numpy(), 0.0, atol=1e-5)
+
+        loss = loss_op({"mask": self.mask}, self.y_true_aux, y_pred)
+        assert np.isclose(loss, 0.5249801, atol=1e-5)
```
lastmansleeping, line 100: This is the same value as before.

python/ml4ir/applications/ranking/tests/test_losses.py
```diff
@@ -80,4 +80,55 @@ def test_rank_one_list_net(self):
...
+            "metadata": {
+                "mask": self.mask
+            }
+        })
+
+        assert np.isclose(y_pred[0][0].numpy(), 0.19868991, atol=1e-5)
+        assert np.isclose(y_pred[2][4].numpy(), 0.0, atol=1e-5)
+
+        loss = loss_op({"mask": self.mask}, self.y_true_aux_ties, y_pred)
+        assert np.isclose(loss, 0.8936954, atol=1e-5)
```
lastmansleeping, line 117: This is different because of `y_true` -> `y_true * mask`. I think the latter is the more "correct" implementation because otherwise if all scores are 0,  like in this example (`[0, 0, 0, 0, 0]`), then you get loss contribution from all of them due to one_hot function using `max`. I don't think this happens in an actual training case, but it is the case in this test setting.

python/ml4ir/applications/ranking/tests/test_losses.py
```diff
@@ -80,4 +80,55 @@ def test_rank_one_list_net(self):
...
+            "metadata": {
+                "mask": self.mask
+            }
+        })
+
+        assert np.isclose(y_pred[0][0].numpy(), 0.19868991, atol=1e-5)
+        assert np.isclose(y_pred[2][4].numpy(), 0.0, atol=1e-5)
+
+        loss = loss_op({"mask": self.mask}, self.y_true_aux, y_pred)
+        assert np.isclose(loss, 0.88127804, atol=1e-5)
```
lastmansleeping, line 134: @mohazahran This is different because of using masked softmax. I've verified that we get same result as before if we use simple softmax.

PR #207, opened 2023-01-27
Adding power analysis in the ranking evaluation

mohazahran:


python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -23,7 +24,17 @@
 
 class RankingConstants:
     NEW_RANK = "new_rank"
+    NEW_MRR = "new_MRR"
+    OLD_MRR = "old_MRR"
+    OLD_RANKNMF = "old_RankNMF"
+    NEW_RANKNMF = "new_RANKNMF"
```
lastmansleeping, line 30: @mohazahran I think we should add these later (or in a different PR).

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -23,7 +24,17 @@
...
+    MRR_DIFF = "MRR_diff"
     TTEST_PVALUE_THRESHOLD = 0.1
+    STATISTICAL_POWER = 0.9
+    # list of metrics for which to compute the variance in batches.
+    VARIANCE_METRIC_LIST = [NEW_MRR, OLD_MRR]
+    # list of metrics for which power analysis to be used.
+    METRIC_POWER_ANALYSIS = ["MRR"]
```
lastmansleeping, line 37: These need to be moved to a model config and not hardcoded. I think an even cleaner/better thing to do is define an `evaluation_config.yaml`. That makes the feature_config, model_config and eval_config as clean loosely connected pieces to control the model run.

lastmansleeping, line 37: These need to be moved to a model config and not hardcoded. I think an even cleaner/better thing to do is define an evaluation_config.yaml. That makes the feature_config, model_config and eval_config as clean loosely connected pieces to control the model run.

python/ml4ir/applications/ranking/t_test.py
```diff
@@ -1,5 +1,324 @@
-from scipy import stats
 import numpy as np
+import scipy
```
lastmansleeping, line 2: @mohazahran This should be usable for all `ml4ir` apps. Can you make it generic? I understand this was initially under `ranking/` but IMO that's not right too. You don't have to integrate it with Classification, but I think you should keep the option open for someone who wants to use this as an `import t_test` when using ml4ir to train models.

python/ml4ir/applications/ranking/t_test.py
```diff
@@ -1,5 +1,324 @@
...
+            tcrit = stats.t.ppf(1 - alpha / tside, dof)
+            return stats.nct.sf(tcrit, dof, nc) + stats.nct.cdf(-tcrit, dof, nc)
+
+    else:  # Alternative = 'greater'
+
+        def func(d, n, power, alpha):
+            dof = (n - 1) * tsample
+            nc = d * np.sqrt(n / tsample)
+            tcrit = stats.t.ppf(1 - alpha / tside, dof)
+            return stats.nct.sf(tcrit, dof, nc)
```
lastmansleeping, line 138: Why can't `alternative` just be an argument and then you define the function if `if-else` blocks? This seems like a lot of code duplication and I am not sure it is a good coding pattern.

python/ml4ir/applications/ranking/t_test.py
```diff
@@ -1,5 +1,324 @@
...
+    else:
+        # Compute achieved alpha (significance) level given d, n and power
+
+        def _eval_alpha(alpha, d, n, power):
+            return func(d, n, power, alpha) - power
+
+        try:
+            return brenth(_eval_alpha, 1e-10, 1 - 1e-10, args=(d, n, power))
+        except ValueError:  # pragma: no cover
+            return np.nan
```
lastmansleeping, line 182: @mohazahran You can use [lambda functions](https://www.geeksforgeeks.org/python-lambda-anonymous-functions-filter-map-reduce/) to wrap this I think. (I still haven't looked at this in detail).

python/ml4ir/applications/ranking/t_test.py
```diff
@@ -1,5 +1,324 @@
...
+
+
+def power_ttest2n(nx, ny, d=None, power=None, alpha=0.05, alternative="two-sided"):
+    """
+    Extracted from https://pingouin-stats.org/build/html/generated/pingouin.power_ttest2n.html
```
lastmansleeping, line 187: @mohazahran Why not just add a `pingouin` dependency? Copy-pasting code means we don't get any of the new fixes/updates they make.

Make sure you request 3PP in parallel.

python/ml4ir/applications/ranking/t_test.py
```diff
@@ -1,5 +1,324 @@
...
+class StreamVariance:
+    # Wrapper class to store mean, variance in batches
+    count = 0
+    mean = 0
+    var = 0
+
+
+def power_ttest(d=None, n=None, power=None, alpha=0.05, contrast="two-samples", alternative="two-sided"):
+    """
+    Extracted from https://pingouin-stats.org/build/html/generated/pingouin.power_ttest.html
```
lastmansleeping, line 17: (same comment as below)
@mohazahran Why not just add a pingouin dependency? Copy-pasting code means we don't get any of the new fixes/updates they make.

Make sure you request 3PP in parallel.

python/ml4ir/applications/ranking/t_test.py
```diff
@@ -88,3 +407,127 @@ def t_test_log_results(t_test_stat, pvalue, ttest_pvalue_threshold, logger):
        logger.warning(
             "With p-value threshold={} < p-value --> we cannot reject the null hypothesis. The click rank distribution of the new model is not significantly different from the old model".format(
                 ttest_pvalue_threshold))
+
+
+def compute_org_running_variance_for_metrics(metric_list, orgs_metric_running_variance_params, running_stats_df):
```
lastmansleeping, line 412: @mohazahran Remove internal lingo like "org"

python/ml4ir/applications/ranking/t_test.py
```diff
@@ -88,3 +407,127 @@ def t_test_log_results(t_test_stat, pvalue, ttest_pvalue_threshold, logger):
...
+                          ((sv.count * new_count * (sv.mean - new_mean)**2) / ((sv.count + new_count)*(sv.count + new_count - 1)))
+            else:
+                combine_var = 0
+
+            sv.count = combine_count
+            sv.mean = combined_mean
+            sv.var = combine_var
+
+            org_params[metric] = sv
+            orgs_metric_running_variance_params[org] = org_params
```
lastmansleeping, line 459: Lot of internal lingo here.

python/ml4ir/applications/ranking/t_test.py
```diff
@@ -88,3 +407,127 @@ def t_test_log_results(t_test_stat, pvalue, ttest_pvalue_threshold, logger):
...
+        A dictionary containing mean, variance and sample size for each org for each metric
+    statistical_power: float
+        Required statistical power
+    pvalue: float
+        Required pvalue
+
+    Returns
+    -------
+    org_metrics_stat_sig: Pandas dataframe
+        A dataframe listing each org and for each metric whether the change is statistically significant
```
lastmansleeping, line 514: @mohazahran Internal lingo alert.

python/ml4ir/base/tests/test_stat_analysis_calculation.py
```diff
@@ -1,17 +1,18 @@
...
-from ml4ir.applications.ranking.t_test import perform_click_rank_dist_paired_t_test, compute_stats_from_stream
+from ml4ir.applications.ranking.t_test import perform_click_rank_dist_paired_t_test, compute_stats_from_stream, compute_org_running_variance_for_metrics, StreamVariance
 
 warnings.filterwarnings("ignore")
 
 np.random.seed(123)
 
 
-class TestTtestCalculation(unittest.TestCase):
+class StatisticalAnalysisCalculation(unittest.TestCase):
```
lastmansleeping, line 13: [Minor] @mohazahran Shouldn't this start with `Test` to be picked up? Regardless, might be good to add the prefix for consistency.

python/ml4ir/applications/ranking/t_test.py
```diff
@@ -88,3 +407,127 @@ def t_test_log_results(t_test_stat, pvalue, ttest_pvalue_threshold, logger):
...
+    try:
+        # compute the effect size (d)
+        d = np.abs(float(mean1) - float(mean2)) / np.sqrt((float(var1) + float(var2)) / 2)
+        req_sample_sz = power_ttest(d, n, statistical_power, pvalue, contrast=typ, alternative=alternative)
+        return req_sample_sz
+    except:
+        return -1.0
+
+
+def run_power_analysis(metric_list, orgs_metric_running_variance_params, statistical_power, pvalue):
```
lastmansleeping, line 497: @mohazahran Add unit test for this.

python/ml4ir/applications/ranking/t_test.py
```diff
@@ -88,3 +407,127 @@ def t_test_log_results(t_test_stat, pvalue, ttest_pvalue_threshold, logger):
...
+
+            sv.count = combine_count
+            sv.mean = combined_mean
+            sv.var = combine_var
+
+            org_params[metric] = sv
+            orgs_metric_running_variance_params[org] = org_params
+
+
+def compute_required_sample_size(mean1, mean2, var1, var2, statistical_power, pvalue):
```
lastmansleeping, line 462: @mohazahran Add unit test for this.

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -185,13 +195,31 @@ def evaluate(
...
+            clicked_records[RankingConstants.NEW_MRR] = 1.0 / clicked_records["new_rank"]
+            clicked_records[RankingConstants.OLD_MRR] = 1.0 / clicked_records[self.feature_config.get_rank("node_name")]
+
+            # computing batch-wise mean and variance per metric
+            metric_stat_df_list = []
+            clicked_records.set_index("organization_id")
+            for metric in RankingConstants.VARIANCE_METRIC_LIST:
+                temp_df = clicked_records.groupby('organization_id').agg({
+                    metric: ['mean', 'var'],
+                    "organization_id": ['count']
```
lastmansleeping, line 213: @mohazahran Internal lingo alert.

lastmansleeping (changes requested):
@mohazahran Did a first pass review and left a few comments for changes. I think you should also add tests for overall metric computation.

python/ml4ir/applications/ranking/t_test.py
mohazahran: I did not want to add another dependency. I remember `pingouin` it was not part of the internal pypi

python/ml4ir/applications/ranking/t_test.py
mohazahran: Perhaps. However, I extracted the code from the lib to not add dependency and I didn't spend time to improve it.

python/ml4ir/applications/ranking/t_test.py
mohazahran: Same as the previous comment.

python/ml4ir/applications/ranking/model/ranking_model.py
mohazahran: Let'm do that in the next story.

python/ml4ir/applications/ranking/t_test.py
mohazahran: I will do that. I believe till this happen we can stick to the extracted functions.

python/ml4ir/applications/ranking/model/ranking_model.py
lastmansleeping: ping

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -185,13 +167,21 @@ def evaluate(
...
-            diff = (1/clicked_records[RankingConstants.NEW_RANK] - 1/clicked_records[
-                self.feature_config.get_rank("node_name")]).to_list()
+            clicked_records[RankingConstants.NEW_MRR] = 1.0 / clicked_records["new_rank"]
+            clicked_records[RankingConstants.OLD_MRR] = 1.0 / clicked_records[self.feature_config.get_rank("node_name")]
+            # Statistical analysis pre-processing
+            if len(group_key) > 0:
+                statistical_analysis_preprocessing(clicked_records, group_metric_running_variance_params, group_key,
+                                                   RankingConstants.VARIANCE_METRIC_LIST)
+
+            diff = (clicked_records[RankingConstants.NEW_MRR] - clicked_records[RankingConstants.OLD_MRR]).to_list()
```
lastmansleeping, line 184: @mohazahran Isn't MRR and ACR already being computed in the `metrics_helper`? Why do we need to do it here again?

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -236,12 +233,21 @@ def evaluate(
...
             )
+            # Add power analysis to group metric dataframe
+            if len(group_key) > 0:
+                df_group_metrics = df_group_metrics.reset_index()
+                if len(group_key) > 1:
+                    df_group_metrics[str(group_key)] = df_group_metrics[group_key].apply(tuple, axis=1)
+                df_group_metrics[str(group_key)] = df_group_metrics[group_key]
+                df_group_metrics = pd.merge(df_group_metrics, group_metrics_stat_sig, on=str(group_key), how='left').drop(
+                    columns=[str(group_key)])
+
```
lastmansleeping, line 250: @mohazahran Can you explain why we need to do this separately?

python/ml4ir/applications/ranking/t_test.py
lastmansleeping: ping.

Just move it to `ml4ir.base.stats` for now. We can refactor and add to ClassificationPipeline later.

python/ml4ir/applications/ranking/t_test.py
lastmansleeping: I would prefer adding the dependency to this PR as opposed to checking in this source code from another repo if avoidable. Let's try to get the 3PP request going and see if it's delayed. Can decide next action based on that.

Also, the reason I ask for adding the dependency is I do not know which version of the library works well for our usecase. So it will be good to take it into consideration for the unittests and such.

python/ml4ir/applications/ranking/t_test.py
lastmansleeping: Yes, I made these comments before I realized this was from pinguoin. Haha

python/ml4ir/applications/ranking/t_test.py
lastmansleeping: Yes, I checked too. But you can easily add it. Will send you link on slack.

python/ml4ir/applications/ranking/t_test.py
```diff
@@ -88,3 +408,213 @@ def t_test_log_results(t_test_stat, pvalue, ttest_pvalue_threshold, logger):
...
+    logger: Logger
+        Logger object to log t-test significance decision
+    mean: float
+        The mean of the rank differences for the entire dataset
+    variance: float
+        The variance of the rank differences for the entire dataset
+    n: int
+        The number of samples in the entire dataset
+    ttest_pvalue_threshold: float
+        P-value threshold for student t-test
```
lastmansleeping, line 427: @mohazahran [Minor] Order is wrong

python/ml4ir/applications/ranking/t_test.py
```diff
@@ -88,3 +408,213 @@ def t_test_log_results(t_test_stat, pvalue, ttest_pvalue_threshold, logger):
...
+        count_df = pd.DataFrame(
+            {str(group_key): grouped_agg.keys().values, str([metric, "count"]): grouped_agg.values})
+
+        concat_df = pd.concat([mean_df, var_df, count_df], axis=1)
+        metric_stat_df_list.append(concat_df)
+
+    running_stats_df = pd.concat(metric_stat_df_list, axis=1)
+    running_stats_df = running_stats_df.loc[:, ~running_stats_df.columns.duplicated()]
+    # Accumulating batch-wise mean and variances
+    compute_org_running_variance_for_metrics(variance_list, group_metric_running_variance_params, running_stats_df, group_key)
```
lastmansleeping, line 483: @mohazahran Why can't this be moved to `metrics_helper`? You just need the mean, variance and count for each group, right? Check it out. It should be just a few new line additions IMO.

lastmansleeping: It is also doing it in batches.

python/ml4ir/base/tests/test_stat_analysis_calculation.py
```diff
@@ -0,0 +1,188 @@
...
+        batch_a = np.random.exponential(scale=0.1, size=n)
+        batch_b = np.random.exponential(scale=0.1, size=n)
+        self.t_test_calculation(batch_a, batch_b, batch_size)
+
+    def test_9(self):
+        n = 50
+        batch_size = 5
+        batch_a = np.random.exponential(scale=0.1, size=n)
+        batch_b = batch_a + 0.001
+        self.t_test_calculation(batch_a, batch_b, batch_size)
```
lastmansleeping, line 184: @mohazahran Thanks for the unit tests, can you also rename them to something appropriate? Add a short docstring about what you are trying to test for reference.

lastmansleeping (changes requested):
@mohazahran Thanks for the updates :)

I have left some comments, but my main "change request" is to look at `metrics_helper` and use it to compute the mean and variance for your t_test instead of rewriting the same code. `metrics_helper` is also doing a batch processing. So all of that is done for you.

Other changes
- Remove source code from pingouin and add dependecy
- Move `t_test` under `ml4ir.base.stats`

python/ml4ir/applications/ranking/model/ranking_model.py
mohazahran: there are just constant definitions. but can remove them from now.

python/ml4ir/applications/ranking/t_test.py
lastmansleeping: Question - after we do the stat analysis, we should I think also generate new macro means for the metrics.

As in we currently have `macro_avg_MRR` (it might be called something else, but). We would also need `macro_avg_MRR_stat_sig`

python/ml4ir/applications/ranking/model/ranking_model.py
mohazahran: because it's hard to modify `get_grouped_stats` to compute var/mean/count for a variable number of metrics without greatly changing the function altogether. so for now I will keep the power analysis as a parallel task

lastmansleeping:
@mohazahran Make sure pull in this PR too -> https://github.com/salesforce/ml4ir/pull/210

python/build-requirements.txt
```diff
@@ -12,7 +12,7 @@ mypy
 tensorflow==2.9.3
 numpy==1.21.6
 pandas==1.2.1
-scipy==1.5.4
+scipy==1.7
```
lastmansleeping, line 15: @mohazahran Can you fix the minor version too based on what's available internally?

python/requirements.txt
```diff
@@ -20,4 +20,7 @@ pyspark==3.0.1
 tensorflow-probability==0.15.0
 
 # DAG visualization
-pygraphviz==1.7
\ No newline at end of file
+pygraphviz==1.7
```
lastmansleeping, line 23: @mohazahran Why aren't you changing `scipy` version here?

python/ml4ir/base/config/default_evaluation_config.yaml
```diff
@@ -0,0 +1,22 @@
+power_analysis:
```
lastmansleeping, line 1: @mohazahran Oh wow, I'm glad you did this. I thought you were planning to do it in a follow up work/PR.

python/ml4ir/base/config/default_evaluation_config.yaml
```diff
@@ -0,0 +1,22 @@
...
+#    NEW_RANK = "new_rank"
+#    NEW_MRR = "new_MRR"
+#    OLD_MRR = "old_MRR"
+#    MRR_DIFF = "MRR_diff"
+#    TTEST_PVALUE_THRESHOLD = 0.1
+#    STATISTICAL_POWER = 0.9
+#    # list of metrics for which to compute the variance in batches.
+#    VARIANCE_METRIC_LIST = [NEW_MRR, OLD_MRR]
+#    # list of metrics for which power analysis to be used.
+#    METRIC_POWER_ANALYSIS = ["MRR"]
```
lastmansleeping, line 22: @mohazahran Cleanup

python/ml4ir/base/pipeline.py
```diff
@@ -579,6 +579,8 @@ def run_pipeline(self, relevance_dataset=None):
                     group_metrics_min_queries=self.args.group_metrics_min_queries,
                     logs_dir=self.logs_dir_local,
                     compute_intermediate_stats=self.args.compute_intermediate_stats,
+                    evaluation_config_path=self.args.evaluation_config,
```
lastmansleeping, line 582: @mohazahran I recommend reading the yaml file and passing a dictionary here instead.

python/ml4ir/applications/ranking/model/metrics/metrics_helper.py
```diff
@@ -231,6 +231,59 @@ def compute_aux_metrics_on_query_group(
     return pd.Series(aux_metrics_dict)
 
 
+def compute_groupwise_running_variance_for_metrics(metric_list, group_metric_running_variance_params, running_stats_df, group_key):
```
lastmansleeping, line 234: @mohazahran Why can't this function be called from `get_grouped_stats()`? That way you can remove the call from `ranking_model`

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -70,34 +75,28 @@ def predict(
...
+                eval_dict["variance_list"].append("new_" + m)
+            eval_dict["power"] = float(eval_config["power_analysis"]["power"])
+            eval_dict["pvalue"] = float(eval_config["power_analysis"]["pvalue"])
+        else:
+            eval_dict["group_by"] = group_key
+            eval_dict["metrics"] = []
+            eval_dict["variance_list"] = []
+            eval_dict["power"] = None
+            eval_dict["pvalue"] = RankingConstants.TTEST_PVALUE_THRESHOLD
+        return eval_dict
```
lastmansleeping, line 99: @mohazahran As mentioned elsewhere, initialize this in the pipeline itself. And create a new class call `EvalConfig` as this looks like a lot of customization.

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -148,6 +149,10 @@ def evaluate(
 
         Override this method to implement your own evaluation metrics.
         """
+        group_key = list(set(self.feature_config.get_group_metrics_keys("node_name")))
```
lastmansleeping, line 152: @mohazahran Should be plural.

python/ml4ir/base/config/default_evaluation_config.yaml
```diff
@@ -0,0 +1,22 @@
+power_analysis:
+  # a list of input feature names (comma separated) used to aggregate results. Leave blank if not used
+  group_by:
```
lastmansleeping, line 3: @mohazahran Isn't this already captured in `FeatureConfig`? I don't think you need this here.

python/ml4ir/base/stats/t_test.py
```diff
@@ -0,0 +1,312 @@
...
+    t_test_log_results(t_test_stat, pvalue, ttest_pvalue_threshold, logger)
+
+    # Log t-test statistic and p-value
+    t_test_metrics_dict = {'Rank distribution t-test statistic': t_test_stat,
+                           'Rank distribution t-test pvalue': pvalue,
+                           'Rank distribution t-test difference is statistically significant': pvalue < ttest_pvalue_threshold}
+    return t_test_metrics_dict
+
+
+def statistical_analysis_preprocessing(clicked_records, group_metric_running_variance_params, group_key, variance_list):
```
lastmansleeping, line 136: @mohazahran Rename the method to something that mentions computing mean and variance for a batch.

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -14,26 +14,31 @@
 from ml4ir.applications.ranking.model.scoring import prediction_helper
 from ml4ir.applications.ranking.model.metrics import metrics_helper
 from ml4ir.applications.ranking.config.keys import PositionalBiasHandler
-from ml4ir.applications.ranking.t_test import perform_click_rank_dist_paired_t_test, compute_stats_from_stream, t_test_log_results
-
+from ml4ir.base.io.file_io import FileIO
+from ml4ir.base.stats.t_test import perform_click_rank_dist_paired_t_test, compute_stats_from_stream, \
+    t_test_log_results, run_ttest, power_ttest, compute_groupwise_running_variance_for_metrics, run_power_analysis, \
+    StreamVariance, statistical_analysis_preprocessing
```
lastmansleeping, line 20: @mohazahran Ideally, you would import most of these in `metrics_helper` and not have this login inside `ranking_model`. That was in my mind the goal of the refactoring.

python/ml4ir/base/tests/test_stat_analysis_calculation.py
lastmansleeping: ping

You can maybe make them `self.subTest("...")`

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -240,6 +260,15 @@ def evaluate(
...
             )
+            # Add power analysis to group metric dataframe
+            if len(group_key) > 0 and len(eval_dict["metrics"]) > 0:
+                df_group_metrics = df_group_metrics.reset_index()
+                if len(group_key) > 1:
+                    df_group_metrics[str(group_key)] = df_group_metrics[group_key].apply(tuple, axis=1)
+                df_group_metrics[str(group_key)] = df_group_metrics[group_key]
+                df_group_metrics = pd.merge(df_group_metrics, group_metrics_stat_sig, on=str(group_key), how='left').drop(
+                    columns=[str(group_key)])
+
```
lastmansleeping, line 271: @mohazahran I don't fully understand what's being done here.

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -251,7 +280,7 @@ def evaluate(
             df_group_metrics_summary = df_group_metrics.describe()
             self.logger.info(
                 "Computing group metrics using keys: {}".format(
-                    self.feature_config.get_group_metrics_keys("node_name")
+                    eval_dict["group_by"]
```
lastmansleeping, line 283: @mohazahran Why change this?

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -184,13 +189,20 @@ def evaluate(
...
-            diff = (1/clicked_records[RankingConstants.NEW_RANK] - 1/clicked_records[
-                self.feature_config.get_rank("node_name")]).to_list()
+            clicked_records[RankingConstants.NEW_MRR] = 1.0 / clicked_records["new_rank"]
+            clicked_records[RankingConstants.OLD_MRR] = 1.0 / clicked_records[self.feature_config.get_rank("node_name")]
+            # Statistical analysis pre-processing
+            if len(group_key) > 0 and len(eval_dict["metrics"]) > 0:
+                statistical_analysis_preprocessing(clicked_records, group_metric_running_variance_params, group_key,
+                                                   eval_dict["variance_list"])
+
+            diff = (clicked_records[RankingConstants.NEW_MRR] - clicked_records[RankingConstants.OLD_MRR]).to_list()
```
lastmansleeping, line 205: @mohazahran `ranking_model` shouldn't be hardcoding these MRR metrics. Move it out to the `metrics_helper`.

lastmansleeping (changes requested):
@mohazahran Reviewed some of the structure changes. I still haven't reviewed the stat analysis component (but I've reviewed this before in your notebooks fully).

python/ml4ir/base/config/parse_args.py
```diff
@@ -83,6 +83,13 @@ def define_args(self):
             help="Path to YAML file or YAML string with feature metadata for training.",
         )
 
+        self.add_argument(
+            "--evaluation_config",
+            type=str,
+            help="Path to YAML file or YAML string with parameters needed for model evaluation of the test set.",
+            default="ml4ir/base/config/default_evaluation_config.yaml"
```
lastmansleeping, line 90: @mohazahran Would this be messing up things for `ClassificationModel`?

python/requirements.txt
mohazahran: It's a Pingoun dependency

python/ml4ir/applications/ranking/model/ranking_model.py
mohazahran: because in the eval yaml you can choose the group keys, so I'm using them rather than this. If none is provided in the yaml file, then `self.feature_config.get_group_metrics_keys("node_name")` will be used.

python/ml4ir/applications/classification/tests/test_classification_serving.py
```diff
@@ -73,7 +73,8 @@ def test_serve_default_pipeline(self):
             tfrecord_signature(protos=tf.gather(sequence_example_protos, [i]))[
                 self.args.output_name
             ]
-            for i in range(self.args.batch_size)
+            for
```
lastmansleeping, line 76: [Minor] @mohazahran This is such a strange formatting change haha.

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -14,26 +14,37 @@
...
+    StreamVariance, statistical_analysis_preprocessing
 
 pd.set_option("display.max_rows", 500)
 pd.set_option("display.max_columns", 500)
 
 
 class RankingConstants:
     NEW_RANK = "new_rank"
+    NEW_MRR = "new_MRR"
+    OLD_MRR = "old_MRR"
```
lastmansleeping, line 28: @mohazahran Remove the notion of MRR from `ranking_model`

python/ml4ir/applications/ranking/model/ranking_model.py
lastmansleeping: ping

python/build-requirements.txt
lastmansleeping: ping

python/ml4ir/applications/classification/tests/test_classification_serving.py
mohazahran: That's weird! I don't know how did that happen

python/ml4ir/base/config/parse_args.py
mohazahran: nope. I handled that

python/ml4ir/applications/ranking/model/metrics/helpers/metrics_helper.py
```diff
@@ -163,3 +164,79 @@ def summarize_grouped_stats(df_grouped):
...
+            group_params = {}
+            group_metric_running_variance_params[group_name] = group_params
+
+        for metric in metric_list:
+            new_mean = row[1][str([metric, "mean"])]
+            new_var = row[1][str([metric, "var"])]
+            new_count = row[1][str([metric, "count"])]
+
+            if metric in group_params:
+                sv = group_metric_running_variance_params[group_name][metric]
```
lastmansleeping, line 200: @mohazahran Useful to keep the variable name as `stream_variance`

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -184,13 +214,20 @@ def evaluate(
...
 
             # Accumulating statistics for t-test calculation using 1/rank
             clicked_records = predictions_df[predictions_df[self.feature_config.get_label("node_name")] == 1.0]
-            diff = (1/clicked_records[RankingConstants.NEW_RANK] - 1/clicked_records[
-                self.feature_config.get_rank("node_name")]).to_list()
+            clicked_records[RankingConstants.NEW_MRR] = 1.0 / clicked_records[RankingConstants.NEW_RANK]
+            clicked_records[RankingConstants.OLD_MRR] = 1.0 / clicked_records[self.feature_config.get_rank("node_name")]
+            # Statistical analysis pre-processing
+            if len(group_keys) > 0 and len(eval_dict[RankingConstants.METRICS]) > 0:
+                statistical_analysis_preprocessing(clicked_records, group_metric_running_variance_params, group_keys,
```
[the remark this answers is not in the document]
lastmansleeping: [Minor] In fact, you can create a new object for `BatchedMetrics` which contains different metrics you need for both micro and macro and pass it here instead of a dictionary.

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -240,6 +285,39 @@ def evaluate(
...
+                stat_sig_df_summary = stat_sig_df.describe(include='all')
+                self.logger.info(
+                    "\nComputing group metrics for {} stat. sig. lifts using keys: {}".format(metric,
+                                                                                              group_keys))
+                self.logger.info("\nGroupwise Metrics for {}: \n{}".format(metric, stat_sig_df_summary.T))
+                metrics_dict["stat_sig_" + metric + "_improved_groups"] = improv_count
+                metrics_dict["stat_sig_" + metric + "_degraded_groups"] = degrad_count
+                metrics_dict["stat_sig_" + metric + "_group_improv_perc"] = stat_sig_groupwise_metric_improv
+                metrics_dict["stat_sig_improved_" + metric + "_groups"] = sorted(improved)
+                metrics_dict["stat_sig_degraded_" + metric + "_groups"] = sorted(degraded)
```
lastmansleeping, line 319: @mohazahran This should definitely not be in `ranking_model`. All this "summarization" should be in the helpers.

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -240,6 +285,39 @@ def evaluate(
...
+                    self.file_io.write_df(
+                        stat_sig_df,
+                        outfile=os.path.join(
+                            logs_dir, "stat_sig_" + metric + '_' + RelevanceModelConstants.GROUP_METRICS_CSV_FILE),
+                    )
+                # Compute group metrics summary
+                stat_sig_df_summary = stat_sig_df.describe(include='all')
+                self.logger.info(
+                    "\nComputing group metrics for {} stat. sig. lifts using keys: {}".format(metric,
+                                                                                              group_keys))
```
lastmansleeping, line 313: @mohazahran Why are you logging them separately? We already had a log step for the group metrics, right?

python/ml4ir/base/config/default_evaluation_config.yaml
lastmansleeping: ping

python/ml4ir/base/model/relevance_model.py
```diff
@@ -91,6 +95,11 @@ def __init__(
         self.tfrecord_type = tfrecord_type
         self.file_io = file_io
 
+        if len(eval_config) == 0:
+            self.eval_config = self.file_io.read_yaml(RelevanceModelConstants.DEFAULT_EVAL_CONFIG_YAML)
```
lastmansleeping, line 99: @mohazahran This should be handled upstream in the pipeline class. The function definition expects a dictionary.

python/ml4ir/base/stats/t_test.py
lastmansleeping: ping

python/ml4ir/base/stats/t_test.py
```diff
@@ -0,0 +1,312 @@
...
+                          ((sv.count * new_count * (sv.mean - new_mean)**2) / ((sv.count + new_count)*(sv.count + new_count - 1)))
+            else:
+                combine_var = 0
+
+            sv.count = combine_count
+            sv.mean = combined_mean
+            sv.var = combine_var
+
+            group_params[metric] = sv
+            group_metric_running_variance_params[group_name] = group_params
```
lastmansleeping, line 228: @mohazahran This is duplicated code. I also see it in ranking `metrics_helper`

python/requirements.txt
lastmansleeping: Fix it to a minor version too.

python/ml4ir/applications/ranking/model/metrics/helpers/metrics_helper.py
```diff
@@ -163,3 +164,79 @@ def summarize_grouped_stats(df_grouped):
         processed_metric_name_suffixes.add(metric_name_suffix)
 
     return df_grouped_metrics
+
+
+def compute_groupwise_running_variance_for_metrics(metric_list, group_metric_running_variance_params, running_stats_df, group_key):
```
lastmansleeping, line 169: @mohazahran Add unit tests for this function.

python/ml4ir/applications/ranking/model/metrics/helpers/metrics_helper.py
```diff
@@ -163,3 +164,79 @@ def summarize_grouped_stats(df_grouped):
...
+
+            sv.count = combine_count
+            sv.mean = combined_mean
+            sv.var = combine_var
+
+            group_params[metric] = sv
+            group_metric_running_variance_params[group_name] = group_params
+
+
+def generate_stat_sig_based_metrics(df, metric, group_keys):
```
lastmansleeping, line 222: @mohazahran Add unit tests for this function

lastmansleeping (changes requested):

python/build-requirements.txt
mohazahran: 1.7 is in the pypi

python/ml4ir/applications/ranking/model/ranking_model.py
mohazahran: I'm generating a per metric log. e.g. logging stats from MRR stat sig groups, and logging stats from NMF stat sig groups.

python/ml4ir/base/config/default_evaluation_config.yaml
mohazahran: yes, but we there is a usecase where we can run multiple evaluations and want to change the group keys

python/ml4ir/base/model/relevance_model.py
mohazahran: I don't follow which function you mean.

python/ml4ir/applications/ranking/model/ranking_model.py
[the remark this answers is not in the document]
mohazahran: I can return it updated.

python/build-requirements.txt
lastmansleeping: I know that, I am asking you to also fix a minor version for the package like `1.7.x` because sometimes someone uploads a newer version to PyPI and then the autobuilds start picking that up instead. It's just an unnecessary maintenance mess.

python/ml4ir/base/config/default_evaluation_config.yaml
lastmansleeping: Then do that in the feature config. Why would we need to do it here?

lastmansleeping: I'm only pushing you because it leads to duplication and unnecessary tech debt.

python/ml4ir/base/model/relevance_model.py
lastmansleeping: The if-else. If there is no eval_config, read the default one in the pipeline class and pass it here. RelevanceModel shouldn't be doing this.

python/ml4ir/base/stats/t_test.py
```diff
@@ -0,0 +1,338 @@
...
+    statistical significant using the predefined statistical power and p-value
+    ----------
+    metric_list: list
+        List of all metrics to be used in the power analysis
+
+    group_key: list
+        The list of keys used to aggregate the metrics
+
+    group_metric_running_variance_params: dict
+        A dictionary containing mean, variance and sample size for each org for each metric
```
lastmansleeping, line 308: @mohazahran Remove internal info

python/ml4ir/base/stats/t_test.py
```diff
@@ -0,0 +1,338 @@
...
+    statistical_power: float
+        Required statistical power
+
+    pvalue: float
+        Required pvalue
+
+    Returns
+    -------
+    group_metrics_stat_sig: Pandas dataframe
+        A dataframe listing each org and for each metric whether the change is statistically significant
```
lastmansleeping, line 319: @mohazahran Remove internal info

python/ml4ir/base/tests/test_stat_analysis_calculation.py
```diff
@@ -0,0 +1,181 @@
...
+                    1.16220405, -1.09720305, -2.12310035, 1.03972709, -0.40336604, -0.12602959, -0.83751672,
+                    -1.60596276, 1.25523737, -0.68886898, 1.66095249, 0.80730819, -0.31475815, -1.0859024, -0.73246199,
+                    -1.21252313, 2.08711336, 0.16444123, 1.15020554, -1.26735205, 0.18103513, 1.17786194, -0.33501076,
+                    1.03111446, -1.08456791, -1.36347154, 0.37940061, -0.37917643])
+        b_bucket = a_bucket + 0.5
+        n = len(a_bucket)
+        group_metric_running_variance_params = {}
+        var_metric_list = ['old_metric1', 'new_metric1', 'old_metric2', 'new_metric2']
+        metric_list = ["metric1", "metric2"]
+        group_key = ["organization_id"]
```
lastmansleeping, line 90: @mohazahran Remove internal info

python/ml4ir/base/stats/t_test.py
```diff
@@ -0,0 +1,338 @@
...
+
+def compute_batched_stats(clicked_records, group_metric_running_variance_params, group_key, variance_list):
+    """
+    This function computes the mean, variance of the provided metrics for the current batch of prediction.
+    ----------
+    clicked_records: dataframe
+            A batch of the model's prediction
+
+    group_metric_running_variance_params: dict
+        A dictionary containing intermediate mean, variance and sample size for each org for each metric
```
lastmansleeping, line 170: @mohazahran Remove internal info.

python/ml4ir/base/stats/t_test.py
```diff
@@ -0,0 +1,338 @@
...
+
+def compute_groupwise_running_variance_for_metrics(metric_list, group_metric_running_variance_params, running_stats_df, group_key):
+    """
+    This function updates the intermediate mean and variance with each batch.
+    ----------
+    metric_list: list
+            Lists of metrics for variance computation in batches
+
+    group_metric_running_variance_params: dict
+        A dictionary containing intermediate mean, variance and sample size for each org for each metric
```
lastmansleeping, line 212: @mohazahran Remove internal info.

python/ml4ir/base/config/eval_config.py
```diff
@@ -0,0 +1,55 @@
...
+    METRICS = "metrics"
+
+
+def prepare_eval_config_for_ranking(eval_config, group_key):
+    """
+    Reads the evaluation config yaml file and convert it into an eval config dict
+
+    Parameters
+    ----------
+    evaluation_config_path : str
```
lastmansleeping, line 19: @mohazahran Name mismatch.

lastmansleeping: I'm also confused about whether yaml file is being passed or dictionary here, because there is no yaml read happening.

python/ml4ir/base/config/eval_config.py
```diff
@@ -0,0 +1,55 @@
+from ml4ir.applications.ranking.model.metrics.helpers.metrics_helper import RankingConstants
+
+class EvalConfigConstants:
```
lastmansleeping, line 3: @mohazahran Do you not need a notion of "higher is better" or "lower is better" to compute the p values for the metrics?

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -241,6 +226,28 @@ def evaluate(
...
+                                                                   group_metrics_stat_sig)
+
+            # Writing stat sig. groupwise results per metric
+            for metric in eval_dict[EvalConfigConstants.METRICS]:
+                stat_sig_df = metrics_helper.generate_stat_sig_based_metrics(df_group_metrics, metric, group_keys, metrics_dict)
+                if logs_dir:
+                    self.file_io.write_df(
+                        stat_sig_df,
+                        outfile=os.path.join(
+                            logs_dir, "stat_sig_" + metric + '_' + RelevanceModelConstants.GROUP_METRICS_CSV_FILE),
```
lastmansleeping, line 242: @mohazahran FYI - this file will be lost in the docker image. You need to also update our internal training repo to add this to the saved model bundle to access it later.

lastmansleeping (commented):
@mohazahran Looks good to me. I will try to run this tomorrow and approve it. Meanwhile, try to take a look at these if you can:
- Fix scipy version
- Add unit tests for `ml4ir.applications.ranking.model.metrics.helpers.metrics_helper.py`
- Move default config loading to `Pipeline` class

Also, why are we not adding the non-inferiority t-tests?

python/ml4ir/base/config/eval_config.py
mohazahran: no, I'm using two sided test

lastmansleeping (approved):
LGTM

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -314,17 +199,25 @@ def evaluate(
...
             clicked_records = predictions_df[predictions_df[self.feature_config.get_label("node_name")] == 1.0]
             clicked_records[RankingConstants.NEW_MRR] = 1.0 / clicked_records["new_rank"]
             clicked_records[RankingConstants.OLD_MRR] = 1.0 / clicked_records[self.feature_config.get_rank("node_name")]
 
-            running_stats_df = clicked_records.groupby('organization_id').agg({
-                    RankingConstants.NEW_MRR: ['mean', 'var'],
-                    RankingConstants.OLD_MRR: ['mean', 'var'],
+            # computing batch-wise mean and variance per metric
+            metric_stat_df_list = []
+            clicked_records.set_index("organization_id")
```
balikasg: This too would break for any dataset that does not contain this column. It feels we are trying to do quite many things together in single framework, but I can not offer an easy solution out of this currently.

PR #214, opened 2023-02-13
Masking -inf, inf in rank match failure computation

mohazahran:
There was a bug where an inf value gets multiplied by zero resulting in a nan computation.

### DIFF
--- a/python/ml4ir/applications/ranking/tests/test_rank_match_failure.py
+++ b/python/ml4ir/applications/ranking/tests/test_rank_match_failure.py
@@ -70,6 +70,25 @@ class RankMachFailureTest(tf.test.TestCase):
         )
         assert np.isclose(actual_dcg, expected_dcg).all()
 
+    def test__mask_inf_scores(self):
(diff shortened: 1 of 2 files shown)


lastmansleeping (changes requested):
@mohazahran Great catch. Can you add a unit test to verify this functionality?

python/ml4ir/applications/ranking/tests/test_rank_match_failure.py
```diff
@@ -70,6 +70,25 @@ def test_discounted_cumulative_gain(self):
         )
         assert np.isclose(actual_dcg, expected_dcg).all()
 
+    def test__mask_inf_scores(self):
+        """Test masking inf, -inf for a given vector"""
+        tf_Vec = tf.constant(
```
lastmansleeping, line 75: [Minor] @mohazahran 
```suggestion
        tf_vec = tf.constant(
```

lastmansleeping (approved):
LGTM. Thanks @mohazahran. Just one cosmetic comment.

PR #220, opened 2023-02-22
Fix division by zero in evaluation

mohazahran:
Fix division by zero in compute_groupwise_running_variance_for_metrics

### DIFF
--- a/python/ml4ir/base/stats/t_test.py
+++ b/python/ml4ir/base/stats/t_test.py
@@ -400,7 +400,8 @@ def compute_groupwise_running_variance_for_metrics(metric_list, group_metric_run
             # update the current mean and variance
             # https://math.stackexchange.com/questions/2971315/how-do-i-combine-standard-deviations-of-two-groups
             if (sv.count + new_count) == 0:
-                # pypass the current batch
+                # skip the current batch when there is no new samples for the metric in the current batch.
+                # this can happen when the batch has bad queries
                 combine_count = sv.count
                 combined_mean = sv.mean
                 combine_var = sv.var
--- a/python/ml4ir/base/stats/t_test.py
+++ b/python/ml4ir/base/stats/t_test.py
@@ -399,13 +399,21 @@ def compute_groupwise_running_variance_for_metrics(metric_list, group_metric_run
 
             # update the current mean and variance
             # https://math.stackexchange.com/questions/2971315/how-do-i-combine-standard-deviations-of-two-groups
-            combined_mean = (sv.count * sv.mean + new_count * new_mean) / (sv.count + new_count)
-            combine_count = sv.count + new_count
-            if (sv.count + new_count - 1) != 0:
-                combine_var = (((sv.count-1) * sv.var + (new_count-1) * new_var) / (sv.count + new_count - 1)) + \
-                          ((sv.count * new_count * (sv.mean - new_mean)**2) / ((sv.count + new_count)*(sv.count + new_count - 1)))
+            if (sv.count + new_count) == 0:
+                # pypass the current batch
+                combine_count = sv.count
+                combined_mean = sv.mean
+                combine_var = sv.var
             else:
-                combine_var = 0
+                combined_mean = (sv.count * sv.mean + new_count * new_mean) / (sv.count + new_count)
+
+                combine_count = sv.count + new_count
+
+                if (sv.count + new_count - 1) != 0:
+                    combine_var = (((sv.count-1) * sv.var + (new_count-1) * new_var) / (sv.count + new_count - 1)) + \
+                              ((sv.count * new_count * (sv.mean - new_mean)**2) / ((sv.count + new_count)*(sv.count + new_count - 1)))
+                else:
+                    combine_var = 0
 
             sv.count = combine_count
             sv.mean = combined_mean
--- a/python/ml4ir/base/stats/t_test.py
+++ b/python/ml4ir/base/stats/t_test.py
@@ -402,23 +402,24 @@ def compute_groupwise_running_variance_for_metrics(metric_list, group_metric_run
             if (sv.count + new_count) == 0:
                 # skip the current batch when there is no new samples for the metric in the current batch.
                 # this can happen when the batch has bad queries
-                combine_count = sv.count
+                combined_count = sv.count
                 combined_mean = sv.mean
-                combine_var = sv.var
+                combined_var = sv.var
             else:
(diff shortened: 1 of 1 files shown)


python/ml4ir/base/stats/t_test.py
```diff
@@ -399,13 +399,21 @@ def compute_groupwise_running_variance_for_metrics(metric_list, group_metric_run
 
             # update the current mean and variance
             # https://math.stackexchange.com/questions/2971315/how-do-i-combine-standard-deviations-of-two-groups
-            combined_mean = (sv.count * sv.mean + new_count * new_mean) / (sv.count + new_count)
-            combine_count = sv.count + new_count
-            if (sv.count + new_count - 1) != 0:
-                combine_var = (((sv.count-1) * sv.var + (new_count-1) * new_var) / (sv.count + new_count - 1)) + \
-                          ((sv.count * new_count * (sv.mean - new_mean)**2) / ((sv.count + new_count)*(sv.count + new_count - 1)))
+            if (sv.count + new_count) == 0:
```
lastmansleeping, line 402: @mohazahran Can you add a comment on when this would happen?

python/ml4ir/base/stats/t_test.py
```diff
@@ -399,13 +399,21 @@ def compute_groupwise_running_variance_for_metrics(metric_list, group_metric_run
...
+                # pypass the current batch
```
lastmansleeping, line 403:
```suggestion
                # skip the current batch
```

python/ml4ir/base/stats/t_test.py
```diff
@@ -399,13 +399,21 @@ def compute_groupwise_running_variance_for_metrics(metric_list, group_metric_run
...
+                combine_count = sv.count
+                combined_mean = sv.mean
+                combine_var = sv.var
             else:
-                combine_var = 0
+                combined_mean = (sv.count * sv.mean + new_count * new_mean) / (sv.count + new_count)
```
lastmansleeping, line 408: @mohazahran Why not just add a small delta `1e-10` to count?

mohazahran: but wouldn't that possibly blow up the metric value. I feel skipping more safe.

python/ml4ir/base/stats/t_test.py
```diff
@@ -399,13 +399,22 @@ def compute_groupwise_running_variance_for_metrics(metric_list, group_metric_run
...
-            combine_count = sv.count + new_count
-            if (sv.count + new_count - 1) != 0:
-                combine_var = (((sv.count-1) * sv.var + (new_count-1) * new_var) / (sv.count + new_count - 1)) + \
-                          ((sv.count * new_count * (sv.mean - new_mean)**2) / ((sv.count + new_count)*(sv.count + new_count - 1)))
+            if (sv.count + new_count) == 0:
+                # skip the current batch when there is no new samples for the metric in the current batch.
+                # this can happen when the batch has bad queries
+                combine_count = sv.count
+                combined_mean = sv.mean
+                combine_var = sv.var
```
lastmansleeping, line 407: @mohazahran Use `combined_` everywhere for consistency. Instead of `combine_` I mean.

balikasg (changes requested):
Team, 
We have 400 lines of core statistical analysis in this file. To me this feels like a completely separate project on the problem of statistical significance with batches. Should we make the effort to separate and open-source? Also, code is copied from `penguin` as the docstrings imply. 
@mohazahran:  
- Why do we copy open source code? This is a bad practise. Is this a short-term thing? We should not be copying code like this.. 
- Why do we need batch analysis? Are the datasets that big that don't fit in memory? Using batch updates seems to complicate the process. Is it necessary? 
- Doesn't it make more sense to have this analysis once we write a predictions file? A small CLI with `penguin` as a dependency should be sufficient. The ABEval code for such analysis was also small.

python/ml4ir/base/stats/t_test.py
mohazahran: Actually you have a point. If the denominator is zero that would make the numerator zero too. In that case using delta to skipping would have similar effect.

lastmansleeping (approved):

lastmansleeping:
> Team, We have 400 lines of core statistical analysis in this file. To me this feels like a completely separate project on the problem of statistical significance with batches. Should we make the effort to separate and open-source? Also, code is copied from `penguin` as the docstrings imply. @mohazahran:
> 
> * Why do we copy open source code? This is a bad practise. Is this a short-term thing? We should not be copying code like this..

We had to because we do not have approval to use pinguoin internally due to license issue.

> * Why do we need batch analysis? Are the datasets that big that don't fit in memory? Using batch updates seems to complicate the process. Is it necessary?

We are dealing with data size of about 20GB on disk (which will be even more in memory)

> * Doesn't it make more sense to have this analysis once we write a predictions file? A small CLI with `penguin` as a dependency should be sufficient. The ABEval code for such analysis was also small.

I am starting to think so too. We can consider moving it to a separate spark job in the coming months as this is taking 6 hours to run `.evaluate()` on 20 GB of data. This is super slow for scalability.

balikasg:
@lastmansleeping thanks! 
20GB: 
are these 20GB of predictions on test? Billions or rows? Or all the info we log/put in that file make it that much? Anyway, I am impressed from this scale. Do you think we need that scale, ie, using smaller test datasets doesn't do the cut? 

Feel free to ignore my requested change..

mohazahran:
> @lastmansleeping thanks! 20GB: are these 20GB of predictions on test? Billions or rows? Or all the info we log/put in that file make it that much? Anyway, I am impressed from this scale. Do you think we need that scale, ie, using smaller test datasets doesn't do the cut?
> 
> Feel free to ignore my requested change..

@balikasg

Yes the new dataset collected has 20GB testset. We started the conversation of studying the min dataset requirement per group in order to see if we can reduce the testset size. We may have to set the eval pipeline as a separate entity for efficiency

lastmansleeping:
> are these 20GB of predictions on test? Billions or rows? Or all the info we log/put in that file make it that much?

This is the actual feature data for the test set. Not the predictions file. We have started discussing about data size thresholds both min and max and should adjust accordingly like @mohazahran mentioned.

mohazahran:
@balikasg Can you approve?

lastmansleeping:
@balikasg Can you approve this PR as it is blocking our model training efforts currently?

balikasg (approved):

balikasg:
Approved. I did not know that my review could not be dismissed, sorry!

PR #222, opened 2023-03-01
Upgrading to py39 and cleaning up requirements

lastmansleeping:
- [x] Upgrade minimum version from py37 to py39
- [x] Update circleci docker images to py39
- [x] Update dependencies to latest
- [x] Clean up `build-requirements.txt` and use single `requirements.txt` file

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,8 @@ jobs:
           name: Build
           command: |
             cd jvm
-            sudo mvn clean -Dmaven.test.skip=true install
+            sudo chmod -R 777 /home/circleci/
+            mvn clean -Dmaven.test.skip=true install
       - save_cache:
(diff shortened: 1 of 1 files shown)


python/requirements.txt
```diff
@@ -1,26 +1,21 @@
...
-# Pyspark for spark_io
-pyspark==3.0.1
-
-#calibration
-tensorflow-probability==0.15.0
-
-# DAG visualization
-pygraphviz==1.7
\ No newline at end of file
+testfixtures==6.18.5
```
FranciscoBorges, line 21: @lastmansleeping is this a test dependency?

FranciscoBorges (approved):
LGTM.

python/requirements.txt
lastmansleeping: Yes, it is used in one of the tests currently. Perhaps we should separate the tests to a different directory and isolate dependencies.

PR #224, opened 2023-03-07
Fixing load_weights() and moving model.build() to pipeline

lastmansleeping:
If `model.build(dataset)` is not called, then the keras model remains uncompiled. Adding it to `predict()`, `evaluate()` and `save()` allows us to use these methods directly without calling `fit()` first. Useful for reloading pretrained models.

### DIFF
--- a/python/ml4ir/applications/ranking/tests/test_ranking_model.py
+++ b/python/ml4ir/applications/ranking/tests/test_ranking_model.py
@@ -129,7 +129,7 @@ class RankingModelTest(RankingTestBase):
             file_io=self.file_io,
             logger=self.logger,
         )
-        ranking_model.build(ranking_dataset.train)
+        ranking_model.build(ranking_dataset)
 
         # Save the model and check if coefficients file was saved
         ranking_model.save(models_dir=self.args.models_dir)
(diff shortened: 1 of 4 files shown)


mohazahran (changes requested):
@lastmansleeping 
I did an on fly QA trying to re-save one of the global models I got this 

`!!! Error Training Model: !!!
Traceback (most recent call last):
  File "ml4ir/base/pipeline.py", line 535, in run_pipeline
    relevance_model.load_weights(self.args.model_file)
  File "ml4ir/base/model/relevance_model.py", line 767, in load_weights
    self.model.set_weights([0., 0., *loaded_model.get_weights()])
  File "envs/ml4ir_new/lib/python3.7/site-packages/keras/engine/base_layer.py", line 1628, in set_weights
    (self.name, len(weights), expected_num_weights, str(weights)[:50]))
ValueError: You called `set_weights(weights)` on layer "relevance_scorer" with a weight list of length 35, but the layer was expecting 29 weights. Provided weights: [0.0, 0.0, array([[-0.03738469,  0.00727513, -0.02...
`
did you try to save one yourself?

mohazahran (approved):
LGTM

PR #225, opened 2023-03-07
Fixing QueryNormalization save

lastmansleeping:
Fixes a consistency issue with the results seen when `QueryNormalization` layer is saved and reloaded from disk.

### DIFF
--- a/python/ml4ir/base/model/relevance_model.py
+++ b/python/ml4ir/base/model/relevance_model.py
@@ -116,6 +116,7 @@ class RelevanceModel:
             """
             self.model: Model = self.load(model_file)
             self.is_compiled = False
+            self.is_built = False
         else:
 
             """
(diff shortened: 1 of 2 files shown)


python/ml4ir/applications/ranking/model/layers/normalization.py
```diff
@@ -48,17 +47,20 @@ def call(self, inputs, mask=None, training=None):
...
+        mean = tf.math.divide(tf.math.reduce_sum(inputs, axis=1), count)
+        mean = tf.multiply(tf.expand_dims(mean, axis=1), mask)
 
-        var = tf.math.reduce_sum(tf.math.pow((inputs - mean), 2), axis=1) / count
-        std = tf.math.sqrt(var)[:, tf.newaxis, :] + QueryNormalization.DELTA
+        var = tf.math.divide(tf.math.reduce_sum(tf.math.pow(tf.math.subtract(inputs, mean), 2), axis=1), count)
+        std = tf.expand_dims(tf.math.sqrt(var), axis=1)
 
-        zscore = (inputs - mean) / std
+        zscore = tf.math.divide_no_nan(tf.math.subtract(inputs, mean), std)
```
lastmansleeping, line 64: These changes are more quality of life as I am paranoid about save-load parity now. They do not make a difference from my experiments.

python/ml4ir/applications/ranking/model/layers/normalization.py
```diff
@@ -48,17 +47,20 @@ def call(self, inputs, mask=None, training=None):
             Normed input feature tensor
             Shape: [batch_size, sequence_len, encoding_size]
         """
-        mask = tf.cast(mask, inputs.dtype)[:, :, tf.newaxis]
+        # NOTE: Tensorflow math ops are not consistent on saving and reloading if they are float32.
+        #       Both float16 and float64 give consistent results.
+        inputs = tf.cast(inputs, tf.float64)
```
lastmansleeping, line 52: This is the key change

PR #237, opened 2023-06-22
Adding StringMultiLabelProcessor

lastmansleeping:


python/ml4ir/base/features/feature_fns/label_processor.py
```diff
@@ -0,0 +1,108 @@
+import tensorflow as tf
+
+from ml4ir.base.features.feature_fns.base import BaseFeatureLayerOp
+from ml4ir.base.io.file_io import FileIO
+
+
+class StringMultiLabelProcessor(BaseFeatureLayerOp):
```
balikasg, line 7: Nice code! 
- Do you plan to add a test on this? 
- Another suggestion is to improve a bit the Example in the docstring

lastmansleeping: Yes, to both of those. WIP...

lastmansleeping: It was a placeholder for me to go back and add an example.

mohazahran (commented):
This is awesome. 

One comment I have is the order of the labels. We have no automatic way to know the order of the labels (label1-label2-label3) corresponding to the weights. But I think there is no "easy" way to set that on the top of my mind.

lastmansleeping:
> This is awesome.
> 
> One comment I have is the order of the labels. We have no automatic way to know the order of the labels (label1-label2-label3) corresponding to the weights. But I think there is no "easy" way to set that on the top of my mind.

Replied offline on slack that we are specifying this info in the column name as a workaround for now.

mohazahran (approved):
LGTM

PR #238, opened 2023-06-23
Updating ranking metrics for graded relevance and ties

lastmansleeping:


python/ml4ir/applications/ranking/model/metrics/metrics_impl.py
```diff
@@ -33,18 +33,22 @@ def update_state(self, y_true, y_pred, mask=None, sample_weight=None):
...
 
-        # Fetch indices of clicked records from y_true
-        y_true_clicks = tf.where(tf.equal(tf.cast(y_true, tf.int32), tf.constant(1)))
+        # Fetch highest relevance grade from the y_true labels
+        y_true_clicks = tf.cast(tf.equal(y_true, tf.math.reduce_max(y_true, axis=-1)[:, tf.newaxis]), tf.float32)
 
         # Compute rank of clicked record from predictions
-        click_ranks = tf.gather_nd(y_pred_ranks, indices=y_true_clicks)
+        # Break ties in case of multiple target click labels using the min rank from y_pred_ranks
+        click_ranks = tf.reduce_min(tf.divide(y_pred_ranks, y_true_clicks), axis=-1)
```
lastmansleeping, line 51: @VenkataKarthikP This is the key change. Instead of checking if the label is 1 or 0, now we get the max of the graded labels. This will automatically work already for 1s and 0s too as you can see from the unchanged tests.

lastmansleeping: @mohazahran FYI

PR #242, opened 2023-07-01
Update t_test.py

mohazahran:
for high d values the call to power_ttest will invoke the c++ boost lib that will throw an exception (below) that will not be caught here.
         We can temporeary circumvent this behavior by saturating the d value at 4 (for statistical_power == 0.9 and pvalue == 0.1) and return
         a pre-computed required sample size for this value.

### DIFF
--- a/python/ml4ir/base/stats/t_test.py
+++ b/python/ml4ir/base/stats/t_test.py
@@ -458,6 +458,16 @@ def compute_required_sample_size(mean1, mean2, var1, var2, statistical_power, pv
         if denominator == 0:
             return np.inf
         d = np.abs(float(mean1) - float(mean2)) / denominator
+        '''
+         for high d values the call to power_ttest will invoke the c++ boost lib that will throw an exception (below) that will not be caught here.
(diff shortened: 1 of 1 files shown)


python/ml4ir/base/stats/t_test.py
```diff
@@ -458,6 +458,16 @@ def compute_required_sample_size(mean1, mean2, var1, var2, statistical_power, pv
         if denominator == 0:
             return np.inf
         d = np.abs(float(mean1) - float(mean2)) / denominator
+        '''
```
balikasg, line 461: Minor Questions: 
- Should we be using `"` instead of `'` as in the docstring? 
- There seems to be an extra indent in the details compared to the `'''`
- Please start with capital (`For`)
- Typos: `temporeary`

Main questions: 
- How is 2.24 calculated? 
- What's the plan to actually fix the issue and remove this temporary fix? Is there one? 
- power==0.9 and pvalue==0.1 are very specifix. Should we have something more generic e.g., `power>=0.9` and `pvalue<=0.1`?    

Thanks for catching this and trying to fix it here. Is there already an open issue on `power_ttest`?

PR #245, opened 2023-09-13
Adding theoretical min max norm

lastmansleeping:
PR to add theoretical_min-max normalization for ranking features.

### DIFF
--- a/python/ml4ir/applications/ranking/features/feature_fns/normalization.py
+++ b/python/ml4ir/applications/ranking/features/feature_fns/normalization.py
@@ -69,4 +69,4 @@ class TheoreticalMinMaxNormalization(BaseFeatureLayerOp):
             tf.math.subtract(inputs, self.theoretical_min),
             tf.math.subtract(query_max, self.theoretical_min))
 
-        return normed_inputs
+        return tf.expand_dims(normed_inputs, axis=-1)
--- a/python/ml4ir/applications/ranking/features/feature_fns/normalization.py
+++ b/python/ml4ir/applications/ranking/features/feature_fns/normalization.py
@@ -1,0 +1,72 @@
+import tensorflow as tf
+
+from ml4ir.base.features.feature_fns.base import BaseFeatureLayerOp
+from ml4ir.base.io.file_io import FileIO
+
+
+class TheoreticalMinMaxNormalization(BaseFeatureLayerOp):
+    """
+    Min Max Normalization of individual query features,
+    where the theoretical min is used instead of the minimum.
+
+    Reference -> An Analysis of Fusion Functions for Hybrid Retrieval
+                 https://arxiv.org/abs/2210.11934
+    """
+    LAYER_NAME = "theoretical_min_max_norm"
+
+    THEORETICAL_MIN = "theoretical_min"
+
+
+    def __init__(self, feature_info: dict, file_io: FileIO, **kwargs):
+        """
+        Initialize layer to define a theoretical min max normalization
+
+        Parameters
+        ----------
+        feature_info : dict
+            Dictionary representing the configuration parameters for the specific feature from the FeatureConfig
+        file_io : FileIO object
+            FileIO handler object for reading and writing
+
+        Notes
+        -----
+        Args under feature_layer_info:
+            theoretical_min : float
+                Theoretical minimum to use for the query's record features
+                Default value of 0. is used if not specified.
+        """
+        super().__init__(feature_info=feature_info, file_io=file_io, **kwargs)
+
+        self.theoretical_min = self.feature_layer_args.get(self.THEORETICAL_MIN, 0.)
+
+
+    def call(self, inputs, training=None):
+        """
+        Defines the forward pass for the layer on the inputs tensor
+
+        Parameters
+        ----------
+        inputs: tensor
+            Input tensor on which the feature transforms are applied
+        training: boolean
+            Boolean flag indicating if the layer is being used in training mode or not
+
+        Returns
+        -------
+        tf.Tensor
+            Resulting tensor after the forward pass through the feature transform layer
+        """
+        # Replace values lower than theoretical minimum with theoretical minimum
+        inputs = tf.clip_by_value(inputs,
+                                  clip_value_min=self.theoretical_min,
+                                  clip_value_max=tf.reduce_max(inputs))
+
+        # Compute max values for each query
+        query_max = tf.expand_dims(tf.reduce_max(inputs, axis=-1), axis=-1)
+
+        # Min max normalization
+        normed_inputs = tf.math.divide_no_nan(
+            tf.math.subtract(inputs, self.theoretical_min),
+            tf.math.subtract(query_max, self.theoretical_min))
+
+        return normed_inputs
--- a/python/ml4ir/applications/ranking/tests/test_feature_fns.py
+++ b/python/ml4ir/applications/ranking/tests/test_feature_fns.py
@@ -105,7 +105,7 @@ class FeatureLayerTest(RelevanceTestBase):
             feature_info, self.file_io
         )(input_feature).numpy()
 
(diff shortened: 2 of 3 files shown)


python/ml4ir/applications/ranking/features/feature_fns/normalization.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        # Replace values lower than theoretical minimum with theoretical minimum
+        inputs = tf.clip_by_value(inputs,
+                                  clip_value_min=self.theoretical_min,
+                                  clip_value_max=tf.reduce_max(inputs))
+
+        # Compute max values for each query
+        query_max = tf.expand_dims(tf.reduce_max(inputs, axis=-1), axis=-1)
+
+        # Min max normalization
```
ArmandGiraud, line 67: For min-max normalization you need to train the normalizer on your train dataset. Using a batch max value is not suitable for normalization. you need to write the max somewhere and read it when loading the normalizer. 
This is one reason to prefer Reciprocal Ranking that does not need any normalization.

ArmandGiraud (changes requested):
Max should come from training distribution, not in sample estimate.

lastmansleeping:
> Max should come from training distribution, not in sample estimate.

But this is not binary classification, we are trying to sort a list. The normalization is at the per-query level.

Here is a description from a recent paper -> https://arxiv.org/pdf/2210.11934.pdf

python/ml4ir/applications/ranking/features/feature_fns/normalization.py
lastmansleeping: Not really, we are interested in only a per-query normalizer here.

ArmandGiraud: Indeed, in the paper they use per-query normalization. Something I don't understand is that this would give a score of 1 to the most relevant document of each result set. we are losing info of absolute similarities provided by the retrievers.  
it would be good to test whether using a p99 instead of min and max. It would mechanically provide the reranker robust information on which retriever models to favor rather than a rough estimate based on - for example - query length.

python/ml4ir/applications/ranking/features/feature_fns/normalization.py
```diff
@@ -0,0 +1,72 @@
+import tensorflow as tf
+
+from ml4ir.base.features.feature_fns.base import BaseFeatureLayerOp
+from ml4ir.base.io.file_io import FileIO
+
+
+class TheoreticalMinMaxNormalization(BaseFeatureLayerOp):
+    """
+    Min Max Normalization of individual query features,
```
ArmandGiraud, line 9:
```suggestion
    Min Max Scaling of individual query features,
```

python/ml4ir/applications/ranking/features/feature_fns/normalization.py
```diff
@@ -0,0 +1,72 @@
...
+                 https://arxiv.org/abs/2210.11934
+    """
+    LAYER_NAME = "theoretical_min_max_norm"
+
+    THEORETICAL_MIN = "theoretical_min"
+
+
+    def __init__(self, feature_info: dict, file_io: FileIO, **kwargs):
+        """
+        Initialize layer to define a theoretical min max normalization
```
ArmandGiraud, line 22:
```suggestion
        Initialize layer to define min max scaling
```

python/ml4ir/applications/ranking/features/feature_fns/normalization.py
```diff
@@ -0,0 +1,72 @@
...
+            Input tensor on which the feature transforms are applied
+        training: boolean
+            Boolean flag indicating if the layer is being used in training mode or not
+
+        Returns
+        -------
+        tf.Tensor
+            Resulting tensor after the forward pass through the feature transform layer
+        """
+        # Replace values lower than theoretical minimum with theoretical minimum
```
ArmandGiraud, line 59: Not sure how safe this is, probably raise an exception in this case, if this case occurs there is something wrong with the theoretical min

python/ml4ir/applications/ranking/tests/test_feature_fns.py
```diff
@@ -90,4 +90,37 @@ def test_sequence_categorical_vector_one_hot(self):
...
+        self.assertTrue(np.isclose(actual_one_hot, expected_one_hot).all())
+
+    def test_theoretical_min_max_norm(self):
+        """
+        Test TheoreticalMinMaxNormalization feature transform
+        """
+        feature_info = copy.deepcopy(FEATURE_INFO)
+        feature_info["feature_layer_info"]["args"]["theoretical_min"] = 0.5
+
+        input_feature = np.array([[-0.2, -0.1, 0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0]])
```
ArmandGiraud, line 102: should raise an exception

ArmandGiraud: or a warning.

python/ml4ir/applications/ranking/tests/test_feature_fns.py
```diff
@@ -90,4 +90,37 @@ def test_sequence_categorical_vector_one_hot(self):
...
+        """
+        feature_info = copy.deepcopy(FEATURE_INFO)
+
+        input_feature = np.array([[-0.2, -0.1, 0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0]])
+
+        actual_normed_feature = normalization.TheoreticalMinMaxNormalization(
+            feature_info, self.file_io
+        )(input_feature).numpy()
+
+        expected_normed_feature = np.array([[0., 0., 0., 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.]])[:, :, np.newaxis]
```
ArmandGiraud, line 124: nit:
```suggestion
        expected_normed_feature = np.array([[0., 0., 0., 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.]]).reshape(1, -1, 1)
```

python/ml4ir/applications/ranking/features/feature_fns/normalization.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def call(self, inputs, training=None):
+        """
+        Defines the forward pass for the layer on the inputs tensor
+
+        Parameters
+        ----------
+        inputs: tensor
+            Input tensor on which the feature transforms are applied
+        training: boolean
```
ArmandGiraud, line 51: not used here

lastmansleeping: Tensorflow expects this function definition I believe.

python/ml4ir/applications/ranking/features/feature_fns/normalization.py
lastmansleeping: > we are losing info of absolute similarities provided by the retrievers.
Yes, this is a trade off we have to do to deal with changing dense indices, unfortunately. Otherwise the model would expect a score of 0.8, say, and when it sees 0.9, it would be wrongly interpreting it.

python/ml4ir/applications/ranking/features/feature_fns/normalization.py
lastmansleeping: "Normalization" is still the correct usage, right?

python/ml4ir/applications/ranking/features/feature_fns/normalization.py
lastmansleeping: Same as above.

python/ml4ir/applications/ranking/features/feature_fns/normalization.py
ArmandGiraud: > This does happen quite often though

What do you mean? how can cosine be lower than -1 ? sparse scores should not go below zero as well ?

PR #246, opened 2023-09-18
Adding query_len feature transformation

lastmansleeping:
PR adds ability to transform query string feature to number of tokens or characters


python/ml4ir/applications/ranking/features/feature_fns/string.py
```diff
@@ -0,0 +1,65 @@
...
+        ----------
+        feature_info : dict
+            Dictionary representing the configuration parameters for the specific feature from the FeatureConfig
+        file_io : FileIO object
+            FileIO handler object for reading and writing
+
+        Notes
+        -----
+        Args under feature_layer_info:
+            tokenize: boolean
```
ArmandGiraud, line 30: we probably want to provide this function an arbitrary tokenizer, in case we want to add support for other languages later.

python/ml4ir/applications/ranking/features/feature_fns/string.py
```diff
@@ -0,0 +1,65 @@
...
+
+from ml4ir.base.features.feature_fns.base import BaseFeatureLayerOp
+from ml4ir.base.io.file_io import FileIO
+
+
+class QueryLength(BaseFeatureLayerOp):
+    """
+    Compute the length of the query string context feature
+    """
+    LAYER_NAME = "query_len"
```
ArmandGiraud, line 11: make it more readable
```suggestion
    LAYER_NAME = "query_length"
```

python/ml4ir/applications/ranking/features/feature_fns/string.py
lastmansleeping: That is true. I'll add a note.

PR #254, opened 2024-01-24
Adding sentence transformer layer

lastmansleeping:
Adding `SentenceTransformerWithTokenizer` layer and feature transform which allows us to model sentence embeddings as part of the tensorflow graph along with the tokenizer.

### DIFF
--- a/python/ml4ir/base/model/layers/sentence_transformers.py
+++ b/python/ml4ir/base/model/layers/sentence_transformers.py
@@ -1,7 +1,6 @@
 import tensorflow as tf
 from tensorflow.keras import layers
-from transformers import TFAutoModel, TFBertTokenizer, AutoTokenizer
-from sentence_transformers import SentenceTransformer
+from transformers import TFAutoModel, TFBertTokenizer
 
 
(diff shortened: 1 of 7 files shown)


mohazahran (commented):
Looks good. Thanks Ashish. it maybe worth adding a test case using a different model than the e5

python/ml4ir/base/model/layers/sentence_transformers.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # Define tokenizer as part of the tensorflow graph
+        self.tokenizer = TFBertTokenizer.from_pretrained(self.model_name_or_path, **kwargs)
+
+        self.transformer_model = None
+        self.dense = None
+        self.apply_dense = False
+        self.normalize_embeddings = False
+        self.pool_embeddings = False
+        for module in self.modules:
```
mohazahran, line 77: Is that's the fix to the issue of loading different models? @lastmansleeping I'm curious what was the issue

mohazahran (approved):
@lastmansleeping I left a comment on how these new additions fixes the issue, but not blocking to merge.

PR #257, opened 2024-01-25
Adding data_compression arg for classification

lastmansleeping:


python/ml4ir/applications/classification/pipeline.py
```diff
@@ -109,6 +109,7 @@ def get_relevance_dataset(
         relevance_dataset = RelevanceDataset(
             data_dir=self.data_dir_local,
             data_format=self.data_format,
+            data_compression=self.args.data_compression,
```
marjanhs, line 112: some checks [here](https://github.com/salesforce/ml4ir/blob/master/python/ml4ir/applications/classification/tests/test_base.py#L90) can improve the testing . thoughts ?

lastmansleeping: Can you give me an example? I am not sure what checks you mean.

marjanhs (approved):
LGTM

PR #258, opened 2024-01-26
Setting urllib version and refactoring robustscaler tests

lastmansleeping:


python/ml4ir/base/tests/test_robust_scaler_transform.py
```diff
@@ -33,7 +33,7 @@ def test_reciprocal_rank_negative_scores(self):
 
         self.assertTrue(np.isclose(actual_reciprocal_ranks, expected_reciprocal_ranks).all())
 
-    def test_reciprocal_rank_zeros(self):
+    def test_robust_scaler_zeros(self):
```
marjanhs, line 36: what is the reason behind renaming ?

python/requirements.txt
```diff
@@ -4,6 +4,7 @@ numpy==1.24.2
 pandas==1.5.3
 scipy==1.10.0
 PyYAML==5.4.1
+urllib3<2.0  # Issue -> https://stackoverflow.com/questions/76187256/importerror-urllib3-v2-0-only-supports-openssl-1-1-1-currently-the-ssl-modu
```
marjanhs, line 7: do you now why did not have it before ?

lastmansleeping: I think the issue is coming from internal PyPi. Someone probably uploaded urllib3>2.0 recently and now it is auto installing that latest version.

lastmansleeping: No, it was installing 2.1.0 earlier too. I am confused now as well.

python/ml4ir/base/tests/test_robust_scaler_transform.py
lastmansleeping: I think the naming was just wrong earlier. Probably a copy-paste error. Can you confirm @mohazahran ?

marjanhs (approved):

PR #262, opened 2024-01-31
update local_io

marjanhs:
update local_io

### DIFF
--- a/python/ml4ir/base/io/local_io.py
+++ b/python/ml4ir/base/io/local_io.py
@@ -89,6 +89,7 @@ class LocalIO(FileIO):
                 error_bad_lines=False,
                 warn_bad_lines=True,
                 engine="c",
+                na_filter=False
             )
 
         except Exception as e:


lastmansleeping (changes requested):
@marjanhs Can you add a comment describing the issue in the PR and in the code for future context?

python/ml4ir/base/io/local_io.py
```diff
@@ -58,6 +59,8 @@ def read_df(
             separator to use for loading file
         index_col : int, optional
             column to be used as index
+        na_filter: bool
```
lastmansleeping, line 62: @marjanhs Can you add a note about why the default is False?

python/ml4ir/base/io/local_io.py
```diff
@@ -118,6 +123,8 @@ def read_df_list(self, infiles, sep=",", index_col=None, **kwargs) -> pd.DataFra
...
```
lastmansleeping, line 126: @marjanhs same as above

lastmansleeping (approved):
LGTM. Thanks @marjanhs

PR #263, opened 2024-02-02
Upgrading tensorflow-core to 0.5.0

lastmansleeping:


lastmansleeping:
We can not upgrade tensorflow to 2.11 as it throws a DenseTInt32 to TFloat32 cast error when model is loaded in JVM. Also, internally we only support scala 11 and hence can not use tensorflow-core==0.5.0

PR #265, opened 2024-04-15
Mo/monte carlo dropout

mohazahran:


python/ml4ir/base/config/parse_args.py
```diff
@@ -401,6 +401,16 @@ def define_args(self):
...
 
+        self.add_argument(
+            "--monte_carlo_inference_trials",
+            type=int,
+            default=0,
+            help="""Adding stochastisity at inference time. Monte Carlo trials is the number of times inference will be run while using dropout during inference.
+            The final output scores will be the aggregate sum of scores produced for each monte carlo trial (run).
+            This argument should be used only if there is a non zero dropout probability"""
+
+        )
```
lastmansleeping, line 412: @mohazahran might be better to just roll this into the model config, no?

lastmansleeping: will also help with the deterministic follow up.

python/ml4ir/base/model/scoring/scoring_model.py
```diff
@@ -381,6 +388,9 @@ def test_step(self, data):
             y = self.interaction_model.label_transform_op(y, training=False)
 
         y_pred = self(X, training=False)[self.output_name]
+        for _ in range(self.monte_carlo_inference_trials):  # MC trials during test.
+            y_pred += self(X, training=True)[self.output_name]
+        y_pred /= (self.monte_carlo_inference_trials+1)
```
lastmansleeping, line 393: @mohazahran we should move this change out to a different Layer or wrapper. I assume this is a first pass direct port of your spike code?

lastmansleeping (changes requested):
@mohazahran Functionally, you want to mask vertically instead of horizontally. See - https://salesforce-internal.slack.com/archives/C02D71EJPRR/p1714757570802159?thread_ts=1713889103.049179&cid=C02D71EJPRR

More importantly, how can we add this as a layer without requiring a model config update with mask. I propose creating a new layer, which you attach, say, before the first dense or setrank layers. Then, once you get predictions, you collapse/merge somehow. You shouldn't be modifying the Keras fit and predict subclass functions for this IMO. Those are more generic.

python/ml4ir/base/config/parse_args.py
mohazahran: I think having it either way is fine.

python/ml4ir/base/model/scoring/Monte_Carlo_Scorer.py
```diff
@@ -0,0 +1,285 @@
...
+
+        # Apply architecture op on train_features
+        features[FeatureTypeKey.LOGITS] = self.architecture_op(features, training=training)
+
+        # Apply final activation layer
+        scores = self.loss_op.final_activation_op(features, training=training)
+
+        return scores
+
+    def train_step(self, data):
```
lastmansleeping, line 217: @mohazahran this is the only method that should be overriden.

python/ml4ir/base/model/scoring/Monte_Carlo_Scorer.py
```diff
@@ -0,0 +1,285 @@
...
+        # Update weights
+        self.optimizer.apply_gradients(zip(gradients, self.trainable_variables))
+
+        # Update metrics
+        self.__update_metrics(inputs=X, y_true=y, y_pred=y_pred)
+
+        # Return a dict mapping metric names to current value
+        return {metric.name: metric.result() for metric in self.metrics}
+
+    def test_step(self, data):
```
lastmansleeping, line 255: @mohazahran this is the only method that should be overriden.

python/ml4ir/base/model/scoring/Monte_Carlo_Scorer.py
```diff
@@ -0,0 +1,285 @@
+
```
lastmansleeping, line 1: @mohazahran lower case the python module name.

python/ml4ir/base/pipeline.py
```diff
@@ -396,19 +397,36 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+                interaction_model=interaction_model,
+                loss=loss,
+                aux_loss=aux_loss,
+                aux_loss_weight=self.args.aux_loss_weight,
+                aux_metrics=aux_metrics,
+                output_name=self.args.output_name,
+                logger=self.logger,
+                file_io=self.file_io,
+                logs_dir=self.logs_dir_local
+            )
```
lastmansleeping, line 429: @mohazahran you can move this to a scorer factory python module to keep this clean - https://www.geeksforgeeks.org/factory-method-for-designing-pattern/

If you move the `monte_carlo_inference_trials` into the model config itself, then you can reduce the code duplication quite a bit.

python/ml4ir/base/model/scoring/prediction_helper.py
```diff
@@ -86,10 +89,23 @@ def _filter_records(x, mask):
...
+                scores += s
+            scores /= (monte_carlo_inference_trials+1)
         else:
+            # inference with no dropout
             scores = infer(**features)[output_name]
+            # running monte carlo iterations
+            for _ in range(monte_carlo_inference_trials):
+                s = infer(**features, training=True)[output_name]
+                scores += s
+            scores /= (monte_carlo_inference_trials+1)
```
lastmansleeping, line 108: @mohazahran I am not sure why this is needed. The way it is implemented the MC trials are already part of the `infer`. Now you are doing a double run as each `infer` will run for `monte_carlo_inference_trials`

python/ml4ir/base/config/parse_args.py
```diff
@@ -401,6 +401,16 @@ def define_args(self):
             help="The number of folds for K-fold Cross Validation. Must be > 2 if testset is included in folds and > 1 otherwise.",
         )
 
+        self.add_argument(
```
lastmansleeping, line 404: @mohazahran I recommend moving this to a model config key as it reduces code duplication a lot. Thoughts?

python/ml4ir/applications/ranking/model/ranking_model.py
```diff
@@ -167,6 +170,7 @@ def evaluate(
                 features_to_return=evaluation_features,
                 additional_features=additional_features,
                 max_sequence_size=self.max_sequence_size,
+                monte_carlo_inference_trials=monte_carlo_inference_trials
```
lastmansleeping, line 173: @mohazahran same comment i left elsewhere - i don't think you need to do this if you are using MC scorer.

python/ml4ir/applications/ranking/model/layers/masking.py
```diff
@@ -98,3 +98,98 @@ def call(self, inputs, mask=None, training=None):
...
+
+        Returns
+        -------
+        Tensor object
+            Feature tensor where mask_rate of records' features have been masked out (or set to 0.)
+            Shape: [batch_size, sequence_len, feature_dim]
+        """
+        batch_size, sequence_len, feature_dim = inputs.shape
+        if self.mask_at_inference or training:
+            mask = tf.cast(tf.random.uniform([batch_size, feature_dim]) > self.mask_rate, dtype=tf.float32)
```
lastmansleeping, line 188: @mohazahran sometimes, tensorflow saves these as constants instead of layers. Worth double checking.

python/ml4ir/applications/ranking/model/layers/masking.py
```diff
@@ -98,3 +98,98 @@ def call(self, inputs, mask=None, training=None):
...
+            mask = tf.expand_dims(mask, axis=1)
+            mask = tf.tile(mask, [1, sequence_len, 1])
+            return inputs * mask
```
lastmansleeping, line 191: @mohazahran Better to use tf native instead of `*`

python/ml4ir/base/model/relevance_model.py
```diff
@@ -605,6 +606,9 @@ def evaluate(
             Value representing how often(in batches) to log status
         compute_intermediate_stats : bool
             Determines if group metrics and other intermediate stats on the test set should be computed
+        monte_carlo_inference_trials: int
```
lastmansleeping, line 609: @mohazahran mentioned somewhere else to move to model config to keep this clean.

python/ml4ir/base/pipeline.py
lastmansleeping: You can have some kind of a `get_scorer` function under `model/scoring/` and use that everywhere in the code.

lastmansleeping (changes requested):
@mohazahran Love it. Left some refactoring comments.

python/ml4ir/applications/ranking/model/layers/masking.py
lastmansleeping: @mohazahran can you change the `>` as well with something like https://www.tensorflow.org/api_docs/python/tf/math/greater

lastmansleeping: @mohazahran can you change the > as well with something like https://www.tensorflow.org/api_docs/python/tf/math/greater

python/ml4ir/base/model/scoring/Monte_Carlo_Scorer.py
lastmansleeping: ping @mohazahran to `monte_carlo_scorer.py`

python/ml4ir/base/model/scoring/Monte_Carlo_Scorer.py
```diff
@@ -0,0 +1,131 @@
...
+from ml4ir.base.model.architectures import architecture_factory
+from ml4ir.base.model.losses.loss_base import RelevanceLossBase
+from ml4ir.base.model.scoring.interaction_model import InteractionModel
+from ml4ir.base.model.scoring.scoring_model import RelevanceScorer
+
+
+class MonteCarloScorer(RelevanceScorer):
+    """
+    Monte Carlo class that defines the neural network layers that convert
+    the input features into scores by using monte carlo trials in inference.
```
lastmansleeping, line 23: @mohazahran worth explicitly calling out that the masking of the features is not done in this and must be done via model config using either dropout, or one of the mask layers available

python/ml4ir/base/model/scoring/Monte_Carlo_Scorer.py
```diff
@@ -0,0 +1,131 @@
...
+        # Apply feature layer and transform inputs
+        features = self.interaction_model(inputs, training=training)
+
+        # Apply architecture op on train_features
+        features[FeatureTypeKey.LOGITS] = self.architecture_op(features, training=training)
+
+        # Apply final activation layer
+        scores = self.loss_op.final_activation_op(features, training=training)
+
+        return scores
```
lastmansleeping, line 131: @mohazahran I believe I had a comment on replacing `make_pred` with `super().call(...)` as it is a duplicate. Did you get a chance to try?

python/ml4ir/base/model/scoring/Monte_Carlo_Scorer.py
```diff
@@ -0,0 +1,131 @@
...
+        inputs : dict of tensors
+            Dictionary of input feature tensors
+
+        Returns
+        -------
+        scores : dict of tensor object
+            Tensor object of the score computed by the model
+        """
+        scores = self.make_pred(inputs, training=False)
+        for _ in range(self.monte_carlo_inference_trials):
```
lastmansleeping, line 116: @mohazahran leave a TODO comment for optimizing in future with vectorized ops instead of for loop

python/ml4ir/base/model/scoring/Monte_Carlo_Scorer.py
```diff
@@ -0,0 +1,131 @@
...
+            scores += self.make_pred(inputs, training=True)
+        scores = tf.divide(scores, tf.constant(self.monte_carlo_inference_trials + 1, dtype=tf.float32))
```
lastmansleeping, line 118: @mohazahran you should move the `tf.constant` to the constructor as it will always be the same.

python/ml4ir/base/pipeline.py
```diff
@@ -396,19 +397,35 @@ def get_relevance_model(self, feature_layer_keys_to_fns={}) -> RelevanceModel:
...
+        else:
+            scorer: RelevanceScorer = RelevanceScorer(
+                feature_config=self.feature_config,
+                model_config=self.model_config,
+                interaction_model=interaction_model,
+                loss=loss,
+                aux_loss=aux_loss,
+                aux_loss_weight=self.args.aux_loss_weight,
+                aux_metrics=aux_metrics,
```
lastmansleeping, line 423: @mohazahran I had a comment about moving this to a separate file under `scorer/` like `scorer_factory.py` which has a `get_scorer` method to keep this pipeline clean. We've bloated it quite a bit already. Want to avoid further.

python/ml4ir/base/tests/data/configs/model_config_monte_carlo_2.yaml
```diff
@@ -0,0 +1,34 @@
+architecture_key: auto-dag-network
```
lastmansleeping, line 1: @mohazahran rename file with actual trials value

python/ml4ir/base/tests/data/configs/model_config_monte_carlo_1.yaml
```diff
@@ -0,0 +1,34 @@
+architecture_key: auto-dag-network
```
lastmansleeping, line 1: @mohazahran rename file with actual trials value

python/ml4ir/base/tests/test_monte_carlo_scorer.py
```diff
@@ -0,0 +1,97 @@
+import unittest
```
lastmansleeping, line 1: @mohazahran sorry, let's not add more end-to-end tests, it is really slowing the build and tf.keras models are leading to memory leaks. Can you test just MC scorer by mocking everything else out using number of calls made to `call()` or `make_pred`?

lastmansleeping (changes requested):
@mohazahran Left some comments. Main one being, removing end to end test for MC scorer and using a unit test. It is leading to lot of memory leaks in autobuilds. :/

python/ml4ir/base/model/scoring/Monte_Carlo_Scorer.py
```diff
@@ -0,0 +1,131 @@
...
+                         loss=loss,
+                         aux_loss=aux_loss,
+                         aux_loss_weight=aux_loss_weight,
+                         aux_metrics=aux_metrics,
+                         output_name=output_name,
+                         logger=logger,
+                         file_io=file_io,
+                         logs_dir=logs_dir,
+                         **kwargs)
+        self.monte_carlo_inference_trials = self.model_config["monte_carlo_inference_trials"]["value"]
```
lastmansleeping, line 99: @mohazahran Can you change value to `num_trials`? You can modify main key to `monte_carlo_inference` to keep it generic for future in case.

Also, move them to `keys.py` maybe and use constants here instead of strings?

python/ml4ir/base/tests/test_monte_carlo_scorer.py
mohazahran: @lastmansleeping the scorer need to init everything else basically. So the only part we will be saving is creating the dataset, which we can keep it to minimum.

mohazahran: The dataset is just 10 queries with two features and the model have 8 neurons.

lastmansleeping (changes requested):
@mohazahran I am going to be a pain and ask you to replace the end to end tests with a unit test for MC Scorer.

Also rename module to `monte_carlo_scorer.py` (lower case).

python/ml4ir/base/tests/test_monte_carlo_scorer.py
lastmansleeping: Doesn't matter. We need a unit test to help verify what MC Scorer does.

python/ml4ir/base/model/scoring/monte_carlo_scorer.py
```diff
@@ -0,0 +1,123 @@
...
+        Returns
+        -------
+        scores : dict of tensor object
+            Tensor object of the score computed by the model
+        """
+        # TODO replace loop with vector operations for performance
+        scores = super().call(inputs, training=False)[self.output_name]
+        for _ in range(self.monte_carlo_inference_trials):
+            scores += super().call(inputs, training=True)[self.output_name]
+        scores = tf.divide(scores, self.monte_carlo_inference_trials_tf)
```
lastmansleeping, line 122: @mohazahran shouldn't you be dividing by num trials + 1?

python/ml4ir/base/tests/test_monte_carlo_scorer.py
```diff
@@ -0,0 +1,49 @@
...
+            file_io=MagicMock()
+        )
+
+        self.scorer.architecture_op = MagicMock()
+
+    def test_10_calls(self):
+        inputs = {"f0": tf.constant([0.5], dtype=tf.float32)}
+
+        # Call the method
+        result = MonteCarloScorer.call(self.scorer, inputs, training=None)
```
lastmansleeping, line 31: @mohazahran I am a little confused by this usage. Why not do `self.scorer.call(...` instead? I ask for my own understanding.

What does the `MonteCarloScorer.call` on the class do?

python/ml4ir/base/tests/test_monte_carlo_scorer.py
```diff
@@ -0,0 +1,49 @@
...
+
+        # Check that the call method has looped the correct number of times
+        self.assertEqual(str(result["score"].name).count("iadd"), self.scorer.monte_carlo_inference_trials)
```
lastmansleeping, line 34: @mohazahran Can you explain the logic behind what's happening here? I haven't seen this before myself.

lastmansleeping (commented):
@mohazahran Left few comments to understand the test. Can you also remove the unused config and test files that were added previously?

python/ml4ir/base/tests/test_monte_carlo_scorer.py
lastmansleeping: Why not just do something like this? This is more common unit testing practice, right?

```
import unittest
from unittest.mock import patch

class B:
    def super_method(self):
        pass

class A(B):
    def method(self):
        # Calls the super method
        self.super_method()

class TestA(unittest.TestCase):
    def test_super_method_call_count(self):
        with patch('path.to.module.B.super_method') as mock_super_method:
            a = A()
            a.method()
            # Call the method as many times as needed
            # ...

            # Assert the super method was called the expected number of times
            self.assertEqual(mock_super_method.call_count, 1)

if __name__ == '__main__':
    unittest.main()
```

lastmansleeping: The unit test should verify number of times the `RelevanceScorer.call` was called IMO.

python/ml4ir/base/tests/test_monte_carlo_scorer.py
```diff
@@ -0,0 +1,49 @@
...
+        self.scorer.monte_carlo_inference_trials = 0
+
+        inputs = {"f0": tf.constant([0.5], dtype=tf.float32)}
+
+        # Call the method
+        result = MonteCarloScorer.call(self.scorer, inputs, training=None)
+
+        # Check that the call method has looped the correct number of times
+        self.assertEqual(str(result["score"].name).count("iadd"), self.scorer.monte_carlo_inference_trials)
+
```
lastmansleeping, line 46: @mohazahran you should leave a TODO note that currently the unit tests only capture number of calls and do not check for the actual scores.

python/ml4ir/base/model/scoring/monte_carlo_scorer.py
lastmansleeping: Ah I see you've already added it above. Perhaps cleaner to create constant in constructor and do the addition here to avoid confusion?

python/ml4ir/base/tests/data/configs/evaluation_config.yaml
```diff
@@ -0,0 +1,15 @@
+power_analysis:
```
lastmansleeping, line 1: @mohazahran remove

python/ml4ir/base/tests/data/configs/feature_config_monte_carlo.yaml
```diff
@@ -0,0 +1,66 @@
+query_key:
```
lastmansleeping, line 1: @mohazahran remove

python/ml4ir/base/tests/data/configs/model_config_monte_carlo_10.yaml
```diff
@@ -0,0 +1,34 @@
+architecture_key: auto-dag-network
```
lastmansleeping, line 1: @mohazahran remove

python/ml4ir/base/tests/data/linear_sanity_tests/dataset_small.csv
```diff
@@ -0,0 +1,143 @@
+,query_id,record_id,fr,f0,f1,clicked
```
lastmansleeping, line 1: @mohazahran remove

python/ml4ir/base/tests/test_monte_carlo_scorer.py
```diff
@@ -0,0 +1,49 @@
...
+from ml4ir.base.model.scoring.monte_carlo_scorer import MonteCarloScorer
+
+
+class TestMonteCarloScorer(unittest.TestCase):
+
+    def setUp(self):
+        model_config_file = pathlib.Path(__file__).parent / "data" / "configs" / "model_config_monte_carlo_10.yaml"
+
+        with open(model_config_file.as_posix(), 'r') as f:
+            model_config = yaml.safe_load(f)
```
lastmansleeping, line 15: @mohazahran you can get rid of this file and load and just pass a dictionary with the fields you need I guess.

python/ml4ir/base/tests/test_monte_carlo_scorer.py
mohazahran: @lastmansleeping That should be a call to self.scorer I have a bad copy/paste error here. Fixed it.

python/ml4ir/base/tests/test_monte_carlo_scorer.py
mohazahran: This is basically checks the number of "addition" operation used in aggregating the MC trials which should be equals to number of MC trials.

python/ml4ir/base/model/scoring/monte_carlo_scorer.py
mohazahran: In the constructor I have two variables, one if the number of MC trials and the other is the one used in the normalization in the call. I added a comment to explain that

lastmansleeping (approved):
LGTM. Thanks for incorporating changes @mohazahran . Let's plan how to make this deterministic and also optimize with vectorized ops instead of loop. Also, we can increase unit test coverage to check for actual scores.

PR #268, opened 2024-06-06
Adding scale_to_one option for rank transform

lastmansleeping:


python/ml4ir/applications/ranking/features/feature_fns/rank_transform.py
```diff
@@ -38,13 +39,17 @@ def __init__(self, feature_info: dict, file_io: FileIO, **kwargs):
                 If k should be a learnable variable; will be initialized with value of k
             ignore_zero_score: bool
                 Use zero reciprocal rank for score value of 0.0
+            scale_to_one: bool
```
mohazahran, line 42: Can you add why this is a needed feature.
Also, if scale_to_one is False, isn't the RR are already (0,1] ? why multiplying by k+1 would make it (0,1]?
@lastmansleeping

lastmansleeping: It makes the linear combination weights more comparable with other [0, 1] features. Otherwise, reciprocal rank with k=60, would result in very LARGE weights due to small feature values like 1 / (60 + 1) = 0.01..

Multiplying by (k + 1) sets the max possible value to 1 as max possible reciprocal rank is 1 / (k + rank=1)

mohazahran: Ok, so it's mostly re-scaling the RR spectrum, right?

lastmansleeping: Yep.

mohazahran: @lastmansleeping 
In that case, can you change the name SCALE_TO_ONE to something else as SMOOTH_RR_RANGE and perhaps adding a comment to convey that.

lastmansleeping: But we are not smoothening the curve. We are simply "scaling to one". This is more descriptive, no?

lastmansleeping: FWIW, this is what it is called serving side too.

lastmansleeping: I can maybe call it `scale_range_to_one`. Does that work?

lastmansleeping: I went with `scale_range_to_one`. Let me know if that works.

mohazahran (approved):
Thanks Ashish, looks good

PR #275, opened 2024-07-18
Adding deterministic masking for MC

mohazahran:


lastmansleeping (commented):
@mohazahran can you update the design doc we had on quip with the changes so it is easier to read?

python/ml4ir/base/model/scoring/monte_carlo_scorer.py
```diff
@@ -99,12 +91,171 @@ def __init__(
...
+                MonteCarloInferenceKey.NUM_TEST_TRIALS]
+            # Adding 1 here to account of the extra inference run with training=False.
+            self.monte_carlo_test_trials_tf = tf.constant(self.monte_carlo_test_trials + 1, dtype=tf.float32)
+
+        self.features_with_fixed_masks = self.model_config[MonteCarloInferenceKey.MONTE_CARLO_TRIALS].get(MonteCarloInferenceKey.FEATURES_WITH_FIXED_MASKS, [])
+        if self.features_with_fixed_masks:
+            self.features_with_fixed_masks = [f.strip() for f in self.features_with_fixed_masks.split()]
+            self.fixed_feature_count = len(self.features_with_fixed_masks)
+            self.fixed_mask = list(itertools.product([0, 1], repeat=self.fixed_feature_count))
+            self.monte_carlo_fixed_trials_count = tf.constant(len(self.fixed_mask), dtype=tf.float32)
```
lastmansleeping, line 115: @mohazahran this feels hacky to me because so far MCScorer did not have any notion of features and it was being handled by the masking layers.

python/ml4ir/base/model/scoring/monte_carlo_scorer.py
```diff
@@ -99,12 +91,171 @@ def __init__(
...
+            self.monte_carlo_test_trials = self.model_config[MonteCarloInferenceKey.MONTE_CARLO_TRIALS][
+                MonteCarloInferenceKey.NUM_TEST_TRIALS]
+            # Adding 1 here to account of the extra inference run with training=False.
+            self.monte_carlo_test_trials_tf = tf.constant(self.monte_carlo_test_trials + 1, dtype=tf.float32)
+
+        self.features_with_fixed_masks = self.model_config[MonteCarloInferenceKey.MONTE_CARLO_TRIALS].get(MonteCarloInferenceKey.FEATURES_WITH_FIXED_MASKS, [])
+        if self.features_with_fixed_masks:
+            self.features_with_fixed_masks = [f.strip() for f in self.features_with_fixed_masks.split()]
+            self.fixed_feature_count = len(self.features_with_fixed_masks)
+            self.fixed_mask = list(itertools.product([0, 1], repeat=self.fixed_feature_count))
```
lastmansleeping, line 114: @mohazahran can you add explanatory comments for the important steps in this file?

python/ml4ir/base/model/scoring/monte_carlo_scorer.py
```diff
@@ -99,12 +91,171 @@ def __init__(
...
+        tf.Tensor
+            The reshaped and normalized scores tensor. The shape of the tensor will be [batch_size, ...].
+        """
+        all_scores_shape = tf.shape(all_scores)
+        all_scores = tf.reshape(all_scores, tf.concat([[mask_count, batch_size], all_scores_shape[1:]], axis=0))
+
+        # Aggregate the scores across the mask dimension
+        all_scores = tf.reduce_sum(all_scores, axis=0)
+        all_scores = tf.divide(all_scores, self.monte_carlo_fixed_trials_count)
+        return all_scores
```
lastmansleeping, line 204: @mohazahran can be a static method

python/ml4ir/base/model/scoring/monte_carlo_scorer.py
```diff
@@ -120,16 +271,17 @@ def call(self, inputs: Dict[str, tf.Tensor], training=None):
...
+                                        self.monte_carlo_training_trials_tf, training=training)
         else:
-            monte_carlo_trials = self.monte_carlo_test_trials
-            monte_carlo_trials_tf = self.monte_carlo_test_trials_tf
+            if self.use_fixed_mask_in_testing:
+                return self.deterministic_call(inputs, training=training)
+            else:
+                return self.stochastic_call(inputs, self.monte_carlo_test_trials, self.monte_carlo_test_trials_tf,
+                                        training=training)
+
```
lastmansleeping, line 286: @mohazahran this is slowly becoming an if-else tree. Let's refactor.

python/ml4ir/base/model/scoring/monte_carlo_scorer.py
```diff
@@ -99,12 +91,171 @@ def __init__(
...
+            masked_inputs = {}
+            for key, value in inputs.items():
+                masked_value = value
+                for j in range(self.fixed_feature_count):
+                    if key == self.features_with_fixed_masks[j]:
+                        mask = tf.constant(self.fixed_mask[i][j], dtype=tf.float32)
+                        masked_value = tf.multiply(masked_value, mask)
+                masked_inputs[key] = masked_value
+            masked_inputs_list.append(masked_inputs)
+        return masked_inputs_list
```
lastmansleeping, line 145: @mohazahran I don't understand why this mask function is here. It should be in QueryFeatureMask, right? You are modifying what the MCScorer does here.

python/ml4ir/base/model/scoring/monte_carlo_scorer.py
```diff
@@ -99,12 +91,171 @@ def __init__(
...
+        """
+        all_scores_shape = tf.shape(all_scores)
+        all_scores = tf.reshape(all_scores, tf.concat([[mask_count, batch_size], all_scores_shape[1:]], axis=0))
+
+        # Aggregate the scores across the mask dimension
+        all_scores = tf.reduce_sum(all_scores, axis=0)
+        all_scores = tf.divide(all_scores, self.monte_carlo_fixed_trials_count)
+        return all_scores
+
+    def deterministic_call(self, inputs: Dict[str, tf.Tensor], training=None):
```
lastmansleeping, line 206: @mohazahran I would recommend both stochastic and deterministic functions have the same function definition (args)

python/ml4ir/base/model/scoring/monte_carlo_scorer.py
```diff
@@ -99,12 +91,171 @@ def __init__(
...
+        batch_size = tf.shape(list(inputs.values())[0])[0]
+        mask_count = len(self.fixed_mask)
+
+        # Apply fixed masks
+        masked_inputs_list = self.mask_inputs(inputs, mask_count)
+        # score with the fixed mask
+        all_scores = self.score_with_fixed_mask(inputs, masked_inputs_list, training)
+        # Reshape all_scores to separate the mask dimension and the original batch dimension
+        all_scores = self.reshape_and_normalize(all_scores, mask_count, batch_size)
+        return {self.output_name: all_scores}
```
lastmansleeping, line 231: @mohazahran The deterministic call is breaking the definition of what MCScorer does and is leaking code from the QueryFeatureMask. Please refactor.

python/ml4ir/base/config/keys.py
```diff
@@ -137,6 +137,9 @@ class MonteCarloInferenceKey(Key):
     MONTE_CARLO_TRIALS = "monte_carlo_trials"
     NUM_TEST_TRIALS = "num_test_trials"
     NUM_TRAINING_TRIALS = "num_training_trials"
+    FEATURES_WITH_FIXED_MASKS = "features_with_fixed_masks"
```
lastmansleeping, line 140: @mohazahran We should not have this additional argument IMO

python/ml4ir/base/config/keys.py
```diff
@@ -137,6 +137,9 @@ class MonteCarloInferenceKey(Key):
...
+    USE_FIXED_MASK_IN_TRAINING = "use_fixed_mask_in_training"
+    USE_FIXED_MASK_IN_TESTING = "use_fixed_mask_in_testing"
```
lastmansleeping, line 142: @mohazahran Can you look into handling this in QueryFeatureMask instead?

python/ml4ir/base/model/scoring/monte_carlo_scorer.py
mohazahran: @lastmansleeping I see your point, however for fixed masks feature masking are dependent. For example, we want to have all combination of the set of features, run inference and aggregate. 
If we want the masking layer do the masking, then it would need to create duplicates of each query as it cannot just simply apply one mask at a time as in the stochastic MC. Creating duplicates can mess up the training/eval pipeline. This made me move this in the scorer itself.

lastmansleeping:
@mohazahran commenting what we discussed on slack

Right now you have code that does something completely different for stochastic vs deterministic. That's a code smell and will soon become tech debt.

I recommend you do some prototyping and write the new design in the quip doc to make this easier to take forward.

Thinking out loud
1. create some global index variable that can be read both by mask layer and scorer
2. before starting loop in scorer, reset the index counter
3. in for loop
     1. increment index after each call
     2. within mask layer, read this global tf variable to pick the mask row and do element wise product

Ideally, the scorer shouldn't care about fixed vs stochastic. That should be an arg in the mask layer.

python/ml4ir/applications/ranking/model/layers/masking.py
```diff
@@ -133,6 +135,10 @@ class QueryFeatureMask(layers.Layer):
                 [0., 1., 1., 1.]]])
     """
 
+    masking_config = {MonteCarloInferenceKey.FIXED_MASK_COUNT: 1,
+                    MonteCarloInferenceKey.USE_FIXED_MASK_IN_TRAINING: False,
+                    MonteCarloInferenceKey.USE_FIXED_MASK_IN_TESTING: False}
+
```
lastmansleeping, line 141: @mohazahran Why not make these arguments to this constructor when initializing?

python/ml4ir/applications/ranking/model/layers/masking.py
```diff
@@ -159,6 +165,31 @@ def __init__(self,
         self.mask_rate = mask_rate
         self.mask_at_inference = mask_at_inference
         self.requires_mask = requires_mask
+        self.fixed_masks = None
+        self.current_mask_index = -1
+
+    def apply_stochastic_mask(self, inputs, training):
```
lastmansleeping, line 171: @mohazahran add docstrings

python/ml4ir/applications/ranking/model/layers/masking.py
```diff
@@ -159,6 +165,31 @@ def __init__(self,
...
+        if self.mask_at_inference or training:
+            mask = tf.cast(tf.math.greater(tf.random.uniform([batch_size, feature_dim]), self.mask_rate),
+                           dtype=tf.float32)
+            mask = tf.expand_dims(mask, axis=1)
+            mask = tf.tile(mask, [1, sequence_len, 1])
+            return tf.multiply(inputs, mask)
+        else:
+            return inputs
+
+    def apply_fixed_mask(self, inputs, training):
```
lastmansleeping, line 182: @mohazahran add docstrings

python/ml4ir/applications/ranking/model/layers/masking.py
```diff
@@ -159,6 +165,31 @@ def __init__(self,
...
+
+    def apply_stochastic_mask(self, inputs, training):
+        batch_size, sequence_len, feature_dim = tf.shape(inputs)[0], tf.shape(inputs)[1], tf.shape(inputs)[2]
+        if self.mask_at_inference or training:
+            mask = tf.cast(tf.math.greater(tf.random.uniform([batch_size, feature_dim]), self.mask_rate),
+                           dtype=tf.float32)
+            mask = tf.expand_dims(mask, axis=1)
+            mask = tf.tile(mask, [1, sequence_len, 1])
+            return tf.multiply(inputs, mask)
+        else:
```
lastmansleeping, line 179: @mohazahran add in-line comment explanation

python/ml4ir/applications/ranking/model/layers/masking.py
```diff
@@ -159,6 +165,31 @@ def __init__(self,
         self.mask_rate = mask_rate
         self.mask_at_inference = mask_at_inference
         self.requires_mask = requires_mask
+        self.fixed_masks = None
+        self.current_mask_index = -1
+
+    def apply_stochastic_mask(self, inputs, training):
+        batch_size, sequence_len, feature_dim = tf.shape(inputs)[0], tf.shape(inputs)[1], tf.shape(inputs)[2]
+        if self.mask_at_inference or training:
```
lastmansleeping, line 173: [minor] @mohazahran maybe `num_features`?

python/ml4ir/applications/ranking/model/layers/masking.py
```diff
@@ -159,6 +165,31 @@ def __init__(self,
...
+            mask = tf.cast(tf.math.greater(tf.random.uniform([batch_size, feature_dim]), self.mask_rate),
+                           dtype=tf.float32)
+            mask = tf.expand_dims(mask, axis=1)
+            mask = tf.tile(mask, [1, sequence_len, 1])
+            return tf.multiply(inputs, mask)
```
lastmansleeping, line 178: @mohazahran just to confirm - this is the original/old code, right?

python/ml4ir/applications/ranking/model/layers/masking.py
```diff
@@ -181,13 +212,13 @@ def call(self, inputs, mask=None, training=None):
...
+            else:
+                return self.apply_stochastic_mask(inputs, training)
         else:
-            return inputs
-
-
+            if QueryFeatureMask.masking_config[MonteCarloInferenceKey.USE_FIXED_MASK_IN_TESTING]:
+                return self.apply_fixed_mask(inputs, training)
+            else:
+                return self.apply_stochastic_mask(inputs, training)
```
lastmansleeping, line 224: @mohazahran this is much better. We can refactor this further if you want to reduce if-else ladder.

Maybe something like
```
if (training and self.use_fixed_mask_in_training) or ((not training) and self.use_fixed_mask_in_testing)):
    return self.apply_fixed_mask(inputs, training)
else:
    return self.apply_stochastic_mask(inputs, training)
```

python/ml4ir/applications/ranking/model/layers/masking.py
lastmansleeping: I am a little confused that these arguments are hardcoded. How does one override them? Wouldn't it be easier to read these as arguments or from model config?

python/ml4ir/applications/ranking/model/layers/masking.py
```diff
@@ -159,6 +165,31 @@ def __init__(self,
...
+            mask = tf.tile(mask, [1, sequence_len, 1])
+            return tf.multiply(inputs, mask)
+        else:
+            return inputs
+
+    def apply_fixed_mask(self, inputs, training):
+        batch_size, sequence_len, feature_dim = tf.shape(inputs)[0], tf.shape(inputs)[1], tf.shape(inputs)[2]
+        if self.fixed_masks is None:
+            self.fixed_masks = list(itertools.product([0, 1], repeat=feature_dim))
+            QueryFeatureMask.masking_config[MonteCarloInferenceKey.FIXED_MASK_COUNT] = len(self.fixed_masks)
```
lastmansleeping, line 186: @mohazahran `QueryFeatureMask.masking_config` is just a class variable here, right? How do I set its value?

python/ml4ir/applications/ranking/model/layers/masking.py
```diff
@@ -159,6 +165,31 @@ def __init__(self,
...
+            mask = tf.expand_dims(mask, axis=1)
+            mask = tf.tile(mask, [1, sequence_len, 1])
+            return tf.multiply(inputs, mask)
+        else:
+            return inputs
+
+    def apply_fixed_mask(self, inputs, training):
+        batch_size, sequence_len, feature_dim = tf.shape(inputs)[0], tf.shape(inputs)[1], tf.shape(inputs)[2]
+        if self.fixed_masks is None:
+            self.fixed_masks = list(itertools.product([0, 1], repeat=feature_dim))
```
lastmansleeping, line 185: @mohazahran Worth moving this logic to a separate function and adding a unit test. You can also add a TODO to make this richer in the future like we discussed - something like a function that can compute all nC1, nc2...nCk combinations based on a k user input.

python/ml4ir/applications/ranking/model/layers/masking.py
```diff
@@ -159,6 +165,31 @@ def __init__(self,
...
+            QueryFeatureMask.masking_config[MonteCarloInferenceKey.FIXED_MASK_COUNT] = len(self.fixed_masks)
+
+        current_mask = tf.expand_dims(tf.expand_dims(self.fixed_masks[self.current_mask_index], 0), 0)
+        current_mask = tf.tile(current_mask, [batch_size, sequence_len, 1])
+        self.current_mask_index += 1
+        self.current_mask_index %= QueryFeatureMask.masking_config[MonteCarloInferenceKey.FIXED_MASK_COUNT]
```
lastmansleeping, line 191: @mohazahran so we just have to make sure to call from scorer the exact same number of times as mask count, correct?

python/ml4ir/base/model/scoring/monte_carlo_scorer.py
```diff
@@ -99,12 +88,21 @@ def __init__(
...
-        # Adding 1 here to account of the extra inference run with training=False.
-        self.monte_carlo_test_trials_tf = tf.constant(self.monte_carlo_test_trials + 1, dtype=tf.float32)
-        self.monte_carlo_training_trials_tf = tf.constant(self.monte_carlo_training_trials + 1, dtype=tf.float32)
+        self.use_fixed_mask_in_testing = bool(self.model_config[MonteCarloInferenceKey.MONTE_CARLO_TRIALS].get(
+            MonteCarloInferenceKey.USE_FIXED_MASK_IN_TESTING, False))
+        QueryFeatureMask.masking_config[MonteCarloInferenceKey.USE_FIXED_MASK_IN_TESTING] = self.use_fixed_mask_in_testing
+        if not self.use_fixed_mask_in_testing:
+            self.monte_carlo_test_trials = self.model_config[MonteCarloInferenceKey.MONTE_CARLO_TRIALS][
+                MonteCarloInferenceKey.NUM_TEST_TRIALS]
+            QueryFeatureMask.masking_config[MonteCarloInferenceKey.NUM_TEST_TRIALS] = self.monte_carlo_test_trials
```
lastmansleeping, line 105: @mohazahran can you move these default values out as well? Easier to read and track in some config file.

python/ml4ir/base/model/scoring/monte_carlo_scorer.py
```diff
@@ -120,16 +118,19 @@ def call(self, inputs: Dict[str, tf.Tensor], training=None):
...
+                monte_carlo_trials = self.monte_carlo_training_trials
+            else:
+                monte_carlo_trials = QueryFeatureMask.masking_config[MonteCarloInferenceKey.FIXED_MASK_COUNT] - 1
         else:
-            monte_carlo_trials = self.monte_carlo_test_trials
-            monte_carlo_trials_tf = self.monte_carlo_test_trials_tf
+            if not self.use_fixed_mask_in_testing:
+                monte_carlo_trials = self.monte_carlo_test_trials
+            else:
+                monte_carlo_trials = QueryFeatureMask.masking_config[MonteCarloInferenceKey.FIXED_MASK_COUNT] - 1
```
lastmansleeping, line 130: @mohazahran why can't this be done in the constructor? It's an unnecessary if-else ladder on each call, no?

python/ml4ir/base/config/keys.py
```diff
@@ -137,6 +137,9 @@ class MonteCarloInferenceKey(Key):
     MONTE_CARLO_TRIALS = "monte_carlo_trials"
     NUM_TEST_TRIALS = "num_test_trials"
     NUM_TRAINING_TRIALS = "num_training_trials"
+    USE_FIXED_MASK_IN_TRAINING = "use_fixed_mask_in_training"
+    USE_FIXED_MASK_IN_TESTING = "use_fixed_mask_in_testing"
+    FIXED_MASK_COUNT = "fixed_mask_count"
```
lastmansleeping, line 142: @mohazahran Isn't `FIXED_MASK_COUNT` the same as `NUM_TRAINING_TRIALS`? Why do we need two args? Also, why is there no `FIXED_MASK_COUNT_TRAIN` and `FIXED_MASK_COUNT_TEST`?

lastmansleeping (changes requested):
@mohazahran this looks much better. Left a few comments asking for clarification. Can you also add unit tests for all the new functions in `QueryFeatureMask`?

python/ml4ir/applications/ranking/model/layers/masking.py
mohazahran: Yes, but this is done automatically in the monteCarloScorer.

python/ml4ir/base/config/keys.py
mohazahran: No they are different. I will add more comments to clear this confusion. I'm adding the support of having both stochastic and deterministic MC for training and testing. So you can choose for example to have stochastic MC in training by setting `NUM_TRAINING_TRIALS` and have deterministic MC in testing by setting `USE_FIXED_MASK_IN_TESTING`

python/ml4ir/base/model/scoring/monte_carlo_scorer.py
mohazahran: do you mean rely on config to set the default values?

python/ml4ir/applications/ranking/model/layers/masking.py
lastmansleeping: So you are using python globals instead of tensorflow globals? Did you check if this works with the sanity tests in python and java? We might have to revisit with tf globals otherwise.

python/ml4ir/applications/ranking/model/layers/masking.py
```diff
@@ -159,6 +166,105 @@ def __init__(self,
...
+
+        Returns:
+        list of tuples: A list containing all possible fixed masks, where each mask
+                        is a tuple of length `num_features` with elements 0 or 1.
+
+        Example:
+        >>> create_fixed_masks(2)
+        [(0, 0), (0, 1), (1, 0), (1, 1)]
+        """
+        return list(itertools.product([0, 1], repeat=num_features))
```
lastmansleeping, line 267: @mohazahran We should probably leave a todo to accept an arg from user config on how many features you want to pick in each mask? For example, if there are 4 features and we want to use 2 each time, you'd then get 4C2 masks. The code here is fine, it might blow up for longer feature sets.

For example, for 10 features you have 1024 masks. We should also add a flag for selecting n masks at random.

Anyway, this is for the future.

NOTE - this function also creates a mask with all 0s - (0, 0, 0, 0)

python/ml4ir/applications/ranking/model/layers/masking.py
lastmansleeping: ping

python/ml4ir/applications/ranking/tests/test_masking.py
```diff
@@ -83,4 +83,54 @@ def test_mask_all_query_feature(self):
...
+                                mask_at_inference=True,
+                                requires_mask=True)
+        feature_dim = 2
+        expected_result = [(0, 0), (0, 1), (1, 0), (1, 1)]
+        result = mask.create_fixed_masks(feature_dim)
+        self.assertEqual(result, expected_result)
+
+        feature_dim = 3
+        expected_result = [
+            (0, 0, 0), (0, 0, 1), (0, 1, 0), (0, 1, 1),
```
lastmansleeping, line 98: @mohazahran note that this also creates mask with all 0s. It might be undesirable.

lastmansleeping: for future - We could accept a range of values from user on how many they want in each mask.

So if there are 10 features and user wants [2, 4] features in each mask, then we do 10C2 + 10C3 + 10C4 masks.

python/ml4ir/base/model/scoring/monte_carlo_scorer.py
lastmansleeping: Yes, I understand why you do it this way. I was wondering if we will use all those different train vs test combinations in practice. Should be fine for now.

lastmansleeping (approved):
LGTM. Some comments, let me know when you reply.

python/ml4ir/base/model/scoring/monte_carlo_scorer.py
mohazahran: The current model we're interested in actually does that. It uses stochastic MC in training and we will use deterministic MC in testing.

python/ml4ir/applications/ranking/model/layers/masking.py
mohazahran: In my mind, 4 features and we want to use 2 each time. Then use only the 2 features to get full masks (2^2).

That's true it will blow up by increasing the number of features and we would need to sample the full combinations and fix those.

mohazahran: - NOTE - this function also creates a mask with all 0s - (0, 0, 0, 0)
That's true, this part I was in split. I can remove this mask from the mask set. However, since we are possibly just using a subset of all the features in masking, having all zero mask would force the model to only use features not in the mask. which might be beneficial in certain situations.

python/ml4ir/applications/ranking/tests/test_masking.py
mohazahran: I addressed these comments in the thread above.

PR #277, opened 2024-09-18
Optimize predictions in classification_model.py to prevent OOM error

adityajn105:
**Issue**
In relevance model pipeline when we try to Inference we use predict function. In classification_model this predict function try to perform inference on whole data at once which often leads to OOM errors.

**Fix**
In this PR, I provide a fix for this
- Predict function of classification_model is now a generator, means it will operate on a batch of test dataset and yield prediction and then again operate on next batch of test dataset.
- This predict function is also used by ranking pipeline (relevance_model). However predict function in ranking pipeline does not evaluate whole data at once but work in batches and then combine all predictions into one dataframe. But both predict function should behave same, so i updated predict function in relevance model as well to yield batches.
- I am not creating new test cases for predict function as those test cases already exists. But I modified test cases for predict function to make it work with the changes made to this function.

### DIFF
--- a/python/ml4ir/applications/classification/model/classification_model.py
+++ b/python/ml4ir/applications/classification/model/classification_model.py
@@ -247,12 +247,12 @@ class ClassificationModel(RelevanceModel):
         for batch_idx, (batch, label) in enumerate(test_dataset.prefetch(tf.data.experimental.AUTOTUNE)):
             predictions_batch = self.model.predict(batch)
             batch_df = self._create_prediction_dataframe(logging_frequency, (batch,label))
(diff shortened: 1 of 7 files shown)


mohazahran (changes requested):
Thanks @adityajn105 The changes looks good. I suggest the following
- Add better description for the PR. Why it's needed and a very brief summary of changes.
- Add more in code comments to know in the future the rational behind your changes and if you suspect any downsides or TODOs
- Add unit tests.

mohazahran (approved):
Thanks @adityajn105.

PR #284, opened 2025-01-10
Saving model using ExportArchive

ivankoval:


jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
```diff
@@ -255,7 +255,7 @@ class TensorFlowInferenceIT extends TestData {
       predictionPath,
       bundlePath,
       featureConfigPath,
-      "serving_tfrecord_protos",
+      "serving_default_protos",
```
mohazahran, line 258: @ivankoval why chaning from tfrecor to default? My understanding is that in prod. data comes in tfrecord and in GES test it's in tfrecord as well.

python/ml4ir/applications/classification/tests/data/configs/feature_config.yaml
```diff
@@ -11,6 +11,7 @@ query_key:
     name: queryId
     required: false
   default_value: 0
+  tfrecord_type: context
```
mohazahran, line 14: what's this new line "tfrecord_type" referring to. why it is needed?

ivankoval: @mohazahran `tfrecord_type` is needed for the Classification inference test to work. Otherwise it fails with the error below:
```
java.lang.NullPointerException: Cannot invoke "String.equalsIgnoreCase(String)" because the return value of "ml4ir.inference.tensorflow.data.FeatureConfig.tfRecordType()" is null
        at ml4ir.inference.tensorflow.TensorFlowInferenceIT.evaluateClassificationInferenceAccuracy(TensorFlowInferenceIT.scala:364)
        at ml4ir.inference.tensorflow.TensorFlowInferenceIT.testClassificationGeneratedModelBundle(TensorFlowInferenceIT.scala:347)
```

For example, in the feature config for classification inference test, we already have `tfrecord_type` fields ([link](https://github.com/salesforce/ml4ir/blob/master/jvm/ml4ir-inference/src/test/resources/classification/feature_config.yaml)). For some reason, we don't have `tfrecord_type` in the  [python/ml4ir/applications/classification/tests/data/configs/feature_config.yaml](https://github.com/salesforce/ml4ir/pull/284/files/91a58cfc754d7ffa4c633050c08cfebde20f31cb#diff-3e428f90aa3062c9d9798825ae7a7be4d84e3c79661688820f0f51ff3a7f31fb) which is used in the test.

jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala
ivankoval: From what I understand, Tensorflow serving requires a `serving_default` endpoint alway set ([link](https://www.tensorflow.org/api_docs/python/tf/keras/export/ExportArchive)):

> Note on TF-Serving: all endpoints registered via add_endpoint() are made visible for TF-Serving in the SavedModel artifact. In addition, the first endpoint registered is made visible under the alias "serving_default" (unless an endpoint with the name "serving_default" was already registered manually), since TF-Serving requires this endpoint to be set.

So, when we export the model with `ServingSignatureKey.TFRECORD` endpoint, we get two endpoints `serving_default_photos` and `serving_tfrecord_protos`. For some reason, `serving_tfrecord_protos` doesn't work in the inference test. It gives `to-do` error. So I have to use `serving_default_protos ` instead which works. Also, we save the model with different endpoints in the different paths. So in the test, we load bundle from the `serving_tfrecord` folder ([ref](https://github.com/salesforce/ml4ir/blob/39986ab3fb65a2c6be09c4fb2783e850ce47d43a/jvm/ml4ir-inference/src/test/integration/ml4ir/inference/tensorflow/TensorFlowInferenceIT.scala#L233C80-L233C96)). Whereas the default endpoint is saved to `serving_default` folder.

PR #285, opened 2025-01-20
Upgrading tesorflow version while keeping keras 2

ivankoval:


mohazahran (approved):
That @ivankoval for this work

